PR #48, opened 2022-07-01
add story for layout example

rezrah:
https://user-images.githubusercontent.com/13340707/176877772-bb7ae139-d87c-432a-aa28-84cab1cebc06.mov

## What's changing?

- Add a Storybook story that showcases all available UI patterns together as an example layout.

## Why?

 To help with integration testing and provide a story that will align closer to end-user usage. 

As new major UI patterns are added to Primer Brand, we should continue adding them here.

## How?

- A single story that composes examples from other stories into a single, new story. This avoids having to maintain yet more code examples while still achieving the stories intention.

#### New section

Screenshot 2022-07-01 at 11 28 26


rezrah:
Superseded by https://github.com/primer/brand/pull/100

PR #70, opened 2022-08-24
Fix duplicate name errors in visual test suites

rezrah:
Fixes a problem where visual tests fail if you add a Story with the same name as an existing one.

E.g. Two stories called Playground result in two tests with the same name being generated.

Before: 
`test('Playground', () => {`

After:
`test('ComponentName / Playground', () => {`

Fixes problem by prefixing the test names with the group name.

Unfortunately this does mean we need to regenerate all snapshots too 😢


colebemis (approved):
🚀

PR #78, opened 2022-09-14
Created a Medium Button Style

JoshBowdenConcepts:
## Summary

In this PR I have created the medium button style for the button component. I have also updated the hero component to use the large button style as it was intended.

image

## List of notable changes:

- **added** medium button style for the button component
- **updated** hero component to use the large button style

Making sure that the Button component's new style meets the spec provided.

## Steps to test:

1. Open the Button component in CI-deployed preview environment
2. Go to Button story in Storybook
3. Verify that Button has two configurable states for size 'large' and 'medium' with 'medium' being the default
4. Check that the Hero component is using the large button style

- [Issue](https://github.com/github/primer/issues/1093)
- [Spec](https://www.figma.com/file/BJ95AjraesmRCWsKA013GS/Primer-Brand?node-id=373%3A8028)

## Contributor checklist:

## Reviewer checklist:


src/Button/Button.module.css
```diff
@@ -50,3 +49,11 @@
 .Button--secondary:hover {
   box-shadow: inset 0 0 0 2px var(--brand-color-neutral-emphasisPlus);
 }
+
+.Button--size-medium {
+  padding: 1rem 1.5rem;
```
rezrah, line 54: can we update these to use the css variables please? 

```suggestion
  padding: var(--base-size-16) var(--base-size-24);
```

src/Button/Button.module.css
```diff
@@ -50,3 +49,11 @@
...
+}
+
+.Button--size-large {
+  padding: 1rem 2rem;
```
rezrah, line 58: Same as previous comment:

can we update these to use the css variables please? 

```suggestion
  padding: var(--base-size-16) var(--base-size-32);
```

src/Button/Button.tsx
```diff
@@ -6,16 +6,23 @@ import '../../lib/design-tokens/css/tokens/functional/components/button/colors-w
 import type {BaseProps} from '../component-helpers'
 import styles from './Button.module.css'
 
+export enum LabelFontSize {
```
rezrah, line 9: What was the reason behind using an enum here, rather than relying on the union like other types. Also noticed that it's being exported from this file, so it'll end up in the package root. I'm not sure that's ideal 🤔. Can we stick to the union?

src/Button/Button.module.css
JoshBowdenConcepts: I was curious about this. The intial implementation was done with the rem values but I knew we had variables.

src/Hero/Hero.tsx
```diff
@@ -34,11 +34,11 @@ export const Hero = forwardRef<HTMLDivElement, HeroProps>(
           </Text>
         ) : null}
         <div className={styles['Hero-actions']}>
-          <Button as="a" variant="primary" href={primaryAction.href}>
+          <Button as="a" variant="primary" size={'large'} href={primaryAction.href}>
```
rezrah, line 37:
```suggestion
          <Button as="a" variant="primary" size="large" href={primaryAction.href}>
```

rezrah: Can we do this? ☝️

src/Button/Button.tsx
JoshBowdenConcepts: I have no issue removing the export if we don't want external access to this

src/Hero/Hero.tsx
JoshBowdenConcepts: done!

src/Button/Button.tsx
rezrah: 👍 yeah, let's not surface the object in the main index.

Ash has confirmed there's no sizes besides `medium` and `large` if that affects the decision around doing it via ternary instead of enum. Also, as it's going to be a private variable, can we call it `ButtonTextSizeLookup` (or similar) instead, so the name indicates it's a map?

src/Button/Button.module.css
rezrah: yep, this component predates the design tokens.. it's the OG component 😁

rezrah (commented):
@JoshBowdenConcepts - can you regenerate the snapshots please? Playwright found some visual differences, which all look correct (we were using the wrong horizontal padding on button before 🙈 oof)... 

https://user-images.githubusercontent.com/13340707/190193884-cea55c1c-56eb-473b-bbe6-930590e6ad30.mov

.changeset/cuddly-peas-try.md
```diff
@@ -0,0 +1,5 @@
+---
+"@primer/react-brand": patch
+---
+
+Created a Medium Button Style
```
rezrah, line 5:
```suggestion
Added a new `size` prop to `Button` and a `medium` sized variant, making the previous `large` size optional.

⚠️ This update should be considered a breaking change, despite being issued in a `minor` release. 

Selectable values: `medium` (default) and `large`.

**Upgrade steps:**

```diff
- <Button>before</Button>
+ <Button size="large">after</Button>
```

.changeset/cuddly-peas-try.md
```diff
@@ -0,0 +1,5 @@
+---
+"@primer/react-brand": patch
```
rezrah, line 2: This is technically a breaking change, so can we change this to `minor` please? (Lib is pre-v1, so we don't need to change major yet)

rezrah:
```suggestion
"@primer/react-brand": minor
```

rezrah:
@JoshBowdenConcepts - this is looking great. Thanks for making this change.

Just a few more things I reckon we should do before it's merged...

- I think `medium` and `large` parameters qualify as 'design logic', so we would ideally put a test around it IMO. You can see some of the other tests for examples of how it's done elsewhere. This is non-blocking though, and happy to discuss when you're online.
- I've suggested some changes to the changeset, given it's going to be published in the Change log and we're technically issuing a breaking change by defaulting to the smaller button
- I've added the new snapshots in [this commit](https://github.com/primer/brand/pull/78/commits/11a6ad4a4aff58ed54fe9b6550a35bdbf07c549c)

.changeset/cuddly-peas-try.md
```diff
@@ -0,0 +1,15 @@
...
+
+⚠️ This update should be considered a breaking change, despite being issued in a `minor` release. 
+
+Selectable values: `medium` (default) and `large`.
+
+**Upgrade steps:**
+
+```diff
+- <Button>before</Button>
++ <Button size="large">after</Button>
```
rezrah, line 15: missing a closing ``` here. Think it was chopped off my suggestion earlier because I nested code blocks

can you add? I can't seem to add it without MD editor converting it :(

```diff
- <Button>before</Button>
+ <Button size="large">after</Button>
```

rezrah (approved):
🙌 Looks great.. nice one @JoshBowdenConcepts.. ship it 🚢

.changeset/cuddly-peas-try.md
JoshBowdenConcepts: Should be fixed now

PR #91, opened 2022-10-07
Add Storybook Storysource addon

langermank:
## Summary

[Storysource](https://storybook.js.org/addons/@storybook/addon-storysource) adds a panel with a code snippet of the story "source" code for easy copy/paste. The source code view depends on the format of the story, and unfortunately isn't compatible with `template.bind()` syntax. However, if we treat "Feature" stories as code snippets and allow the Playground story to use controls, we kind of get the best of both worlds. We might consider investing in our own source code addon that is more consistent.

Closes https://github.com/github/primer/issues/1385

Are we okay with some of the existing stories showing code like this?

image

## Steps to test:

- Open the Storybook PR preview deploy
- Navigate to the "Story" tab in the addon panel

## Contributor checklist:

## Reviewer checklist:

- [x] Check that pull request and proposed changes adhere to our [contribution guidelines](../../CONTRIBUTING.md) and [code of conduct](../../CODE_OF_CONDUCT.md)
- [x] Check that tests prove the feature works and covers both happy and unhappy paths
- [x] Check that there aren't other open Pull Requests for the same update/change

## Screenshots:

image

### DIFF
--- a/.storybook/main.js
+++ b/.storybook/main.js
@@ -5,7 +5,8 @@ module.exports = {
     '@storybook/addon-links',
     '@storybook/addon-essentials',
     '@storybook/addon-interactions',
-    'storybook-css-modules-preset'
+    'storybook-css-modules-preset',
+    '@storybook/addon-storysource'
   ],
   framework: '@storybook/react',
   core: {
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@primer/react-brand",
-  "version": "0.5.0",
+  "version": "0.5.1",
   "lockfileVersion": 2,
   "requires": true,
   "packages": {
     "": {
       "name": "@primer/react-brand",
-      "version": "0.5.0",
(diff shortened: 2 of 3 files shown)


rezrah:
> Are we okay with some of the existing stories showing code like this?

I think it's fine for now. I guess the actual way to do this is to put the entire snippet in there rather than binding, like you said in the description? 

Also noticed it doesn't work for interactions, for the same reason. We'll see how many folks use this feature, and if it's popular we can refactor those snippets.

langermank:
@rezrah I'm going to write an ADR, but I'm proposing we write "Feature" stories without a template like [this](https://github.com/primer/react/pull/2374/files#diff-d5b6e9810b3e7a084c2d27d59b7cf2e63895c5b8ca45634afce0f41cf1c10923). And then each component will have a Playground story which utilizing all of the Storybook controls. 

... but its very silly that the very syntax Storybook recommends doesn't play nice with their own addon (there's several open issues)

PR #110, opened 2022-10-28
Button group component

JoshBowdenConcepts:
## Summary

Creating a ButtonGroup component that will handle the positioning and styling of buttons to be used in the CTABanner and Hero components

## List of notable changes:

- **created** ButtonGroup component
- **added** story to Storybook documentation

- The functionality and robustness of the component

## Steps to test:

1. Go to ButtonGroup story in Storybook
2. Verify that the ButtonGroup works as expected
3. Create a demo site and import the ButtonGroup
4. Try putting in both expected and not expected items within it

- [Figma design](https://www.figma.com/file/BJ95AjraesmRCWsKA013GS/Primer-Brand?node-id=1020%3A29442)
- [Storybook documentation](https://primer-f4e099c77f-26139705.drafts.github.io/storybook/?path=/story/components-buttongroup--primary)

## Contributor checklist:

## Reviewer checklist:


packages/react/src/ButtonGroup/ButtonGroup.module.css
```diff
@@ -0,0 +1,8 @@
+.ButtonGroup {
+  display: inline-grid;
+  gap: 8px;
```
rezrah, line 3: We have an equivalent design token for this.. can we switch to that please? 

```suggestion
  gap: var(--base-size-8);
```

rezrah, line 3: Can we switch this out for the design token equivalent?

```suggestion
  gap: var(--base-size-8);
```

packages/react/src/ButtonGroup/ButtonGroup.module.css
```diff
@@ -0,0 +1,8 @@
...
+}
+
+.ButtonGroup--horizontal {
+  grid-template-columns: auto auto;
+}
```
rezrah, line 8: Are you missing styles for vertical? I can see it listed in the direction props.

packages/react/src/ButtonGroup/ButtonGroup.module.css
```diff
@@ -0,0 +1,8 @@
+.ButtonGroup {
+  display: inline-grid;
+  gap: 8px;
+}
+
+.ButtonGroup--horizontal {
```
rezrah, line 6: Could direction be controlled responsively instead of explicitly through props?

I would have assumed that buttons should always stack vertically in narrow viewports and span horizontally on larger ones. 

Right now, this happens on the narrow viewports 👇 
two super squished buttons in a narrow viewport

If we keep this approach of direction via a prop, but still wanted the functionality I mentioned it would force app-side devs to have a matchMedia listener, which probably isn't ideal. 

cc. @ashygee also for any design considerations.

packages/react/src/ButtonGroup/ButtonGroup.stories.tsx
```diff
@@ -0,0 +1,69 @@
...
+Primary.args = {}
+
+const SingleButtonTemplate: ComponentStory<typeof ButtonGroup> = args => (
+  <ButtonGroup {...args}>
+    <Button>This is one button</Button>
+  </ButtonGroup>
+)
+
+export const SingleButton = SingleButtonTemplate.bind({})
+Primary.args = {}
```
rezrah, line 28: Duplicate?

packages/react/src/ButtonGroup/ButtonGroup.stories.tsx
```diff
@@ -0,0 +1,69 @@
...
+
+const LargeButtonTemplate: ComponentStory<typeof ButtonGroup> = args => (
+  <ButtonGroup buttonSize="large" {...args}>
+    <Button>This is one button</Button>
+    <Button>This is the second button</Button>
+  </ButtonGroup>
+)
+
+export const LargeButtons = LargeButtonTemplate.bind({})
+Primary.args = {}
```
rezrah, line 38: Another duplicate. Should be removed.

packages/react/src/ButtonGroup/ButtonGroup.tsx
```diff
@@ -0,0 +1,46 @@
+import clsx from 'clsx'
+import React, {forwardRef, type Ref} from 'react'
+import '@primer/brand-primitives/lib/design-tokens/css/tokens/functional/components/button/colors-with-modes.css'
```
rezrah, line 3: I don't think you need this here, it's already imported in Button?

packages/react/src/ButtonGroup/ButtonGroup.module.css
JoshBowdenConcepts: vertical styles happen implicitly I think relying on browsers for this would make more sense. We can update this if the responsive update you mentioned affects this.

packages/react/src/ButtonGroup/index.ts
```diff
@@ -0,0 +1 @@
+export * from './ButtonGroup'
```
rezrah, line 1: Can you also add the same export statement in `src/index.ts` please?

It won't get published in the package without that.

Also can't see a changeset here, could you add one in too please? [Example here from ComparisonTable](https://github.com/primer/brand/blob/6e7bdecee69458c1432d34f598479ac44029e20f/.changeset/honest-weeks-shake.md).

packages/react/src/ButtonGroup/ButtonGroup.stories.tsx
```diff
@@ -0,0 +1,64 @@
...
+const SingleButtonTemplate: ComponentStory<typeof ButtonGroup> = args => (
+  <ButtonGroup {...args}>
+    <Button>This is one button</Button>
+  </ButtonGroup>
+)
+
+export const SingleButton = SingleButtonTemplate.bind({})
+
+const LargeButtonTemplate: ComponentStory<typeof ButtonGroup> = args => (
+  <ButtonGroup buttonSize="large" {...args}>
```
rezrah, line 30: Can this be set through args instead?

something like: 

```js
LargeButtonTemplate.args = {
  size: 'large'
}
```

I think the argTypes need to be manually configured too 👇 

https://user-images.githubusercontent.com/13340707/199083146-699275f6-f523-4e74-b213-34f4c8c59524.mov

packages/react/src/ButtonGroup/ButtonGroup.stories.tsx
```diff
@@ -0,0 +1,64 @@
...
+  <ButtonGroup direction="vertical" {...args}>
+    <Button>This is one button</Button>
+    <Button>This is the second button</Button>
+  </ButtonGroup>
+)
+
+export const VerticalButtons = VerticalButtonTemplate.bind({})
+
+const LinkButtonTemplate: ComponentStory<typeof ButtonGroup> = args => (
+  <ButtonGroup buttonsAs={'a'} {...args}>
```
rezrah, line 48: This one is missing a `href`, which should have been a required prop. Wonder why type checker didn't pick it up 🤔 

```suggestion
  <ButtonGroup buttonsAs={'a'} href="#" {...args}>
```

packages/react/src/ButtonGroup/ButtonGroup.stories.tsx
JoshBowdenConcepts: Unfortunately no because of the way the props are initialized on the copy. I was unable to use that specific prop name.

packages/react/src/ButtonGroup/ButtonGroup.stories.tsx
JoshBowdenConcepts: I will add this but it should likely but checked in the main component

JoshBowdenConcepts: And it wouldn't go on the group but on the button itself.

rezrah: ah yes, good point.

packages/react/src/ButtonGroup/index.ts
JoshBowdenConcepts: done

rezrah (approved):
Looks great 🎉 . Will defer to @ashygee for final design review and approval.

rezrah (commented):
Forgot to mention, I can see that unit tests aren't in this PR. Are you able to backfill those in a future PR please? There's some important logic here that should have coverage.

ashygee:
@JoshBowdenConcepts overall looks good! A few notes:

- In ButtonGroup pairings, and I think more generally now, we do **not** show the `>` arrow in secondary buttons. cc @tobiasahlin to double check but I believe it was him who provided this guidance.
- I am curious if "direction" makes more sense than "layout". In Figma I am using a prop name of "layout" which has two values, "inline" and "stacked". I think that either can work but for parity sake, it would be good to align.
- As I'm not a developer, not being able to easily swap between sizes with prop values seems cumbersome to me. What do others think?
- How would I control the pairing if I wanted the ButtonGroup to house a Button with a secondary variant + a Button with a subtle variant? In Figma I made a prop called "pair" but that's more for designers and I don't think helpful for developers. The "pair" property allows for the following combinations:
  - primary + secondary
  - primary + link (subtle)
  - secondary + link (subtle)

JoshBowdenConcepts:
> @JoshBowdenConcepts overall looks good! A few notes:
> 
> * In ButtonGroup pairings, and I think more generally now, we do **not** show the `>` arrow in secondary buttons. cc @tobiasahlin to double check but I believe it was him who provided this guidance.
> * I am curious if "direction" makes more sense than "layout". In Figma I am using a prop name of "layout" which has two values, "inline" and "stacked". I think that either can work but for parity sake, it would be good to align.
> * As I'm not a developer, not being able to easily swap between sizes with prop values seems cumbersome to me. What do others think?
> * How would I control the pairing if I wanted the ButtonGroup to house a Button with a secondary variant + a Button with a subtle variant? In Figma I made a prop called "pair" but that's more for designers and I don't think helpful for developers. The "pair" property allows for the following combinations:
>   
>   * primary + secondary
>   * primary + link (subtle)
>   * secondary + link (subtle)

1. The arrow is configurable by the button itself
2. I think direction makes more sense because layout could mean more than two things but not picky on that either
3. You can swap between sizes with a property. It is buttonSize.
4. You could just pass the variant to each button as children. It wouldn’t be the default interaction but it is easy to do.

rezrah:
> The arrow is configurable by the button itself

@JoshBowdenConcepts this feels like design logic, can we disable this by default? Should also have some unit test coverage around it too eventually.

> You can swap between sizes with a property. It is buttonSize.

This is somewhat related to the Storybook control issue I mentioned previously. Are you able to manually declare these controls if the props can't be automatically inferred from code in the Story? Would be useful for anyone looking at the Storybook playground for this component, and it's not working correctly right now.

button size showing as a object instead of a selection


> You could just pass the variant to each button as children. It wouldn’t be the default interaction but it is easy to do.

Sounds like a good candidate for a Storybook story?

.changeset/rude-parents-sing.md
```diff
@@ -0,0 +1,16 @@
+---
+"@primer/react-brand": minor
+---
+
+Added `ButtonGroup` component
+
+Usage example:
+
+```jsx
+<ButtonGroup {...args}>
```
rezrah, line 10: Changeset shouldn't contain Storybook `args`.

packages/react/src/ButtonGroup/ButtonGroup.stories.tsx
```diff
@@ -0,0 +1,64 @@
...
+} as ComponentMeta<typeof ButtonGroup>
+
+const Template: ComponentStory<typeof ButtonGroup> = args => (
+  <ButtonGroup {...args}>
+    <Button>This is one button</Button>
+    <Button>This is the second button</Button>
+  </ButtonGroup>
+)
+
+export const Primary = Template.bind({})
```
rezrah, line 18: Forgot to mention earlier, but the first story will be called 'Playground' going forward, and all controls should be configurable...

Related ADR: https://github.com/github/primer/pull/1387

While PR is still open, it's still very likely to be merged soon so good to pre-emptively align with it.

.changeset/rude-parents-sing.md
JoshBowdenConcepts: fixed

packages/react/src/ButtonGroup/ButtonGroup.stories.tsx
JoshBowdenConcepts: This is updated.

JoshBowdenConcepts:
@rezrah let me know if the new storybook looks the way you would expect and I think the new controls around size and declaration for buttonsAs makes more sense to a user now.

rezrah:
> @rezrah let me know if the new storybook looks the way you would expect and I think the new controls around size and declaration for buttonsAs makes more sense to a user now.

👍 looks good, thanks for removing direction prop too. I've regenerated the snapshots and pulled in `main` to get your integration tests to passs.

Somewhat unrelated, but is our large button supposed to enlarge text in addition to widen padding? or just latter @ashygee? I know we discussed this before, but it feels strange that only the horizontal padding gets bigger and everything else stays the same.

Issue #99, opened 2022-10-14
Feature Request / Idea - View Grid columns for development


rfearing:
#### What is proposed:
I propose adding a "toggle-able" visual layer to our development environments. This layer would display the Primer grid columns as an overlay overtop our markup, similar to how Figma allows us to toggle the grid overlay inside of our designs.

This would utilize our tokens and would be a visual aide in development. 

#### Reasoning
Often, particularly when grids are nested, it's easy to veer from the design without realizing it. A margin or padding in a parent element may affect the alignment of an element within the intended grid.

I believe having a visual indicator inside our design system would:
- help the speed and ease development as developers could turn on the grid to quickly and easily see how closely the markup matches the design (more easily comparing apples to apples) by placing items precisely on the grid
- improve the final product: differences between design and dev should be easier to spot (Less of *Something looks off but I don't know what*)
- Help developers see nested grids work in relation to the overall grid	

#### Proposed implementation:
- This could be "toggle-able" with a short key to turn on/off this feature.
- Renders only in dev / staging
- Inspiration from [Griddle](https://github.com/wearebraid/griddle)
  - UVA uses Griddle and can be seen in [a production site](https://www.commerce.virginia.edu/): (shift + control + L)
  - <img width="1820" alt="Screen Shot 2022-10-14 at 3 05 10 PM" src="https://user-images.githubusercontent.com/9965014/195922291-ad7ac4ec-445e-4ab1-bad5-dd803374aec3.png">
  - On the site of the creators' of Griddle: 
  - ![](https://user-images.githubusercontent.com/9965014/195922500-5b18a6d4-4266-4496-acd2-884e4b305d70.gif)

rfearing:
Note to self: Add examples of use of the grid system

PR #112, opened 2022-10-31
Designer / Developer Integration Grid

rfearing:
## Summary (WIP / Proposal)

This adds a visual layer to display grid-columns as an overlay, overtop our markup. This is similar to how Figma allows us to toggle the grid overlay inside of our designs. I feel this could greatly benefit developer by speeding up the development time and potentially shortening the feedback loop:
- Assist in creating faithful translation of source design into code
- Help eliminate guesswork (Does this line up correctly?)  by placing items precisely on the grid
- Eliminate or reduce the need for alignment feedback in the Q&A process
- Improve integration between stakeholder, designer and developer

#### Not yet added:
- Only show in dev environments
- Add tests

#### Additional Props Discussion: (What should we add?)
- maxWidth
- gutterSize (with breakpoints?)
- baseProps
  - className
  - id
  - ref

- What should be added?
- Anything added that's unnecessary?
- Any defaults should change?

## Steps to test:

1. View:
    - DevGrid Component Story
    - KitchenSink Example Layout
    - Resources Example Layout
2. Click inside the story (to give the iFrame focus) and type `control + shift + L`
3. Verify that
    - You can toggle the grid
    - It does not interfere with any mouse events

## Supporting resources

- For https://github.com/primer/brand/issues/99

## Screenshots

https://user-images.githubusercontent.com/9965014/199043921-a9b2bc18-9a2c-4874-ab40-3d7215b3b446.mov

image


rfearing:
@tobiasahlin I'd love to hear your thoughts

tobiasahlin:
I think this is great! Love that we're thinking about this 😌  And does in fact show in this PR that the top nav is misaligned with the grid? 🤔 

Gutters: We have two different kind of gutters, normal and spacious, and it would be great to be able to account for both…! Also, what happens on mobile? Will we keep to 12 columns even on `sm` and `md`? Feels like it could be very stripy… but maybe it makes sense?

packages/react/src/LayoutExamples/KitchenSink.tsx
```diff
@@ -23,6 +24,7 @@ type KitchenSinkProps = React.HTMLAttributes<HTMLDivElement>
 export function KitchenSink(props: KitchenSinkProps) {
   return (
     <>
+      <DevGrid />
```
rezrah, line 27: I think you should add this as a global decorator, so that it can be used in every story.

packages/react/src/LayoutExamples/KitchenSink.tsx
```diff
@@ -46,6 +48,23 @@ export function KitchenSink(props: KitchenSinkProps) {
...
+            <code
+              style={{
+                padding: '0.25rem 0.5rem',
+                backgroundColor: 'var(--brand-color-canvas-subtle)'
+              }}
+            >
+              ctrl + shift + L
+            </code>
+          </Text>
+        </div>
```
rezrah, line 66: I'm not sure this should go in here 🤔 .. as this functions a recipe layout for Primer components. Instead, could we turn the DevGrid into a global decorator? Down the line, I can also see this toggle-able through the Storybook toolbar via a Storybook addon.

packages/react/src/DevGrid/DevGrid.tsx
```diff
@@ -0,0 +1,46 @@
...
+  useDevGridKeyCombo(handleToggle)
+  useEffect(() => {
+    const root = document.documentElement
+    root.style.setProperty('--devGrid-grid-size', `${size}`)
+  }, [size])
+
+  let col = size
+  const columns: ReactElement[] = []
+
+  while (col > 0) {
```
rezrah, line 40: Is it possible to move imperative code like this inside the return statement where you need it? or  encapsulate it in a function and then call that?

packages/react/src/DevGrid/DevGrid.module.css
```diff
@@ -0,0 +1,43 @@
+:root {
+  --scrollbar-width: 0px;
+  --devGrid-width: 90%;
+  --devGrid-max-width: 80rem;
+  --devGrid-grid-size: 12;
```
rezrah, line 5: @rfearing we can make this responsive through a design token! let's pair up and i'll show you how.

packages/react/src/DevGrid/DevGrid.module.css
```diff
@@ -0,0 +1,43 @@
...
+  z-index: var(--devGrid-zIndex);
+}
+
+.DevGrid--show {
+  transform: translateX(-50%);
+}
+
+@media screen and (min-width: 768px) {
+  .DevGrid {
+    column-gap: var(--devGrid-column-gap);
```
rezrah, line 36: wonder if we can do this through a token instead. what do you think? Check out typography components for an example.

packages/react/src/DevGrid/DevGrid.module.css
rezrah: I also think these should match the grid columns used in Figma by site design. @danielguillan @ashygee, can we chat about this at the next sync?

packages/react/src/index.ts
```diff
@@ -14,3 +14,4 @@ export * from './FAQ'
 export * from './forms'
 export * from './Stack'
 export * from './Testimonial'
+export * from './DevGrid'
```
rezrah, line 17: Crazy idea, but what do you think about publishing this as a separate package? Something like `@primer/brand-devtools`.. it would still live in this repo, just isolated and we can keep it enabled in Storybook as a dependency.

rfearing: I like that idea

packages/react/src/LayoutExamples/KitchenSink.tsx
rfearing: > Gutters: We have two different kind of gutters, normal and spacious, and it would be great to be able to account for both…! Also, what happens on mobile? Will we keep to 12 columns even on `sm` and `md`? Feels like it could be very stripy… but maybe it makes sense?

I guess it really depends on how the designers utilize the grid in mobile, right?

tobiasahlin:
@rfearing Yes! Could you reach out to site-design and check how they usually think about the grid on mobile? 😌 

Also, to bring this up again:
> And does in fact show in this PR that the top nav is misaligned with the grid? 

Either the grid is wrong here, or the navigation component 🙏 can we tell which one it is?

packages/react/src/DevGrid/DevGrid.tsx
```diff
@@ -0,0 +1,92 @@
...
+        setNumCols(Breakpoints.NARROW)
+      }
+      if (width >= mediumBreakpointToPx && width < largeBreakpointToPx) {
+        setNumCols(Breakpoints.MEDIUM)
+      }
+      if (width >= largeBreakpointToPx) {
+        setNumCols(Breakpoints.WIDE)
+      }
+    }
+  }, [numCols, setNumCols, width, mediumBreakpointToPx, largeBreakpointToPx])
```
rfearing, line 72: @rezrah Should we remove `numCols` from requirements?

Requirements that don't change (shouldn't cause extra render):
- mediumBreakpointToPx
- largeBreakpointToPx
- setNumCols

Requirement that we want to trigger effect:
- width

Does `numCols` cause a change? So, for example, we see that `width` changes, so then `numCols` changes. Since it's in our list, does that mean another render because it will see `numCols` change?

packages/react/src/DevGrid/DevGrid.module.css
```diff
@@ -0,0 +1,43 @@
+:root {
+  --scrollbar-width: 0px;
+  --devGrid-width: 90%;
+  --devGrid-max-width: 80rem;
+  --devGrid-grid-size: 4;
+  --devGrid-column-gap: var(--brand-stack-gap-normal);
+  --devGrid-column-gap-mobile: var(--brand-stack-gap-condensed);
+  --devGrid-column-color: rgba(255, 0, 0, 0.1);
```
rezrah, line 8:
```suggestion
  --devGrid-column-color: hsla(var(--base-color-scale-purple-0-hsl / 0.9));
```

packages/react/src/DevGrid/DevGrid.tsx
```diff
@@ -0,0 +1,92 @@
...
+ * Main stylesheet (as a CSS Module)
+ */
+import styles from './DevGrid.module.css'
+
+/*
+ * Generate from design tokens or FIgma
+ */
+enum Breakpoints {
+  NARROW = 4,
+  MEDIUM = 8,
```
rezrah, line 29: This one should be regular, as pointed out by @danielguillan

PR #128, opened 2022-11-10
Add cta banner component

JoshBowdenConcepts:
## List of notable changes:

-
-

-
-

## Steps to test:

1.
1.
1.

-
-

## Contributor checklist:

## Reviewer checklist:

## Screenshots:

<table>
<tr>
</tr>
<tr>
<td valign="top">

 </td>
<td valign="top">

</td>
</tr>
</table>

### DIFF
--- a/packages/react/src/CTABanner/CTABanner.module.css
+++ b/packages/react/src/CTABanner/CTABanner.module.css
@@ -4,10 +4,12 @@
   flex-direction: column;
   border-radius: 6px;
   padding: 40px 32px;
-  /* Box-Shadow */
-  /* box-shadow: 0 0 0 1px rgb(53 72 91 / 14%), 0 2.75px 2.21px rgb(0 0 0 / 7%), 0 6.65px 5.32px rgb(0 0 0 / 4%),
-    0 12.5px 10px rgb(0 0 0 / 3%), 0 22px 18px rgb(0 0 0 / 3%), 0 42px 33.4px rgb(0 0 0 / 2%),
-    0 100px 80px rgb(0 0 0 / 2%) !important; */
+  box-shadow: 2.8px 2.8px 2.2px rgba(26, 127, 55, 0.002), 6.7px 6.7px 5.3px rgba(26, 127, 55, 0.005),
+    12.5px 12.5px 10px rgba(26, 127, 55, 0.01), 22.3px 22.3px 17.9px rgba(26, 127, 55, 0.016),
+    41.8px 41.8px 33.4px rgba(26, 127, 55, 0.027), 100px 100px 80px rgba(26, 127, 55, 0.07),
+    -2.8px -2.8px 2.2px rgba(3, 102, 214, 0.002), -6.7px -6.7px 5.3px rgba(3, 102, 214, 0.005),
+    -12.5px -12.5px 10px rgba(3, 102, 214, 0.01), -22.3px -22.3px 17.9px rgba(3, 102, 214, 0.016),
+    -41.8px -41.8px 33.4px rgba(3, 102, 214, 0.027), -100px -100px 80px rgba(3, 102, 214, 0.07);
 }
 
 .CTABanner:after {
--- a/packages/react/src/CTABanner/CTABanner.module.css
+++ b/packages/react/src/CTABanner/CTABanner.module.css
@@ -1,0 +1,57 @@
+.CTABanner {
+  display: flex;
+  gap: var(--base-size-16);
+  flex-direction: column;
+  border-radius: 6px;
+  padding: 40px 32px;
+  /* Box-Shadow */
+  /* box-shadow: 0 0 0 1px rgb(53 72 91 / 14%), 0 2.75px 2.21px rgb(0 0 0 / 7%), 0 6.65px 5.32px rgb(0 0 0 / 4%),
+    0 12.5px 10px rgb(0 0 0 / 3%), 0 22px 18px rgb(0 0 0 / 3%), 0 42px 33.4px rgb(0 0 0 / 2%),
+    0 100px 80px rgb(0 0 0 / 2%) !important; */
+}
+
+.CTABanner:after {
+  position: absolute;
+  content: '';
+  display: block;
+  width: 100%;
+  height: 20px;
+  margin-top: 190px;
+  z-index: -1;
+  background: -webkit-radial-gradient(rgba(0, 0, 0, 0.3), rgba(255, 255, 255, 0));
+  background: -moz-radial-gradient(rgba(0, 0, 0, 0.3), rgba(255, 255, 255, 0));
+  background: -o-radial-gradient(rgba(0, 0, 0, 0.3), rgba(255, 255, 255, 0));
+  background: -ms-radial-gradient(rgba(0, 0, 0, 0.3), rgba(255, 255, 255, 0));
+}
+
+.CTABanner--center {
+  align-items: center;
+  text-align: center;
+}
+
+.CTABanner > * {
+  align-self: start;
+}
+
+.CTABanner--center > * {
+  align-self: center;
+}
+
+.CTABannerDescription {
+  color: var(--brand-color-text-muted);
+  margin-bottom: var(--base-size-16);
+}
+
+/* Medium breakpoint and up */
+@media screen and (min-width: 48rem) {
+  .CTABanner {
+    padding: 80px 64px;
+  }
+}
+
+/* Large breakpoint and up */
+@media screen and (min-width: 63.25rem) {
+  .CTABanner {
+    padding: 128px 64px;
+  }
+}
--- a/packages/react/src/CTABanner/CTABanner.stories.tsx
+++ b/packages/react/src/CTABanner/CTABanner.stories.tsx
@@ -1,0 +1,49 @@
+import React from 'react'
+import {ComponentMeta, ComponentStory} from '@storybook/react'
+import {CTABanner} from '.'
+import {ButtonGroup} from '../ButtonGroup'
+import {Button} from '../Button'
+
+export default {
+  title: 'Components/CTABanner',
+  component: CTABanner,
+  subcomponents: {ButtonGroup},
+  args: {
+    align: 'start'
+  },
+  argTypes: {
+    align: {
+      description: 'The alignment of the content',
+      control: {
+        type: 'radio',
+        options: ['start', 'center']
+      }
+    },
+    children: {
+      table: {
+        disable: true
+      }
+    }
+  }
+} as ComponentMeta<typeof CTABanner>
+
+const Template: ComponentStory<typeof CTABanner> = args => (
+  <CTABanner {...args}>
+    <CTABanner.Heading as={'h1'}>Where the most ambitious teams build great things</CTABanner.Heading>
+    <CTABanner.Description>
+      Lorem ipsum dolor sit amet, consectetur adipiscing elit. In sapien sit ullamcorper id. Aliquam luctus sed turpis
+      felis nam pulvinar risus elementum.
+    </CTABanner.Description>
+    <ButtonGroup buttonSize="large">
+      <Button>Primary Action</Button>
+      <Button>Secondary Action</Button>
+    </ButtonGroup>
+  </CTABanner>
+)
+
+export const Playground = Template.bind({})
+
+// export const LinkButtonGroup = Template.bind({})
+// LinkButtonGroup.args = {
+//   buttonsAs: 'a'
+// }
--- a/packages/react/src/CTABanner/CTABanner.test.tsx
+++ b/packages/react/src/CTABanner/CTABanner.test.tsx
@@ -1,0 +1,55 @@
+import React, {render} from '@testing-library/react'
+import '@testing-library/jest-dom'
+
+import {CTABanner} from './CTABanner'
+import {axe, toHaveNoViolations} from 'jest-axe'
+
+expect.extend(toHaveNoViolations)
+
+describe('CTABanner', () => {
+  it('no a11y violations', async () => {
+    const {container} = render(
+      <CTABanner>
+        <CTABanner.Heading>This is your heading</CTABanner.Heading>
+        <CTABanner.Description>This is your description</CTABanner.Description>
+      </CTABanner>
+    )
+    const results = await axe(container)
+
+    expect(results).toHaveNoViolations()
+  })
+
+  //   it('renders default tag and size correctly', () => {
+  //     const expectedClass = 'Heading--2'
+  //     const expectedTag = 'h2'
+  //     const {getByRole} = render(<Heading>{mockHeading}</Heading>)
+  //     const headingEl = getByRole('heading')
+
+  //     expect(headingEl.tagName).toBe(expectedTag.toUpperCase())
+  //     expect(headingEl.classList).toContain(expectedClass)
+  //   })
+
+  //   it('renders alternative tag correctly with corresponding size', () => {
+  //     const expectedTag = 'h1'
+  //     const expectedClass = 'Heading--1'
+  //     const {getByRole} = render(<Heading as={expectedTag}>{mockHeading}</Heading>)
+  //     const headingEl = getByRole('heading')
+
+  //     expect(headingEl.tagName).toBe(expectedTag.toUpperCase())
+  //     expect(headingEl.classList).toContain(expectedClass)
+  //   })
+
+  //   it('can render a tag with a different visual appearance', () => {
+  //     const expectedTag = 'h1'
+  //     const expectedClass = 'Heading--3'
+  //     const {getByRole} = render(
+  //       <Heading as={expectedTag} size="3">
+  //         {mockHeading}
+  //       </Heading>
+  //     )
+  //     const headingEl = getByRole('heading')
+
+  //     expect(headingEl.tagName).toBe(expectedTag.toUpperCase())
+  //     expect(headingEl.classList).toContain(expectedClass)
+  //   })
+})
--- a/packages/react/src/CTABanner/CTABanner.tsx
+++ b/packages/react/src/CTABanner/CTABanner.tsx
@@ -1,0 +1,57 @@
+import React, {forwardRef, type Ref} from 'react'
+import clsx from 'clsx'
+import type {BaseProps} from '../component-helpers'
+import styles from './CTABanner.module.css'
+import {Heading as PlainHeading} from '../Heading'
+import {Text} from '../Text'
+
+export type CTABannerProps = BaseProps<HTMLDivElement> & {
+  children: React.ReactNode | React.ReactNode[]
+  align?: 'start' | 'center'
+}
+
+export const Root = forwardRef(
+  ({align = 'start', className, children, ...props}: CTABannerProps, ref: Ref<HTMLDivElement>) => {
+    return (
+      <section
+        ref={ref}
+        className={clsx(styles.CTABanner, align === 'center' && styles['CTABanner--center'], className)}
+        {...props}
+      >
+        {children}
+      </section>
+    )
+  }
+)
+
+export const CTABannerHeadingTags = ['h1', 'h2'] as const
+
+export type CTABannerHeadingProps = BaseProps<HTMLHeadingElement> & {
+  children: React.ReactNode | React.ReactNode[]
+  as?: typeof CTABannerHeadingTags[number]
+}
+
+export const Heading = forwardRef(
+  (
+    {as = CTABannerHeadingTags[1], className, children, ...props}: CTABannerHeadingProps,
+    ref: Ref<HTMLHeadingElement>
+  ) => {
+    return <PlainHeading {...props}>{children}</PlainHeading>
+  }
+)
+
+export type CTABannerDescriptionProps = BaseProps<HTMLHeadingElement> & {
+  children: React.ReactNode | React.ReactNode[]
+}
+
+export const Description = forwardRef(
+  ({className, children, ...props}: CTABannerDescriptionProps, ref: Ref<HTMLParagraphElement>) => {
+    return (
+      <Text ref={ref} className={clsx(styles.CTABannerDescription, className)} size={'400'} as="div" {...props}>
+        {children}
+      </Text>
+    )
+  }
+)
+
+export const CTABanner = Object.assign(Root, {Heading, Description})
--- a/packages/react/src/ButtonGroup/ButtonGroup.module.css.d.ts
+++ b/packages/react/src/ButtonGroup/ButtonGroup.module.css.d.ts
@@ -1,6 +1,5 @@
 declare const styles: {
   readonly "ButtonGroup": string;
-  readonly "ButtonGroup--horizontal": string;
 };
 export = styles;
 
--- a/packages/react/src/CTABanner/CTABanner.module.css.d.ts
+++ b/packages/react/src/CTABanner/CTABanner.module.css.d.ts
@@ -1,0 +1,7 @@
+declare const styles: {
+  readonly "CTABanner": string;
+  readonly "CTABanner--center": string;
+  readonly "CTABannerDescription": string;
+};
+export = styles;
+
--- a/packages/react/src/CTABanner/index.ts
+++ b/packages/react/src/CTABanner/index.ts
@@ -1,0 +1,1 @@
+export * from './CTABanner'


JoshBowdenConcepts:
- [ ] Create the background blur to be a circle blurring outwards like in the Figma design

packages/react/src/ButtonGroup/ButtonGroup.tsx
```diff
@@ -20,6 +20,7 @@ export const ButtonGroup = forwardRef(
             size: buttonSize,
             as: buttonsAs,
             variant: index === 0 ? 'primary' : 'secondary',
+            hasArrow: index === 0 ? true : false,
```
enstyled, line 23: FYI, we recently had to update this in PVC to support overwrite, because some designs required both buttons in a group to have arrows.

JoshBowdenConcepts: This supports overwrite but it takes away the default behavior which is to have a arrow.

rezrah, line 23: Possible to add a test around the default behaviour?

packages/react/src/CTABanner/CTABanner.module.css
```diff
@@ -0,0 +1,76 @@
...
+  background: radial-gradient(circle at 50% 50%, rgba(26, 127, 55, 1), rgba(26, 127, 55, 0));
+}
+
+.CTABanner--LeftShadow {
+  left: 70%;
+  background: radial-gradient(circle at 50% 50%, rgba(3, 102, 214, 1), rgba(3, 102, 214, 0));
+}
+
+.CTABanner--content {
+  border-radius: 6px;
```
rezrah, line 32:
```suggestion
  border-radius: var(--brand-borderRadius-medium)
```

packages/react/src/CTABanner/CTABanner.module.css
```diff
@@ -0,0 +1,76 @@
...
+}
+
+.CTABanner--RightShadow {
+  left: 30%;
+  background: radial-gradient(circle at 50% 50%, rgba(26, 127, 55, 1), rgba(26, 127, 55, 0));
+}
+
+.CTABanner--LeftShadow {
+  left: 70%;
+  background: radial-gradient(circle at 50% 50%, rgba(3, 102, 214, 1), rgba(3, 102, 214, 0));
```
rezrah, line 28: We should use the css variables here if possible, which would also apply the correct values for light **_and_** dark modes (guessing these are light?)

gradients for light and dark mode

For now, I think you can resolve this by swapping to the gradient tokens directly, and in a separate PR we can create the required tokens for all gradients and backfill.

packages/react/src/CTABanner/CTABanner.module.css
```diff
@@ -0,0 +1,76 @@
...
+}
+
+.CTABanner--content {
+  border-radius: 6px;
+  padding: 40px 32px;
+  z-index: 1;
+  background: white;
+}
+
+.CTABanner--content--noshadow {
```
rezrah, line 38: Is there an extra dash here?

```suggestion
.CTABanner-content--noshadow {
```

packages/react/src/CTABanner/CTABanner.module.css
```diff
@@ -0,0 +1,76 @@
...
+
+.CTABanner > * {
+  align-self: start;
+}
+
+.CTABanner--center > * {
+  align-self: center;
+}
+
+.CTABannerDescription {
```
rezrah, line 59:
```suggestion
.CTABanner-description {
```

packages/react/src/CTABanner/CTABanner.module.css
```diff
@@ -0,0 +1,76 @@
...
+  border: 1px solid var(--brand-color-border-default);
+}
+
+.CTABanner--center {
+  align-items: center;
+  text-align: center;
+}
+
+.CTABanner > * {
+  align-self: start;
```
rezrah, line 52: Can you use `align-items` instead of the wildcard selector?

packages/react/src/CTABanner/CTABanner.module.css
```diff
@@ -0,0 +1,76 @@
...
+@media screen and (min-width: 48rem) {
+  .CTABanner--content {
+    padding: 80px 64px;
+  }
+}
+
+/* Large breakpoint and up */
+@media screen and (min-width: 63.25rem) {
+  .CTABanner--content {
+    padding: 128px 64px;
```
rezrah, line 74:
```suggestion
    padding: var(--base-size-128) var(--base-size-64);
```

packages/react/src/CTABanner/CTABanner.module.css
```diff
@@ -0,0 +1,76 @@
...
+
+.CTABannerDescription {
+  color: var(--brand-color-text-muted);
+  margin-bottom: var(--base-size-16);
+}
+
+/* Medium breakpoint and up */
+@media screen and (min-width: 48rem) {
+  .CTABanner--content {
+    padding: 80px 64px;
```
rezrah, line 67:
```suggestion
    padding: var(--base-size-80) var(--base-size-64);
```

packages/react/src/CTABanner/CTABanner.module.css
```diff
@@ -0,0 +1,76 @@
...
+  }
+}
+
+/* Large breakpoint and up */
+@media screen and (min-width: 63.25rem) {
+  .CTABanner--content {
```
rezrah, line 73: content feels like the element rather than modifier, or have I misread it?

```suggestion
  .CTABanner-content {
```

packages/react/src/CTABanner/CTABanner.tsx
```diff
@@ -0,0 +1,89 @@
...
+  ) => {
+    return (
+      <section
+        ref={ref}
+        className={clsx(styles.CTABanner, align === 'center' && styles['CTABanner--center'])}
+        {...props}
+      >
+        {hasShadow && (
+          <>
+            <div className={clsx(styles['CTABanner--Shadow'], styles['CTABanner--RightShadow'])} />
```
rezrah, line 37: As these are presentation only, can you add `aria-hidden` to them?

rezrah: Alternatively, can they be done through CSS pseudo elements?

packages/react/src/CTABanner/CTABanner.tsx
```diff
@@ -0,0 +1,89 @@
...
+            <div className={clsx(styles['CTABanner--Shadow'], styles['CTABanner--LeftShadow'])} />
+          </>
+        )}
+        <div
+          className={clsx(
+            styles['CTABanner--content'],
+            hasBorder && styles['CTABanner--border'],
```
rezrah, line 44: Left a comment about this in the stylesheet, but assuming we're following the BEM approach for naming this should be probably be..

```suggestion
            hasBorder && styles['CTABanner-content--border'],
```

packages/react/src/CTABanner/CTABanner.tsx
```diff
@@ -0,0 +1,89 @@
+import React, {forwardRef, type Ref} from 'react'
+import clsx from 'clsx'
+import type {BaseProps} from '../component-helpers'
+import styles from './CTABanner.module.css'
+import {Heading as PlainHeading} from '../Heading'
```
rezrah, line 5: Recommend we don't rename this to something else entirely.. you can keep as is and prefix the internal version.

https://github.com/primer/brand/pull/128/files#r1025088323

packages/react/src/CTABanner/CTABanner.tsx
```diff
@@ -0,0 +1,89 @@
...
+  as?: typeof CTABannerHeadingTags[number]
+}
+
+export const Heading = forwardRef(
+  (
+    {as = CTABannerHeadingTags[1], className, children, ...props}: CTABannerHeadingProps,
+    ref: Ref<HTMLHeadingElement>
+  ) => {
+    return (
+      <PlainHeading {...props} className={className} as={as}>
```
rezrah, line 68: Are we missing the forwarded ref here?

packages/react/src/CTABanner/CTABanner.tsx
```diff
@@ -0,0 +1,89 @@
...
+        {children}
+      </PlainHeading>
+    )
+  }
+)
+
+export type CTABannerDescriptionProps = BaseProps<HTMLHeadingElement> & {
+  children: React.ReactNode | React.ReactNode[]
```
rezrah, line 76: I think  `PropsWithChildren` covers this? 🤔

packages/react/src/CTABanner/CTABanner.tsx
```diff
@@ -0,0 +1,89 @@
...
+}
+
+export const Description = forwardRef(
+  ({className, children, ...props}: CTABannerDescriptionProps, ref: Ref<HTMLParagraphElement>) => {
+    return (
+      <Text ref={ref} className={clsx(styles.CTABannerDescription, className)} size={'400'} as="div" {...props}>
```
rezrah, line 82: Can we make the body text a paragraph instead? and wrap it in a div if you need? Better semantics.

Also the ref that you accept is a paragraph, which is different to HTMLDivElement

packages/react/src/CTABanner/CTABanner.module.css
JoshBowdenConcepts: All set thanks for the help here

packages/react/src/ButtonGroup/ButtonGroup.tsx
JoshBowdenConcepts: Yes I will add a test specifically for this on this component and the button group

packages/react/src/CTABanner/CTABanner.module.css
```diff
@@ -0,0 +1,82 @@
...
+  background: white;
+  align-items: start;
+}
+
+.CTABanner-content--noshadow {
+  background: none;
+}
+
+.CTABanner-content--border {
+  border: 1px solid var(--brand-color-border-default);
```
rezrah, line 53:
```suggestion
  border: var(--brand-borderWidth-thin) solid var(--brand-color-border-default);
```

packages/react/src/CTABanner/CTABanner.module.css
```diff
@@ -0,0 +1,82 @@
...
+}
+
+.CTABanner-content--center {
+  align-items: center;
+  text-align: center;
+}
+
+.CTABanner--center > * {
```
rezrah, line 61: Do you still need this, given the rule about it?

packages/react/src/CTABanner/CTABanner.stories.tsx
```diff
@@ -0,0 +1,80 @@
...
+    </CTABanner.Description>
+    <ButtonGroup buttonSize="large">
+      <Button>Primary Action</Button>
+      <Button>Secondary Action</Button>
+    </ButtonGroup>
+  </CTABanner>
+)
+
+export const Playground = Template.bind({})
+
```
rezrah, line 66: Can you add a Default story too please? Can be same as Playground.

packages/react/src/CTABanner/CTABanner.module.css
```diff
@@ -0,0 +1,82 @@
...
+  padding: var(--base-size-40) var(--base-size-32);
+  z-index: 1;
+  display: flex;
+  flex-direction: column;
+  gap: var(--base-size-16);
+  background: white;
+  align-items: start;
+}
+
+.CTABanner-content--noshadow {
```
rezrah, line 48: should this be `.CTABanner-content--no-background`?

packages/react/src/CTABanner/CTABanner.module.css
```diff
@@ -0,0 +1,82 @@
...
+}
+
+.CTABanner-content {
+  border-radius: var(--brand-borderRadius-medium);
+  padding: var(--base-size-40) var(--base-size-32);
+  z-index: 1;
+  display: flex;
+  flex-direction: column;
+  gap: var(--base-size-16);
+  background: white;
```
rezrah, line 44: Don't think this will work on dark mode..

```suggestion
  background: var(--brand-color-canvas-default);
```

packages/react/src/CTABanner/CTABanner.module.css
```diff
@@ -0,0 +1,82 @@
+.CTABanner {
+  position: relative;
+  max-width: 1280px;
+  margin: 0 auto;
+  display: flex;
+}
+
+.CTABanner--Shadow {
```
rezrah, line 8: is there a reason for Shadow being uppercased here?

```suggestion
.CTABanner--shadow{
```

Alternatively, how about naming it after the prop to make it easier to correlate the prop with the class.

```suggestion
.CTABanner--hasShadow {
```

packages/react/src/CTABanner/CTABanner.tsx
```diff
@@ -0,0 +1,93 @@
...
+      children,
+      ...props
+    }: CTABannerProps,
+    ref: Ref<HTMLDivElement>
+  ) => {
+    return (
+      <section ref={ref} className={styles.CTABanner} {...props}>
+        {hasShadow && (
+          <>
+            <div aria-hidden className={clsx(styles['CTABanner--Shadow'], styles['CTABanner--StartShadow'])} />
```
rezrah, line 40:
```suggestion
            <div aria-hidden className={clsx(styles['CTABanner-shadow'], styles['CTABanner-shadow--start'])} />
```

packages/react/src/CTABanner/CTABanner.tsx
```diff
@@ -0,0 +1,93 @@
...
+            <div aria-hidden className={clsx(styles['CTABanner--Shadow'], styles['CTABanner--EndShadow'])} />
```
rezrah, line 41:
```suggestion
            <div aria-hidden className={clsx(styles['CTABanner-shadow'], styles['CTABanner-shadow--end'])} />
```

packages/react/src/CTABanner/CTABanner.tsx
rezrah: ref should be part of BaseProps

rezrah (commented):
Looking good..

Few general comments:

- The heading and description seem to have a max-width in the designs, that isn't accounted for in code. It's particularly noticeable on the description.

  **Figma:**  
  Screenshot 2022-11-21 at 20 25 04

  **Code:** 
  Screenshot 2022-11-21 at 20 27 18

- We're missing a docs page for this, can you add one in please? You can find the mdx files in `apps/docs/content`

- We're missing visual snapshots, so let me know just before you finally merge this PR and I'll add them in.

- Can you add an extra story to show how you can customise the colors of the shadows to something different using css variables?

packages/react/src/CTABanner/CTABanner.module.css
```diff
@@ -0,0 +1,82 @@
...
+  height: calc(100% + 60px);
+  border-radius: 50% 22% 40% 80%;
+  filter: blur(50px);
+  top: 0%;
+  opacity: 0.7;
+}
+
+.CTABanner--StartShadow {
+  left: 30%;
+  background: radial-gradient(
```
rezrah, line 21: Shadow feels a bit faint compared to the Figma design. @danielguillan can you please help confirm we're using the right values here?

packages/react/src/CTABanner/CTABanner.stories.tsx
JoshBowdenConcepts: Oh I thought you mentioned last time that playground was the new name for the default playground. Wasn't this mentioned in the ADR

packages/react/src/CTABanner/CTABanner.stories.tsx
```diff
@@ -0,0 +1,80 @@
...
+export const CTABannerWithBorder = Template.bind({})
```
rezrah, line 67: Can we shorten these names a bit? 
Screenshot 2022-11-21 at 20 37 26



either 

```jsx
export const WithBorder = Template.bind({})
```

or `CTABannerWithBorder.storyName = 'with border'

rezrah, line 67: Based on ADR, all stories except Playground should go in a few file called `CTABanner.features.stories.tsx` under a new folder called Features.

packages/react/src/CTABanner/CTABanner.stories.tsx
rezrah: The ADR recommended that Playground is the one that lets you experiment with controls.. and Default is mandatory as the main code snippet example.

packages/react/src/CTABanner/CTABanner.module.css
danielguillan: That's correct. Building the shadow using two circular gradients is causing this problem. Additionally, the shadow doesn't extend properly around the corners, making these almost fade into the background.

Comparison of the current shadow rendering vs. the Figma component

I think we can get pretty close to the rendering in Figma by using a single rectangular element and a linear gradient. I'm not sure if we've already tried something like this:

```css
.CTABanner::before {
  content: '';
  position: absolute;
  inset: var(--base-size-16);
  filter: blur(var(--base-size-96));
  transform: translateZ(0); /* Force GPU to fix blurring in Safari */
  background: linear-gradient(
    to right,
    hsla(var(--brand-CTABanner-shadow-color-start) / 1),
    hsla(var(--brand-CTABanner-shadow-color-end) / 1)
  ); /* We no longer need the alpha channel on the colors. */
}
```

_Note: I've used a pseudo-element here, but we can still use a dedicated element._


Comparison of the current shadow rendering vs. the Figma component


It's not 1:1 with Figma (the colors are slightly duller and darker) but I think it's close enough.

packages/react/src/CTABanner/CTABanner.tsx
JoshBowdenConcepts: When I remove the children it creates a type error which makes it no work

packages/react/src/CTABanner/CTABanner.module.css
JoshBowdenConcepts: Removing in next push

packages/react/src/CTABanner/CTABanner.stories.tsx
JoshBowdenConcepts: The first story is the one that has the controls. Are you saying the ideal solution would be to have Default after this?

packages/react/src/CTABanner/CTABanner.tsx
JoshBowdenConcepts: This may be different in next push if we go with the generic box shadow

packages/react/src/CTABanner/CTABanner.tsx
```diff
@@ -0,0 +1,93 @@
...
+  ) => {
+    return (
+      <Heading ref={ref} {...props} className={className} as={as}>
+        {children}
+      </Heading>
+    )
+  }
+)
+
+export type CTABannerDescriptionProps = BaseProps<HTMLHeadingElement> & {
```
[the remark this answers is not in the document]
JoshBowdenConcepts: In next push

packages/react/src/CTABanner/CTABanner.module.css
JoshBowdenConcepts: I will update it to this aapproach.

packages/react/src/CTABanner/CTABanner.test.tsx
```diff
@@ -0,0 +1,152 @@
...
+    expect(ctaBannerEl).not.toHaveClass(classToCheck)
+  })
+
+  it('renders the CTABanner with a border', () => {
+    const mockTestId = 'test'
+    const classToCheck = 'CTABanner-content--border'
+
+    const {getByTestId} = render(
+      <CTABanner hasBorder data-testid={mockTestId}>
+        <CTABanner.Heading>{'This is your heading'}</CTABanner.Heading>
```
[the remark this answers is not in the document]
JoshBowdenConcepts: No this is a previous pattern that was enforced at my last company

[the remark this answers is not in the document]
JoshBowdenConcepts: Removing for next push

packages/react/src/CTABanner/CTABanner.stories.tsx
JoshBowdenConcepts: In next push

JoshBowdenConcepts: In next push

JoshBowdenConcepts:
@danielguillan looking at @rezrah comment about the max-width for the text within. This isn't something that seems to be presented correctly in the design. Are there specific values you would like me to supply? The padding is what I would imagine would handle these widths as it is doing now.

packages/react/src/CTABanner/CTABanner.module.css
JoshBowdenConcepts: no, I will remove in next push

packages/react/src/CTABanner/CTABanner.module.css
JoshBowdenConcepts: updated

danielguillan:
@JoshBowdenConcepts @rezrah as discussed, the designs didn't have a `max-width` defined. Instead, there was a manual, forced line break that achieved this effect. I think we can use a `max-width` of `800px` here while we work on supporting maximum line lengths inside our typographic tokens.

apps/docs/content/components/CTABanner.mdx
```diff
@@ -0,0 +1,205 @@
...
+<CTABanner>
+  <CTABanner.Heading>
+    Where the most ambitious teams build great things
+  </CTABanner.Heading>
+  <CTABanner.Description>
+    Lorem ipsum dolor sit amet, consectetur adipiscing elit. In sapien sit
+    ullamcorper id. Aliquam luctus sed turpis felis nam pulvinar risus
+    elementum.
+  </CTABanner.Description>
+  <ButtonGroup buttonSize="large">
```
rezrah, line 28: should `buttonSize` be configurable @danielguillan? If not, i'd recommend we set this by default inside CTABanner. Buttons feel a bit large to me, so wondering if it should be `medium` too.

apps/docs/content/components/CTABanner.mdx
```diff
@@ -0,0 +1,205 @@
...
+    <Button>Primary Action</Button>
+    <Button>Secondary Action</Button>
+  </ButtonGroup>
+</CTABanner>
+```
+
+### Aligned center
```
rezrah, line 35: For consistency with [Hero](https://primer.style/brand/components/Hero) and other doc pages... 

```suggestion
### Alignment
```

apps/docs/content/components/CTABanner.mdx
```diff
@@ -0,0 +1,205 @@
...
+
+The content alignment can be changed using the `align` prop on the root `<ComparisonTable>`. This can be configured to either `start` or `center`.
```
rezrah, line 37: ComparisonTable? 😅 

```suggestion
The content alignment can be changed using the `align` prop on the root `CTABanner`. This can be configured to either `start` or `center`.
```

apps/docs/content/components/CTABanner.mdx
```diff
@@ -0,0 +1,205 @@
...
+  <ButtonGroup buttonSize="large">
+    <Button>Primary Action</Button>
+    <Button>Secondary Action</Button>
+  </ButtonGroup>
+</CTABanner>
+```
+
+### Optional border
+
+A border can be provided using the `hasBorder` prop on the root `<ComparisonTable>`. This will render a border around the CTABanner component giving separatoin between the foreground and background. Especially when there is no shadow present on the background.
```
rezrah, line 58: typo.. and reference to `ComparisonTable`

```suggestion
A border can be provided using the `hasBorder` prop on the root `<CTABanner>`. This will render a border around the CTABanner component giving further separation between the foreground and background. Especially when there is no shadow present on the background.
```

> Especially when there is no shadow present on the background.

Worth adding hasShadow={false} here to demonstrate this?

apps/docs/content/components/CTABanner.mdx
```diff
@@ -0,0 +1,205 @@
...
+  </ButtonGroup>
+</CTABanner>
+```
+
+### Optional background
+
+A background can be removed using the `hasBackground` prop on the root `<ComparisonTable>`. This will render the CTABanner component without a background allowing it to blend in with its parent.
+
+```jsx live
+<CTABanner hasBackground={false}>
```
rezrah, line 82: example lacking contrast between foreground and background

Not sure if this is the example I'd expect given the heading. 

Can we either provide a custom image or background color (maybe something from base scale) in the background or change the title to say "remove background fill" or similar. cc. @danielguillan for other ideas.

Also that shadow should probably be turned off here, it's a bit much 😀

apps/docs/content/components/CTABanner.mdx
```diff
@@ -0,0 +1,205 @@
...
+  <ButtonGroup buttonSize="large">
+    <Button>Primary Action</Button>
+    <Button>Secondary Action</Button>
+  </ButtonGroup>
+</CTABanner>
+```
+
+### Optional shadow
+
+The shadow can be removed using the `hasShadow` prop on the root `<ComparisonTable>`. This will render the CTABanner component without the background shaow.
```
rezrah, line 100: typo's and another reference to `ComparisonTable`

```suggestion
The shadow can be removed using the `hasShadow` prop on the root `CTABanner`. This will render the component without the background shadow.
```

```

File: apps/docs/content/components/CTABanner.mdx [hunk 34]
```diff
@@ -0,0 +1,205 @@
...
+```javascript live noinline
+const App = () => {
+  const designTokenOverrides = `
+  .custom-colors[data-color-mode='dark'] {
+
+    /*
+     * Modify the value of these tokens.
+     * Remember to apply light mode equivalents if you're enabling theme switching.
+     */
+    --brand-CTABanner-shadow-color-start: var(--base-color-scale-purple-5-hsl);
```
rezrah, line 137: just checking that these are our official gradient pairings?

apps/docs/content/components/CTABanner.mdx
```diff
@@ -0,0 +1,205 @@
...
+    </ThemeProvider>
+  )
+}
+
+render(<App />)
+```
+
+## Component props
+
+### ComparisonTable <Label>Required</Label>
```
rezrah, line 168: Heads up.. Label isn't working here. You might need to import it.
Screenshot 2022-11-30 at 12 55 40

apps/docs/content/components/CTABanner.mdx
```diff
@@ -0,0 +1,205 @@
...
+
+The first row in a table is automatically styled as a table header row.
+
+| name            | type                       | default   | required | description                                                  |
+| --------------- | -------------------------- | --------- | -------- | ------------------------------------------------------------ |
+| `children`      | `ReactNode`, `ReactNode[]` | undefined | `true`   | Content to include within the banner component               |
```
rezrah, line 174: syntax highlighting on undefined...

```suggestion
`| `children`      | `ReactNode`, `ReactNode[]` | `undefined` | `true`   | Content to include within the banner component               |`
```

packages/react/src/CTABanner/CTABanner.tsx
```diff
@@ -0,0 +1,88 @@
...
+          <div className={clsx(styles['CTABanner-content'], align === 'center' && styles['CTABanner-content--center'])}>
+            {children}
+          </div>
+        </div>
+      </section>
+    )
+  }
+)
+
+export const CTABannerHeadingTags = ['h1', 'h2'] as const
```
rezrah, line 54: I don't think this would ever be used as a H1 or H2 given where we expect it to appear (near the bottom of the page)... 

Is there a reason for constraining it, rather than forwarding the types from `Heading`? I think setting some defaults is fine, but we should also allow app-side customisation. We've received related feedback around the River.

apps/docs/content/components/CTABanner.mdx
```diff
@@ -0,0 +1,205 @@
...
+
+### ComparisonTable.Heading
+
+Optional node that can be used to provide a heading for the banner.
+
+| name        | type                       | default   | required | description                                                           |
+| ----------- | -------------------------- | --------- | -------- | --------------------------------------------------------------------- |
+| `className` | `string`                   |           | `false`  | Custom class name for the ComparisonTable.Heading component           |
+| `children`  | `ReactNode`, `ReactNode[]` | undefined | `true`   | Content to be displayed inside the ComparisonTable.Heading component. |
+| `as`        | `h1`, `h2`                 | `h1`      | `false`  | Content to be displayed inside the column cell.                       |
```
rezrah, line 188: related feedback https://github.com/primer/brand/pull/128/files#r1035940710

apps/docs/content/components/CTABanner.mdx
```diff
@@ -0,0 +1,205 @@
+---
+title: CTABanner
+status: Experimental
+figma: 'https://www.figma.com/file/BJ95AjraesmRCWsKA013GS/Primer-Brand?node-id=539%3A14285&t=RJEWZhVUtrRQ0pKh-0'
+source: https://github.com/primer/brand/tree/main/src/CTABanner/CTABanner.tsx
+storybook: '/brand/storybook/?path=/docs/components-ctabanner--playground'
+description: Use CTABanner to create call to action content in a separated section.
```
rezrah, line 7: How about this..

"Use CTABanner to draw attention to or create urgency around a user action"

cc. @ajashams @tobiasahlin @danielguillan for their suggestions also.

apps/docs/content/components/CTABanner.mdx
```diff
@@ -0,0 +1,205 @@
+---
+title: CTABanner
+status: Experimental
+figma: 'https://www.figma.com/file/BJ95AjraesmRCWsKA013GS/Primer-Brand?node-id=539%3A14285&t=RJEWZhVUtrRQ0pKh-0'
+source: https://github.com/primer/brand/tree/main/src/CTABanner/CTABanner.tsx
+storybook: '/brand/storybook/?path=/docs/components-ctabanner--playground'
```
rezrah, line 6: This links to Docs tab, can we change this to Canvas please? We don't use the former.

```suggestion
storybook: '/brand/storybook/?path=/story/components-ctabanner--playground'
```

```

File: packages/react/src/CTABanner/CTABanner.test.tsx [hunk 41]
```diff
@@ -0,0 +1,152 @@
...
+        </ButtonGroup>
+      </CTABanner>
+    )
+    const ctaBannerEl = getByTestId(mockTestId)
+    expect(ctaBannerEl.tagName).toBe(expectedTag.toUpperCase())
+    expect(ctaBannerEl.classList).toContain(expectedClass)
+  })
+
+  it('renders the correct default heading type', () => {
+    const expectedTag = 'h2'
```
rezrah, line 50: related comment https://github.com/primer/brand/pull/128/files#r1035940710

apps/docs/content/components/CTABanner.mdx
JoshBowdenConcepts: Oh yes so after the chat we had last week the docs were supposed to only show this with the shadow and background removed. So I will also remove the shadow. We discussed linking the two potentially if that proves to be how it is used most.

apps/docs/content/components/CTABanner.mdx
JoshBowdenConcepts: When I was working with Ash he said to use the large buttons medium is set by default. This is why it was created this way.

apps/docs/content/components/CTABanner.mdx
JoshBowdenConcepts: I pulled it from the gradient file in Figma

packages/react/src/CTABanner/CTABanner.tsx
JoshBowdenConcepts: This was the feedback given to me by Ash. We can open it up for sure. It was that way before and he Ash had me restrict it.

apps/docs/content/components/CTABanner.mdx
JoshBowdenConcepts: Sounds much better to me.

apps/docs/content/components/CTABanner.mdx
```diff
@@ -0,0 +1,205 @@
...
+</CTABanner>
+```
+
+#### Color customization
+
+The default colors can be customized through dedicated CSS variables.
+
+| Name                                   | Default                          |
+| :------------------------------------- | :------------------------------- |
+| `--brand-CTABanner-shadow-color-start` | `--base-color-scale-green-5-hsl` |
```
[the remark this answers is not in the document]
JoshBowdenConcepts: I like default because it still tells the user what it is currently set to and is consistant with our other pages.

packages/react/src/CTABanner/CTABanner.tsx
danielguillan: I think using an `h3` by default would make sense, as the banner will usually be part of a section on the page, which will likely already have an `h2`.

JoshBowdenConcepts: Ok this will change the size of the text within the banner. Should I make it visually something else?

apps/docs/content/components/CTABanner.mdx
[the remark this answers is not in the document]
JoshBowdenConcepts: That makes sense I could add dark mode to it to be clear to users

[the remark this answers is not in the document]
JoshBowdenConcepts: These docs will be in the next push

packages/react/src/CTABanner/CTABanner.tsx
```diff
@@ -0,0 +1,88 @@
...
+    {
+      align = 'start',
+      hasBorder = false,
+      hasShadow = true,
+      hasBackground = true,
+      className,
+      children,
+      ...props
+    }: CTABannerProps,
+    ref: Ref<HTMLDivElement>
```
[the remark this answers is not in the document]
JoshBowdenConcepts: This actually doesn't appear to be a type.

apps/docs/content/components/CTABanner.mdx
JoshBowdenConcepts: Other than the default which we will discuss in maintainers sync I updated it to include all options.

packages/react/src/CTABanner/CTABanner.tsx
rezrah: yeah, you can use the `size` prop to visually lock to "2" (or whatever it was before)

apps/docs/content/components/CTABanner.mdx
rezrah: > When I was working with Ash he said to use the large buttons medium is set by default. This is why it was created this way.

That's fine, if large is default, can you set it _inside_ the component and remove from the docs examples? This is part of design logic, so you'd also probably want a test around that. I'd expect the only time you'd need to manually the buttonSize prop, is to change it to 'medium'.

JoshBowdenConcepts:
- [ ] ButtonGroup within the CTABanner should use the name-spacing method as the heading uses so you won't need to pass button size.
- [ ] Semantic heading should be h3 by default and appear as it does now.
- Button size will be large the size issue will be handled by the changes to the Button component itself

packages/react/src/CTABanner/CTABanner.tsx
```diff
@@ -0,0 +1,99 @@
...
+/**
+ * Design tokens
+ */
+import '@primer/brand-primitives/lib/design-tokens/css/tokens/functional/components/cta-banner/colors-with-modes.css'
+
+/** * Main Stylesheet (as a CSS Module) */
+import styles from './CTABanner.module.css'
+
+export type CTABannerProps = BaseProps<HTMLDivElement> & {
+  children: React.ReactNode | React.ReactNode[]
```
rezrah, line 17: Can we add JSDoc annotations to these props? [Example from SubdomainNavBar](https://github.com/primer/brand/blob/f41b6cf4c020e67dc94cce8c6816d68ebaeb7b75/packages/react/src/SubdomainNavBar/SubdomainNavBar.tsx#L17). 

These are really useful for app devs, as IDE's like VSCode will forward them to all their references on hover. I will usually reuse the description in the docs prop table too, so it's helpful in other ways also.

packages/react/src/CTABanner/CTABanner.tsx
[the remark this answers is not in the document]
rezrah: 🤔 okay, can we just change it to HTMLElement instead then? As that's the catch all. Using a `div` one looks wrong here when the tag is a section, as pedantic as it sounds 😅 .

apps/docs/content/components/CTABanner.mdx
rezrah: Unresolving as it's still broken.. 

https://primer-25ab31e23f-26139705.drafts.github.io/components/CTABanner#comparisontable-required

Also says ComparisonTable?

```suggestion
### CTABanner <Label>Required</Label>
```

apps/docs/content/components/CTABanner.mdx
```diff
@@ -0,0 +1,205 @@
...
+
+| name            | type                       | default   | required | description                                                  |
+| --------------- | -------------------------- | --------- | -------- | ------------------------------------------------------------ |
+| `children`      | `ReactNode`, `ReactNode[]` | undefined | `true`   | Content to include within the banner component               |
+| `align`         | `start`, `center`          | `start`   | `true`   | The layout applied to the content wihin the banner component |
+| `hasBorder`     | `boolean`                  | `false`   | `false`  | A flag used to provide a border to the banner component      |
+| `hasBackground` | `boolean`                  | `true`    | `false`  | A flag used to add a background to the banner component      |
+| `className`     | `string`                   |           | `false`  | CTABanner custom class                                       |
+
+### ComparisonTable.Heading
```
rezrah, line 180:
```suggestion
### CTABanner.Heading
```

rezrah:
This is looking good! Some additional feedback below:

- Can you add a changeset? Currently missing one.
- Your CRA integration test is failing. Is there some logging that's been left in there?

packages/react/src/CTABanner/CTABanner.tsx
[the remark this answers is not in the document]
JoshBowdenConcepts: I think we might be better off getting as close as possible although the use case is unlikely to be used I would imaging giving more support for these kinds of uses would be better than less.
- https://developer.mozilla.org/en-US/docs/Web/API/HTMLDivElement

packages/react/src/CTABanner/CTABanner.tsx
JoshBowdenConcepts: Sounds good to me!

danielguillan (approved):

.changeset/shy-boxes-relate.md
```diff
@@ -0,0 +1,6 @@
+---
+'@primer/react-brand': patch
```
rezrah, line 2: We use `minor` when adding a new component.

packages/react/src/SubdomainNavBar/SubdomainNavBar.module.css
```diff
@@ -220,7 +220,7 @@
 }
 
 @media screen and (max-width: 767px) {
-  .SubdomainNavBar-link:first-child {
+  .SubdomainNavBar-link:first-of-type {
```
rezrah, line 223: Nice catch! Any regressions from this? Guessing not 🤔

.changeset/shy-boxes-relate.md
```diff
@@ -0,0 +1,24 @@
...
+    elementum.
+  </CTABanner.Description>
+  <CTABanner.ButtonGroup>
+    <Button>Primary Action</Button>
+    <Button>Secondary Action</Button>
+  </CTABanner.ButtonGroup>
+</CTABanner>
+```
+
+:link: [See the documentation for more details and usage examples.](https://primer.style/brand/components/ComparisonTable)
```
rezrah, line 24: 😅 

```suggestion
:link: [See the documentation for more details and usage examples.](https://primer.style/brand/components/CTABanner)
```

apps/docs/content/components/CTABanner.mdx
```diff
@@ -0,0 +1,219 @@
...
+    </ThemeProvider>
+  )
+}
+
+render(<App />)
+```
+
+## Component props
+
+### CTABanner <Label>Required</Label>
```
rezrah, line 168: My previous comments about this seem to have disappeared 😕 

This is breaking in the rendered version because Label isn't imported.

fix is - from memory (try it locally) - to add `import {Label} from "@primer/react"` at the top of the file. 

Screenshot 2022-12-07 at 13 38 53

packages/react/src/SubdomainNavBar/SubdomainNavBar.module.css
JoshBowdenConcepts: There seems to be a test breaking due to this change.

.changeset/shy-boxes-relate.md
JoshBowdenConcepts: Weird I fixed this when I updated the PR and it didn't commit.

apps/docs/content/components/CTABanner.mdx
JoshBowdenConcepts: Ahh I thought I fixed this but I imported it from primer-brand which didn't make sense.

packages/react/src/SubdomainNavBar/SubdomainNavBar.module.css
JoshBowdenConcepts: But it isn't breaking locally. It said accessibility issue in description within the breaking CI pipeline

apps/docs/content/components/CTABanner.mdx
JoshBowdenConcepts: This will be in the next push

apps/docs/content/components/CTABanner.mdx
```diff
@@ -0,0 +1,220 @@
...
+status: Experimental
+figma: 'https://www.figma.com/file/BJ95AjraesmRCWsKA013GS/Primer-Brand?node-id=539%3A14285&t=RJEWZhVUtrRQ0pKh-0'
+source: https://github.com/primer/brand/tree/main/src/CTABanner/CTABanner.tsx
+storybook: '/brand/storybook/?path=/story/components-ctabanner--playground'
+description: Use CTABanner to draw attention to or create urgency around a user action.
+---
+
+```js
+import {CTABanner, Button} from '@primer/react-brand'
+import {Label} from '@primer/react'
```
rezrah, line 12: This needs to be outside the code snippet scope.. it's currently rendering in the docs..

Screenshot 2022-12-07 at 15 12 21

rezrah (approved):

PR #145, opened 2022-12-14
OrderedList and UnorderedList Components

JoshBowdenConcepts:
image

## Summary

Two list component 'unordered' and 'ordered' 

## List of notable changes:

- **added** UnorderedList component
- **added** OrderedList component

- Making sure the components work as expected

## Steps to test:

1. Open the storybook instance in CI-deployed preview environment
2. Go to both UnorderedList and OrderedList stories in Storybook
3. Verify that both components behaves as described.

- [Design Ticket](https://github.com/github/primer/issues/1587)
- [Component Epic](https://github.com/github/primer/issues/1352)

## Contributor checklist:

## Reviewer checklist:

## Preview Deployments:
- [Markdown Docs](https://primer-1e22b62640-26139705.drafts.github.io/)
- [Storybook](https://primer-1e22b62640-26139705.drafts.github.io/storybook)


.changeset/blue-dots-battle.md
```diff
@@ -0,0 +1,27 @@
+---
+"@primer/react-brand": minor
+---
+
+Adding two components, OrderedList and UnorderedList
```
rezrah, line 5: For correct tense, as this is used post-release..

```suggestion
Added `OrderedList` and `UnorderedList`
```

apps/docs/content/components/List.mdx
```diff
@@ -0,0 +1,65 @@
+---
+title: List
```
rezrah, line 2: Can we have two different doc pages for each component instead of one. `List` isn't a component so this is somewhat confusing.

apps/docs/content/components/List.mdx
```diff
@@ -0,0 +1,65 @@
...
+status: Experimental
+figma: 'https://www.figma.com/file/BJ95AjraesmRCWsKA013GS/Primer-Brand?node-id=354%3A9513'
+source: https://github.com/primer/brand/tree/main/src/List/UnorderedList.tsx
```
rezrah, line 5: Related to previous comment, this refers to UnorderedList but leaves out OrderedList, they should be different doc pages.

apps/docs/content/components/List.mdx
```diff
@@ -0,0 +1,65 @@
...
+storybook: '/brand/storybook/?path=/story/components-unorderedlist--scale'
```
rezrah, line 6: what's scale? this should ideally link to `default` or `playground`

apps/docs/content/components/List.mdx
```diff
@@ -0,0 +1,65 @@
...
+description: Use the List component to a list of items. You can use either the OrderedList or the UnorderedList component.
```
rezrah, line 7: > Use the List component to a list of items.

is this missing a verb like render?

.changeset/blue-dots-battle.md
```diff
@@ -0,0 +1,27 @@
...
+
+```
+<OrderedList>
+    <li>Automatic security and version updates</li>
+    <li>GitHub Security Advisories</li>
+    <li>Code and secret scanning</li>
+    <li>Dependency review</li>
+    <li>Automated authentication and identity management</li>
+</OrderedList>
+```
```
rezrah, line 27: Can you link to the docs page from here too please? 

Good examples in the previous changesets for new components.

.changeset/blue-dots-battle.md
```diff
@@ -0,0 +1,27 @@
...
+"@primer/react-brand": minor
+---
+
+Added `OrderedList` and `UnorderedList`
+
+Example Usage:
+
+```
+<UnorderedList variant="checked">
+    <li>Automatic security and version updates</li>
```
rezrah, line 11: This is using the <li> API?

packages/react/src/List/ListItem.tsx
```diff
@@ -0,0 +1,40 @@
...
+   * React.ReactNode and React.ReactNode[] are valid children.
+   */
+  children?: React.ReactNode | React.ReactNode[]
+  /**
+   * The ref object to be attached to the list item.
+   */
+  ref?: Ref<HTMLLIElement>
+}
+
+export const ListItem = ({className, children, ...props}: ListItemProps) => {
```
rezrah, line 20: What's the context for having this as a separate component, rather than a namespaced member? I can't think of a use case for having this outside of List, so feels inconsistent with our other API's.

packages/react/src/List/ListItem.tsx
```diff
@@ -0,0 +1,40 @@
...
+import type {BaseProps} from '../component-helpers'
+import {ListContext} from './UnorderedList'
+
+import styles from './List.module.css'
+
+export type ListItemProps = BaseProps<HTMLElement> & {
+  /**
+   * React.ReactNode and React.ReactNode[] are valid children.
+   */
+  children?: React.ReactNode | React.ReactNode[]
```
rezrah, line 13: You can use `PropsWithChildren` instead of manually defining children like this. Should be examples in other components. Maybe AnchorNav.

packages/react/src/List/ListItem.tsx
```diff
@@ -0,0 +1,40 @@
...
+  /**
+   * The ref object to be attached to the list item.
+   */
+  ref?: Ref<HTMLLIElement>
```
rezrah, line 17: BaseProps includes `ref`, you just need to provide this HTML type def as a generic value.

packages/react/src/List/OrderedList.tsx
```diff
@@ -0,0 +1,25 @@
...
+  /**
+   * Ref object to be attached to the list.
+   */
+  ref?: Ref<HTMLOListElement>
+}
+
+export function OrderedList({children, ...props}: OrderedListProps) {
+  return (
+    <ol className={clsx(styles.List, props.className)} {...props}>
+      <ListContext.Provider value={{variant: ''}}>{children}</ListContext.Provider>
```
rezrah, line 22: Why do we need a context provider here? 🤔 context is really expensive and there doesn't seem to be any important stateful logic warranting its usage, or am I missing something? Iterating through children would be faster than using context.

packages/react/src/List/UnorderedList.tsx
```diff
@@ -0,0 +1,36 @@
+import React, {createContext, type Ref} from 'react'
+import clsx from 'clsx'
+import type {BaseProps} from '../component-helpers'
+
+import styles from './List.module.css'
+
+export const ListContext = createContext({
```
rezrah, line 7: Same question, what's reason for using context here? [Related comment.](https://github.com/primer/brand/pull/145/files#r1053565117)

packages/react/src/List/UnorderedList.tsx
```diff
@@ -0,0 +1,36 @@
...
+// QUESTION: Should children be optional? I could see a case for an empty list. But best practice would be to remove it if it is empty for a while. I think leaving it would be a safer bet for any use case.
+export type UnorderedListProps = BaseProps<HTMLElement> & {
+  /**
+   * The semantic structure of list that is presented visually setting 'ol' vs 'ul' based on the style the style of the list.
+   */
+  variant?: 'default' | 'checked'
+  /**
+   * React.ReactNode and React.ReactNode[] are valid children.
+   */
+  children?: React.ReactNode | React.ReactNode[]
```
rezrah, line 23: Recommend we use `PropsWithChildren` to avoid typing this out manually.

packages/react/src/List/OrderedList.tsx
```diff
@@ -0,0 +1,25 @@
...
+
+export type OrderedListProps = BaseProps<HTMLElement> & {
+  /**
+   * React.ReactNode and React.ReactNode[] are valid children.
+   */
+  children?: React.ReactNode | React.ReactNode[]
+  /**
+   * Ref object to be attached to the list.
+   */
+  ref?: Ref<HTMLOListElement>
```
rezrah, line 16: Related comment https://github.com/primer/brand/pull/145/files#r1053562728

packages/react/src/List/OrderedList.tsx
JoshBowdenConcepts: So I would have to do the analysis but looping through the objects and rerendering I would think would be more expensive than adding a context object that doesn't get updated. To do it without using either you would need a name-spaced component for each item. I didn't namespace because we can use a single ListItem component which is easier for developers.

packages/react/src/List/ListItem.tsx
rezrah: If the reason is that it's being used across UnorderedList and OrderedList, I'd recommend importing it into those components as a dependency and then namespacing it. E.g. `List.Item`

packages/react/src/List/List.module.css
```diff
@@ -0,0 +1,35 @@
...
+  font-size: var(--brand-text-size-200);
+  line-height: var(--brand-text-lineHeight-400);
+  color: var(--brand-color-text-default);
+}
+
+ol.List {
+  counter-reset: li;
+}
+
+.ListItem {
```
rezrah, line 19: If ListItem is a separate component, this should be in a dedicated stylesheet for it.

packages/react/src/List/ListItem.tsx
```diff
@@ -0,0 +1,40 @@
...
+}
+
+export const ListItem = ({className, children, ...props}: ListItemProps) => {
+  const {variant} = useContext(ListContext)
+
+  const startingIcon = () => {
```
rezrah, line 23:
```suggestion
  const leadingVisual = () => {
```

packages/react/src/List/ListItem.tsx
```diff
@@ -0,0 +1,40 @@
...
+    switch (variant) {
+      case 'checked':
+        return <CheckIcon className={styles.StartingContent} />
```
rezrah, line 26:
```suggestion
        return <CheckIcon className={styles['ListItem__leading-visual']} />
```

packages/react/src/List/ListItem.tsx
```diff
@@ -0,0 +1,40 @@
...
+      case 'default':
+        return <DashIcon className={styles.StartingContent} />
```
rezrah, line 28:
```suggestion
        return <DashIcon className={styles['ListItem__leading-visual']} />
```

packages/react/src/List/index.ts
```diff
@@ -0,0 +1,3 @@
+export * from './UnorderedList'
+export * from './OrderedList'
+export * from './ListItem'
```
rezrah, line 3: Does `ListItem` need to be exported from here?

packages/react/src/List/OrderedList.tsx
rezrah: There's already [data available ](https://blog.logrocket.com/pitfalls-of-overusing-react-context/) on this. I think the larger point here is that I can't see why you need this as you're just passing props, and you shouldn't need context for that? If you're concerned about performance specifically, surely using namespaced component is the fastest option as you don't need to iterate through children or use context.

rezrah: > adding a context object that doesn't get updated

☝️ if it doesn't get updated, why do you need to create a context to store data? 

Would the easiest option not be to just import ListItem as a private component, map the children and forward the variant to it?

JoshBowdenConcepts: Mapping over the children would be an expensive operation as well we could have the user include props. But thats an extra step for them. I think mapping over children would be more expensive than adding a context object that wouldn't change.

JoshBowdenConcepts: How would we go about testing which approach is more performant?

packages/react/src/List/ListItem.tsx
JoshBowdenConcepts: When I use this I have run into the following error:
```
Type '{ children: Element; accessKey: string; accessKeyLabel: string; autocapitalize: string; dir: string; draggable: boolean; hidden: boolean; inert: boolean; innerText: string; lang: string; ... 277 more ...; className: string; }' is not assignable to type 'DetailedHTMLProps<HTMLAttributes<HTMLUListElement>, HTMLUListElement>'.
  Type '{ children: Element; accessKey: string; accessKeyLabel: string; autocapitalize: string; dir: string; draggable: boolean; hidden: boolean; inert: boolean; innerText: string; lang: string; ... 277 more ...; className: string; }' is not assignable to type 'HTMLAttributes<HTMLUListElement>'.
```

packages/react/src/List/ListItem.tsx
JoshBowdenConcepts: So I would change it to UnorderedList.Item. That works.

.changeset/blue-dots-battle.md
JoshBowdenConcepts: That is a mistake. Updating in next push

packages/react/src/List/List.module.css
JoshBowdenConcepts: How about the issue that they are linked for the OL and would need access to both the List and ListItem class in one file. Would you put that declaration in the ListItem file?

packages/react/src/List/ListItem.tsx
JoshBowdenConcepts: In next push

packages/react/src/List/ListItem.tsx
JoshBowdenConcepts: In next push

packages/react/src/List/index.ts
JoshBowdenConcepts: In next push

packages/react/src/List/ListItem.tsx
JoshBowdenConcepts: In next push

packages/react/src/List/ListItem.tsx
JoshBowdenConcepts: In next push

.changeset/blue-dots-battle.md
JoshBowdenConcepts: In next push

packages/react/src/List/ListItem.tsx
JoshBowdenConcepts: Fixed in next push

packages/react/src/List/ListItem.tsx
JoshBowdenConcepts: In next push

packages/react/src/List/UnorderedList.tsx
JoshBowdenConcepts: In next push

packages/react/src/List/OrderedList.tsx
JoshBowdenConcepts: In next push

packages/react/src/List/OrderedList.tsx
rezrah: We can discuss the time complexity around this component, but bear in mind that this is going to be used on marketing pages, where we are intentional about not using large data sets. We don't necessarily need to worry about time complexity as much as introducing space complexity for no immediately beneficial reason. 

[Official context docs](https://reactjs.org/docs/context.html#before-you-use-context) state the typical use-case for it, and to avoid using it for situations that don't need it. I can't see any prop-drilling happening, or parent to grandchild data flow, or the parent needing to hold and mutate unique state for each child instance. 

>  context object that wouldn't change

I think it would here though,  because I can't see any memoization being used so you would incur extra rerenders using this approach right now. In fact, you're calling useContext `n` times on each re-render. This makes it no different to iterating through children, except with that approach you're not increasing memory utilisation. 

Going back to the wider point here around this implementation. It feels like we're overengineering this by introducing unnecessary complexity. The user should have a convenient API for something that they expect will be fairly straight forward to render. As a potential user, I'd find it strange to know that there is internal state being maintained in a component that's just an abstraction of `ul` and `li`. 

If we're going for the most performant and pragmatic option, we should create component-level encapsulation around ListItem, that is local to UnorderedList and OrderedList respectively. Something like...

```js

// UnorderedList.tsx

function Root() {} // private

function Item {} // private. use shared styles but otherwise just a wrapper around li

export const OrderedList  = Object.assign(Root, { Item })

```

☝️ I think ListItem should generally be local to each UnorderedList and OrderedList. Yes, there's duplication, but these components are only relevant to their parent components so scoping might make sense.

> Mapping over the children would be an expensive operation as well we could have the user include props

Mapping is functionally pure and deterministic. Built-ins like `map` are also browser optimised and native to JS, unlike hooks like context and their providers which are framework-level functions. Not sure if that statement is universally applicable, but would be interesting to see the benchmarks.

packages/react/src/List/List.module.css
rezrah: > How about the issue that they are linked for the OL and would need access to both the List and ListItem class in one file. Would you put that declaration in the ListItem file?

What did you mean by linked? I see ListItem declared a separate component, and there doesn't seem to be a reference to `.ListItem` in the `UnorderedList` component. We typically create stylesheets per components or pattern, and given `UnorderedList `, `OrderedList` and `ListItem` are all separate, I would create individual stylesheets for each of them. Related comment about locally scoping `ListItem` so maybe you only need two stylesheets instead.

packages/react/src/List/OrderedList.tsx
JoshBowdenConcepts: The official context docs state that you shouldn't use context for 

> context objects that wouldn't change

Because you should pass the props to the component directly in this case. You have mentioned not wanting to do this because the user shouldn't have to supply more than top level props. Even though this is the most performant and expected way in React. (I do agree that not supplying props on the children does make sense).

Context change would need to happen by passing a new value to the context which this wouldn't do and has no way of doing internally. We could additionally memoize it for further security but at this moment it is the same result.

We could create a component for each list but this would expand our lists to UnorderedList, CheckedList, and OrderedList. If I were using these components I would be less interested in having to import a new list each time I want a slightly different leadingVisual (But maybe our engineers would prefer this approach). Because there are two leading visual options for UnorderedList you will need to manage state anyway to supply these within one component. This comes down to how we would do that.

Since that is the case I think that the only way to know would be to benchmark the two approaches. We can take the time to do that but seeing as I am out starting tomorrow and we wanted to get this done by end of year I wasn't sure if moving it forward was more pressing. Since I am a fan of the context API approach (if it works for you) this could be a good case for passion week to benchmark to find out for sure which makes sense. We could even create an ADR around this based on results because I believe other engineers are making these same decisions.

packages/react/src/List/List.module.css
JoshBowdenConcepts: The CSS needs access to the List class as well as the ListItem class in the same declaration to elegantly create the numbers in the ordered list. If you split them the CSS breaks. How would you approach this kind of CSS?

JoshBowdenConcepts: Also wouldn't repeating CSS declarations bloat the CSS bundles when both are in use? Or does it remove duplicates?

packages/react/src/List/OrderedList.tsx
rezrah: > We could create a component for each list but this would expand our lists to UnorderedList, CheckedList, and OrderedList.

Only UnorderedList and OrderedList are needed. Your switch logic for the icon in ListItem can move to the UnorderedList component file as it doesn't concern OrderedList at all, right? (given you're relying on the default value and returning an empty fragment). 

> Because there are two leading visual options for UnorderedList you will need to manage state anyway to supply these within one component.

No so I'm saying that there is no need for local (useState) or external (useContext) component state at all for this type of component 😅. You can just forward the prop for variant onto the child list item (see example in thread above ☝️). 

> Since I am a fan of the context API approach

I like Context for state management too 😄. I'm trying to be mindful of over-utilising React-specific features when there are already native JS built-in that can achieve the same thing without all the indirection. React provides a nice first-class API for iterating children (React.Children), which is why we use it so heavily in both Primer Brand and in Primer React Components.

> I believe other engineers are making these same decisions.

That's interesting, can you point me at these examples please? AFAIK, the use-cases for Context to date have been to use it to send downstream data back _up_ the tree, or avoid prop-drilling. Otherwise, we use React.Children API to iterate immediate children and forward props, which is already documented in the ADR I linked previously. Has that changed in PRC now?

packages/react/src/List/List.module.css
rezrah: They should get deduped now but I haven't tested it recently. Fairly easy to fix using postcss if it doesn't. Either way, we should leave that to Webpack to figure out through the dependency tree, as it will become important to reduce side-effects when we enable code splitting.

rezrah: > The CSS needs access to the List class as well as the ListItem class in the same declaration to elegantly create the numbers in the ordered list. If you split them the CSS breaks. How would you approach this kind of CSS?

That's a good question 🤔 

So first, I don't think this stylesheet should be called `List`, as there's no `List` component anymore. Perhaps `list-shared.module.css.` or similar, and should live in the root `/src/list/list-shared.css`

Then you have two more stylesheets for `UnorderedList.module.scss` and `OrderedList.module.css`, and they only have styles pertaining to their respective component files.

As part of this change, I think the folder structure should generally change to be more like `src/forms`. 

```ascii
.
├── src
├── ...
└── lists/
    ├── index.ts
    ├── list.shared.module.css
    ├── UnorderedList/
    │   ├── UnorderedList.module.css
    │   ├── UnorderedList.tsx
    │   └── index.ts
    └── OrderedList/
        ├── OrderedList.module.css
        ├── OrderedList.tsx
        └── index.ts
```

apps/docs/content/components/OrderedList.mdx
```diff
@@ -0,0 +1,44 @@
+---
+title: List
```
rezrah, line 2: Should be 👇 

```suggestion
title: OrderedList
```

apps/docs/content/components/OrderedList.mdx
```diff
@@ -0,0 +1,44 @@
...
+status: Experimental
+figma: 'https://www.figma.com/file/BJ95AjraesmRCWsKA013GS/Primer-Brand?node-id=354%3A9513'
```
rezrah, line 4: This goes to typography page in the Figma library 🤔

apps/docs/content/components/UnorderedList.mdx
```diff
@@ -0,0 +1,63 @@
...
+  <UnorderedList.Item>GitHub Security Advisories</UnorderedList.Item>
+  <UnorderedList.Item>Code and secret scanning</UnorderedList.Item>
+  <UnorderedList.Item>Dependency review</UnorderedList.Item>
+  <UnorderedList.Item>
+    Automated authentication and identity management
+  </UnorderedList.Item>
+</UnorderedList>
+```
+
+### Checked UnorderedList
```
rezrah, line 32:
```suggestion
### Checked variant
```

apps/docs/content/components/UnorderedList.mdx
```diff
@@ -0,0 +1,63 @@
...
+```
+
+## Component props
+
+### UnorderedList
+
+| Name        | Type                                   |  Default  | Description                                                |
+| :---------- | :------------------------------------- | :-------: | :--------------------------------------------------------- |
+| `className` | `string`                               |           | Sets a custom class on the root element                    |
+| `variant`   | `default`, `checked`                   | `default` | Specify alternative List appearance                        |
```
rezrah, line 55:
```suggestion
| `variant`   | `default`, `checked`                   | `default` | Specify alternative leading visuals for list items                     |
```

packages/react/src/List/List.module.css
```diff
@@ -0,0 +1,35 @@
...
+  display: flex;
+  flex-direction: column;
+  margin: 0;
+  list-style: none;
+  gap: var(--base-size-8);
+  padding-inline-start: 0px;
+  font-family: var(--brand-body-fontFamily);
+  font-weight: var(--brand-text-weight-400);
+  font-size: var(--brand-text-size-200);
+  line-height: var(--brand-text-lineHeight-400);
```
rezrah, line 11: Quite a big difference between 400 and 200 in the type scale. This is rendering at 28px but the Figma spec is saying it should be 20px. Can we change this to whichever position maps to 20px, maybe 200 or 300?

packages/react/src/List/List.module.css
```diff
@@ -0,0 +1,35 @@
...
+  display: flex;
+  gap: var(--base-size-12);
+  align-items: center;
+}
+
+.ListItem__leading-visual {
+  fill: var(--brand-color-text-muted);
+}
+
+ol.List > .ListItem::before {
```
rezrah, line 29: This is making me wonder if I forgot to enable stylelint 🙀, because linter should usually flag this as a violation. We don't use tag in selector to qualify the class, there's a specific rule around this. 

I'd approach this slightly differently. There are examples in other components too that might be worth checking out.

```css
.OrderedList__item::before {
```

OrderedList being the block, and item being the element of it. No need for qualifying selector then, or direct descendent. Can we do that?

packages/react/src/List/ListItem.tsx
```diff
@@ -0,0 +1,40 @@
...
+
+  const leadingVisual = () => {
+    switch (variant) {
+      case 'checked':
+        return <CheckIcon className={styles['ListItem__leading-visual']} />
+      case 'default':
+        return <DashIcon className={styles['ListItem__leading-visual']} />
+
+      default:
+        return <></>
```
rezrah, line 31: This feels strange to return an empty fragment, do you want null instead to provide an actual falsey value?

packages/react/src/List/OrderedList.stories.tsx
```diff
@@ -0,0 +1,27 @@
...
+import React from 'react'
+import {OrderedList} from '.'
+
+export default {
+  title: 'Components/OrderedList',
+  component: OrderedList,
+  argTypes: {
+    children: {
+      table: {
+        disable: true
```
rezrah, line 11: Nice to have for future, we should re-enable this so folks can change the text, especially in Playground. Here's how I [did it for AnchorNav](https://primer.style/brand/storybook/?path=/story/components-anchornav--playground).

packages/react/src/List/UnorderedList.stories.tsx
```diff
@@ -0,0 +1,38 @@
...
+export default {
+  title: 'Components/UnorderedList',
+  component: UnorderedList,
+  args: {
+    variant: 'default'
+  },
+  argTypes: {
+    variant: {
+      description:
+        'The semantic structure of list that is presented visually setting "ol" vs "ul" based on the style the style of the list.',
```
rezrah, line 14: Finding this description a bit hard to read as it's quite long, and I'm unclear what it's saying.

```suggestion
'Specify alternative leading visuals for list items ',
```

packages/react/src/List/UnorderedList.test.tsx
```diff
@@ -0,0 +1,48 @@
...
+    const {getByTestId} = render(
+      <UnorderedList data-testid={mockId}>
+        <UnorderedList.Item>Automatic security and version updates</UnorderedList.Item>
+      </UnorderedList>
+    )
+    const el = getByTestId(mockId)
+    expect(el.tagName).toBe('UL')
+  })
+
+  it('renders the starting svg', () => {
```
rezrah, line 38: Can we have some additional tests for the variant prop too? One for default tag and the other for optional check variant.

packages/react/src/List/OrderedList.tsx
```diff
@@ -0,0 +1,19 @@
...
+import {ListItem} from './ListItem'
+
+import styles from './List.module.css'
+
+export type OrderedListProps = PropsWithChildren<BaseProps<HTMLOListElement>>
+
+function Root({children, ...props}: OrderedListProps) {
+  return (
+    <ol className={clsx(styles.List, props.className)} {...props}>
+      <ListContext.Provider value={{variant: ''}}>{children}</ListContext.Provider>
```
rezrah, line 14: Looking at this fresh, I wouldn't know that an empty string means it's for ordered lists. There should be type safety around this. Can we extend the switch case to mention numeric and handle it in a truthy way, rather than falsey via empty string.

packages/react/src/List/listContext.ts
```diff
@@ -0,0 +1,5 @@
+import {createContext} from 'react'
+
+export const ListContext = createContext({
+  variant: 'default'
```
rezrah, line 4: This is shared with OrderedList, so should it be default? Initialising as default makes sense for Unordered though.

packages/react/src/List/OrderedList.tsx
JoshBowdenConcepts: > No so I'm saying that there is no need for local (useState) or external (useContext) component state at all for this type of component 😅. You can just forward the prop for variant onto the child list item (see example in thread above ☝️).

You're right I think my use of "state" was misused in the context of react. There is data to be managed regardless of  the approach we would need to toggle on the leadingVisual depending on the type of Unordered list used. (The data  is what I was referring to as "state")

> I like Context for state management too 😄. I'm trying to be mindful of over-utilising React-specific features when there are already native JS built-in that can achieve the same thing without all the indirection. React provides a nice first-class API for iterating children (React.Children), which is why we use it so heavily in both Primer Brand and in Primer React Components.

I understand the mindfulness here. We are trying to create an anti-pattern within react by avoiding passing props directly to the children. I think that both approaches are functional I just want to know what is more performant. Whichever is more performant has my vote for sure.

> That's interesting, can you point me at these examples please? AFAIK, the use-cases for Context to date have been to use it to send downstream data back up the tree, or avoid prop-drilling. Otherwise, we use React.Children API to iterate immediate children and forward props, which is already documented in the ADR I linked previously. Has that changed in PRC now?

For TreeView it seems that it is being used to pass props to children without manually applying them:
https://github.com/primer/react/blob/main/src/TreeView/TreeView.tsx#L506

apps/docs/content/components/UnorderedList.mdx
JoshBowdenConcepts: In next push

apps/docs/content/components/UnorderedList.mdx
JoshBowdenConcepts: In next push

apps/docs/content/components/OrderedList.mdx
JoshBowdenConcepts: In next push

packages/react/src/List/UnorderedList.test.tsx
JoshBowdenConcepts: I was trying to do this but I wasn't sure how to check for specific svg's being loaded. If you have any ideas let me know. I could do a string match but that seems like a cop out

packages/react/src/List/List.module.css
JoshBowdenConcepts: Updated in next push

packages/react/src/List/List.module.css
JoshBowdenConcepts: This doesn't quite work because you want the CSS split and the modules for ListItem and OrderedList don't have a relation to one another. This is the pro for the consolidated css.

packages/react/src/List/ListItem.tsx
JoshBowdenConcepts: This just feels like a different approach to accomplish the same thing but maybe it is more performant? So updating in next push.

packages/react/src/List/OrderedList.stories.tsx
JoshBowdenConcepts: Thats an interesting approach. My only concern is that it would muddle the story by adding data and the context how you have done it which might confuse a user

`({data, ...args}: {data: MockData; offset: number}) => {`

This is in the beginning of the story in Storybook

packages/react/src/List/UnorderedList.stories.tsx
JoshBowdenConcepts: Updated in next push

packages/react/src/List/OrderedList.tsx
JoshBowdenConcepts: I think this might be a preference thing as the user does not have access to this in any way. It accurately displays that no variant is supplied which refers the the context of leading visuals. I think that empty string is more accurate to its functionality than passing numeric would be.

packages/react/src/List/listContext.ts
JoshBowdenConcepts: It should be initialized to something picking ordered or unordered list to make default doesn't really mean much to me. And it functionally acts the same way as the state is handled internally.

packages/react/src/List/List.module.css
JoshBowdenConcepts: Updated in next push

packages/react/src/List/UnorderedList.test.tsx
rezrah: Could you use a data-attribute or pass a custom className and pass the name of the icon? It doesn't need to be an equality check in the actual SVG markup or even file path, as that's a project side-effect. Calling the correct name seems important though I feel.

packages/react/src/List/ListItem.tsx
rezrah: Yeah there's probably some optimisations under the hood, but from a readability pov they're quite different, because a react fragment will return a truthy value while null be falsy.

packages/react/src/List/OrderedList.tsx
rezrah: > I think this might be a preference thing as the user does not have access to this in any way. 

Yep, feedback is coming from maintainability perspective, not user.

> I think that empty string is more accurate to its functionality than passing numeric would be

In future we may need to extend support to definition lists, or other type of lists. The switch works well for that scenario, but there's an assumption that `'' === 'numeric`, which isn't immediately clear for anyone that has to implement this in future. I would also assume some type safety here, given we have it for everything else.

apps/docs/content/components/OrderedList.mdx
```diff
@@ -0,0 +1,44 @@
+---
+title: Ordered List
+status: Experimental
+figma: 'https://www.figma.com/file/BJ95AjraesmRCWsKA013GS/branch/NzBMPIfPrFU9I87d3k7KVq/Primer-Brand?node-id=4045%3A17560&t=W8B9ArDngbzJCC2m-0'
+source: https://github.com/primer/brand/tree/main/src/List/OrderedList/OrderedList.tsx
+storybook: '/brand/storybook/?path=/story/components-orderedlist--default'
+description: Use the OrderedList component to render a list of ordered items.
+---
+
```
rezrah, line 9: FYI, there's an issue with the import in docs. Previews aren't working.

Screenshot 2023-01-10 at 12 11 27

apps/docs/content/components/OrderedList.mdx
```diff
@@ -0,0 +1,44 @@
+---
+title: Ordered List
```
rezrah, line 2:
```suggestion
title: OrderedList
```

apps/docs/content/components/OrderedList.mdx
```diff
@@ -0,0 +1,44 @@
...
+description: Use the OrderedList component to render a list of ordered items.
+---
+
+```js
+import {OrderedList} from '@primer/react-brand'
+```
+
+## Examples
+
+### OrderedList
```
rezrah, line 16: For consistency with other components docs

```suggestion
### Default
```

apps/docs/content/components/UnorderedList.mdx
```diff
@@ -0,0 +1,63 @@
+---
+title: Unordered List
```
rezrah, line 2:
```suggestion
title: UnorderedList
```

```

File: apps/docs/content/components/UnorderedList.mdx [hunk 36]
```diff
@@ -0,0 +1,63 @@
...
+```js
+import {UnorderedList} from '@primer/react-brand'
+```
+
+## Examples
+
+### Default
+
+```jsx live
+<UnorderedList variant="default">
```
rezrah, line 19: Nit. Do you need variant="default" prop here if it's set as a default value?

apps/docs/content/components/UnorderedList.mdx
```diff
@@ -0,0 +1,63 @@
...
+status: Experimental
+figma: 'https://www.figma.com/file/BJ95AjraesmRCWsKA013GS/branch/NzBMPIfPrFU9I87d3k7KVq/Primer-Brand?node-id=4045%3A17093&t=XJ9W58VgtbVo74gb-0'
+source: https://github.com/primer/brand/tree/main/src/List/UnorderedList/UnorderedList.tsx
+storybook: '/brand/storybook/?path=/story/components-unorderedlist--default'
+description: Use the UnorderedList component to render a list of unordered items.
+---
+
+```js
+import {UnorderedList} from '@primer/react-brand'
+```
```
rezrah, line 12: FYI, same issue with previews for UnorderedList. I think it's the missing exports in list folder
Screenshot 2023-01-10 at 12 28 36

packages/react/src/List/OrderedList/OrderedList.module.css
```diff
@@ -0,0 +1,3 @@
+ol.OrderedList {
```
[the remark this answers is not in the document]
JoshBowdenConcepts: Updated in next push

apps/docs/content/components/UnorderedList.mdx
JoshBowdenConcepts: Yes because not setting the variant would default to nothing which would in turn be numeric. The variant displays the leading visual for a default list. Had these all been one component the logic behind these would have been simpler. And easier for a user to use.

apps/docs/content/components/OrderedList.mdx
JoshBowdenConcepts: Updated in next push

apps/docs/content/components/OrderedList.mdx
JoshBowdenConcepts: Updated in next push

packages/react/src/List/OrderedList.tsx
JoshBowdenConcepts: I think this is another reason if we want to improve user experience and developer experience these should in fact be one component rather than 2. This would simplify use and would mitigate these issues that come down to preference and maintainability.

packages/react/src/List/UnorderedList.test.tsx
JoshBowdenConcepts: I am still not sure how that would work. What are you suggesting the output would be for passing either a custom className or a data-attribute. Those are used for grabbing pieces of the rendered output but I would have to change the component internally to gain access to those in the testing suite and that doesn't seem like the right approach because it is adding unnecessary complication. I do believe there would be a way to see the svg element using this testing suite but I haven't figured out how to see these nested elements yet.

JoshBowdenConcepts: It could be a follow up ticket when I have learned more but I can keep digging.

apps/docs/content/components/UnorderedList.mdx
JoshBowdenConcepts: Updated in next push

apps/docs/content/components/OrderedList.mdx
JoshBowdenConcepts: In next push

packages/react/src/List/List.module.css
JoshBowdenConcepts: Pushed in changes

danielguillan:
A couple of small issues:

- The list icon/number is misaligned When the text wraps into multiple lines. It should remain at the same position on the first line.
- The items are missing internal vertical padding, so the total height of each item sums `24px`. 
  - UnorderedList should have `4px` vertical padding (`16px icon + 4px padding * 2 = 24px`) - 
  - OrderedList should have `2px` vertical padding (`20px line-height + 2px padding * 2 = 24px`)

danielguillan (approved):

rezrah (approved):
This looks great! Thanks for following up on the feedback. 

For posterity, I'm still a little wary about the proposed implementation paradigm here, but given the surface area is small here I agree that we should ship these components 🚀 :shipit:

JoshBowdenConcepts:
> This looks great! Thanks for following up on the feedback.
> 
> For posterity, I'm still a little wary about the proposed implementation paradigm here, but given the surface area is small here I agree that we should ship these components 🚀 :shipit:

I think we should revisit this for sure when we have some time. Maybe end of quarter.

PR #162, opened 2023-01-18
Add Card component

josepmartins:
## Summary

The first iteration of the Card component. The component now covers all the necessities for the Galaxy Phase 2 (composable API that allows a `Heading` and `Description`) and will be upgraded following the alignment work that is happening across .com pages (see more: https://github.com/github/brand-marketing-design/issues/720)

This closes: https://github.com/github/galaxy.github.com/issues/121
Tracking issue: https://github.com/github/primer/issues/1806
Figma branch: https://www.figma.com/file/BJ95AjraesmRCWsKA013GS/branch/sZ6jw7aenZ9VWl0dLM9vrK/Primer-Brand?node-id=5407%3A32782&t=iQ8GUOpbHoYhemEe-1

## List of notable changes:

- **added** `Card` component with `Heading` and `Description` items.
- **added** basic functional primitives.  
- **added** default story to showcase `Card`.
- **added** documentation.

## Steps to test:

- Open the CTAForm component in CI-deployed preview environment
- Go to Card story in Storybook
- Verify that the Card behaves as described in the following issue.

👀 Storybook: https://primer-f13cea2127-26139705.drafts.github.io/storybook/?path=/story/components-card--default
📖 Docs: https://primer-f13cea2127-26139705.drafts.github.io/components/Card

## Contributor checklist:

## Reviewer checklist:

## Screenshots:

> Card in light mode
> Screenshot 2023-02-15 at 17 13 15
> 

> Card in dark mode
> Screenshot 2023-02-15 at 17 13 19
>


packages/react/src/Card/Card.module.css
```diff
@@ -0,0 +1,29 @@
...
+  display: flex;
+  flex-direction: column;
+  text-decoration: none;
+  color: var(--brand-color-text-default);
+  background-color: var(--brand-Card-background-default);
+  border: 1px solid var(--brand-color-border-default);
+  box-shadow: var(--brand-Card-visual-shadow);
+  border-radius: 6px;
+  padding: var(--base-size-32);
+  transition: transform 0.6s cubic-bezier(0.16, 1, 0.3, 1);
```
josepmartins, line 12: Stole this transition + hover from the `resourcesCard` component in .com pages. Not sure if these need to be a custom Card tokens 🤔

packages/react/src/Card/Card.tsx
```diff
@@ -0,0 +1,95 @@
...
+  })
+
+  return (
+    <a href={linkHref} className={clsx(styles.Card, className)}>
+      {React.Children.toArray(filteredChildren).map(child => {
+        return child
+      })}
+      <div className={styles.Card__action}>
+        {/* Temporary using Link component to not repeat styles */}
+        <Link>{link ? link : 'Learn more'}</Link>
```
rezrah, line 53: One option could be to add an `as` prop into Link, to force it to be a span but retain styling. Alternatively, you could import the css module for Link into the Card file. Let me know if you want to pair on this. I think we should change this as nesting `a` tags is problematic.

packages/react/src/Card/Card.tsx
```diff
@@ -0,0 +1,95 @@
...
+  children: React.ReactNode | React.ReactElement<CardHeadingProps> | React.ReactElement<CardDescriptionProps>
+  /**
+   * Changes the link content of the card
+   * * @default 'Learn more'
+   * */
+  link?: string
+  /**
+   * The href of the link
+   * */
+  linkHref: string
```
rezrah, line 32: Can we change this to be `href` to make it standard?

packages/react/src/Card/Card.tsx
```diff
@@ -0,0 +1,95 @@
...
+export type CardProps = {
+  /**
+   * Valid children include Card.Image, Card.Heading, and Card.Description
+   */
+  children: React.ReactNode | React.ReactElement<CardHeadingProps> | React.ReactElement<CardDescriptionProps>
+  /**
+   * Changes the link content of the card
+   * * @default 'Learn more'
+   * */
+  link?: string
```
rezrah, line 28: Should this be `label` or similar? I feel like `link` has a connotation to hyperlinks. WDYT?

packages/react/src/Card/Card.tsx
```diff
@@ -0,0 +1,95 @@
...
+  /**
+   * The href of the link
+   * */
+  linkHref: string
+} & BaseProps<HTMLLinkElement>
```
rezrah, line 33: Should this be `HTMLAnchorElement`?

```suggestion
} & BaseProps<HTMLAnchorElement>
```

packages/react/src/Card/Card.tsx
```diff
@@ -0,0 +1,95 @@
...
+
+const CardRoot = forwardRef<HTMLLinkElement, CardProps>(({children, className, link, linkHref}) => {
```
rezrah, line 35:
```suggestion
const CardRoot = forwardRef< HTMLAnchorElement, CardProps>(({children, className, link, linkHref}) => {
```

packages/react/src/Card/Card.tsx
```diff
@@ -0,0 +1,95 @@
...
+    if (React.isValidElement(child) && typeof child.type !== 'string') {
+      if (childrenHasFragment || child.type === CardHeading || child.type === CardDescription) {
+        return true
+      }
+    }
+    return false
+  })
+
+  return (
+    <a href={linkHref} className={clsx(styles.Card, className)}>
```
rezrah, line 47: Could we forward all native `a` element attributes here too using spread? Imagine things like `target` will become useful in some use-cases, as well as `referrer`.

packages/react/src/Card/Card.tsx
```diff
@@ -0,0 +1,95 @@
...
+      <div className={styles.Card__action}>
+        {/* Temporary using Link component to not repeat styles */}
+        <Link>{link ? link : 'Learn more'}</Link>
+      </div>
+    </a>
+  )
+})
+
+type CardHeadingProps = BaseProps<HTMLHeadingElement> & {
+  children: string
```
rezrah, line 60: I _think_ this may cause problems with eval or interpolated text. Can we use ReactElement instead?

packages/react/src/Card/Card.module.css
```diff
@@ -0,0 +1,29 @@
+.Card {
+  max-width: var(--brand-Card-maxWidth);
+  display: flex;
+  flex-direction: column;
+  text-decoration: none;
+  color: var(--brand-color-text-default);
+  background-color: var(--brand-Card-background-default);
+  border: 1px solid var(--brand-color-border-default);
```
rezrah, line 8:
```suggestion
  border: var(--brand-borderWidth-thin) solid var(--brand-color-border-default);
```

packages/react/src/Card/Card.module.css
```diff
@@ -0,0 +1,29 @@
...
+  box-shadow: var(--brand-Card-visual-shadow);
+  border-radius: 6px;
```
rezrah, line 10:
```suggestion
  border-radius: var(--brand-borderRadius-medium);
```

packages/react/src/Card/Card.module.css
rezrah: I think it should be, yeah 👍. We have a similar token for `--brand-control-animation-easing`, which is possibly even the same value.

josepmartins: It is close, but not exactly the same. We use the `brand control animation` tokens for small elements like checkboxes show/hide I think for this case `Cards` deserve an special animation treatment as it's a "bigger" element on the interface.

packages/react/src/Card/Card.tsx
josepmartins: Done!

packages/react/src/Card/Card.tsx
josepmartins: yep, I wasn't sure what to use here. `Label` sounds good to me

josepmartins: done!

packages/react/src/Card/Card.tsx
```diff
@@ -0,0 +1,95 @@
...
+  as?: 'h2' | 'h3' | 'h4' | 'h5' | 'h6'
```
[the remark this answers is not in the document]
josepmartins: I'd rather use `Link` styles with custom markup + passing the hover/focus state to an `ExpandableArrow`. I think this allows us more control over the hover styles and doesn't involve changing the API:

Made a commit with those changes in case you want to 👁️ : https://github.com/primer/brand/pull/162/commits/dc6056341dbe70696374943c33f288cc7ed5da48

packages/react/src/Card/Card.tsx
josepmartins: Done!

packages/react/src/Card/Card.tsx
josepmartins: I'd rather use `Link` styles with custom markup + passing the hover/focus state to an `ExpandableArrow`. I think this allows us more control over the hover styles and doesn't involve changing the API:

Made a commit with those changes in case you want to 👁️ : https://github.com/primer/brand/pull/162/commits/dc6056341dbe70696374943c33f288cc7ed5da48

packages/react/src/Card/Card.tsx
```diff
@@ -0,0 +1,152 @@
...
+        className={clsx(styles.Card, className)}
+        onMouseEnter={handleMouseEnter}
+        onMouseLeave={handleMouseLeave}
+        onFocus={handleOnFocus}
+        onBlur={handleOnBlur}
+        {...props}
+      >
+        {React.Children.toArray(filteredChildren).map(child => {
+          return child
+        })}
```
rezrah, line 104: Unless I'm mistaken, `filteredChildren` is already an array? Could you just reference it directly 

```jsx
{filteredChildren}
```

josepmartins: Done!

packages/react/src/Card/Card.tsx
```diff
@@ -0,0 +1,152 @@
...
+        setIsFocused(!isFocused)
+        onBlur(event)
+      },
+      [onBlur, isFocused]
+    )
+
+    const childrenHasFragment = React.Children.toArray(children).some(child => isFragment(child))
+    const filteredChildren = React.Children.toArray(children).filter(child => {
+      if (React.isValidElement(child) && typeof child.type !== 'string') {
+        if (childrenHasFragment || child.type === CardHeading || child.type === CardDescription) {
```
rezrah, line 85: To optimise this a little, could we call `isFragment(child)` here directly? It'd avoid having to create `childrenHasFragment` reference using a separate loop as we only call it once. We can update FAQ to do the same also?

```suggestion
        if (isFragment(child) || child.type === CardHeading || child.type === CardDescription) {
```

packages/react/src/Card/Card.tsx
```diff
@@ -0,0 +1,152 @@
...
+        onMouseLeave={handleMouseLeave}
+        onFocus={handleOnFocus}
+        onBlur={handleOnBlur}
+        {...props}
+      >
+        {React.Children.toArray(filteredChildren).map(child => {
+          return child
+        })}
+        <div className={styles.Card__action}>
+          <Text as="span" size="300" className={clsx(stylesLink['Link--label'])}>
```
rezrah, line 104: Just an FYI that we typically follow BEM naming convention in Primer Brand. So this should probably be `Link__label`. The current name indicates it would be a modifier.

packages/react/src/Card/Card.tsx
rezrah: We probably should yeah. Feel free to do it here, or we can do that in a follow-up. It might be a non-issue though, so maybe worth trying it in Storybook and seeing if you can produce the issue.

packages/react/src/Card/Card.stories.tsx
```diff
@@ -0,0 +1,26 @@
...
+      control: {
+        type: 'text'
+      },
+      defaultValue: 'https://github.com'
+    }
+  }
+} as ComponentMeta<typeof Card>
+
+export const Default = (args: CardProps) => (
+  <div style={{display: 'flex', flexWrap: 'wrap', gap: '3rem', justifyContent: 'center', marginTop: '10vh'}}>
```
rezrah, line 20: The default example should be the Card on its own. I think this centred example could go into /features folder or Playground stories instead perhaps? Also is it possible to use Stack here instead of div? There should be a vertical center example in the docs for it.

packages/react/src/Card/Card.tsx
```diff
@@ -0,0 +1,142 @@
...
+  children: React.ReactNode | React.ReactElement<CardHeadingProps> | React.ReactElement<CardDescriptionProps>
+  /**
+   * The href of the link
+   * */
+  href: string
+  /**
+   * Changes the link label of the card
+   * * @default 'Learn more'
+   * */
+  label?: string
```
rezrah, line 34: @josepmartins - with the new requirements for supporting Label component in here I'm not wondering if `label` could be confusing. One option could be to replace `label` with `linkText` (which is what I think you did before? 😅). Then make Card.Label a separate child component. What do you think?

packages/design-tokens/src/tokens/functional/components/card/base.json
```diff
@@ -0,0 +1,22 @@
+{
+  "brand": {
+    "Card": {
+      "maxWidth": {
+        "value": "420px"
+      },
+      "visual": {
```
rezrah, line 7: is `visual` needed here? I think we could go straight to shadow IMO.

apps/docs/content/components/Card.mdx
```diff
@@ -0,0 +1,89 @@
...
+
+```js
+import {Card} from '@primer/react-brand'
+```
+
+## Examples
+
+### Default
+
+The default card can be built with a custom heading and a description. The link label defaults to `Read more` and can be customized using the `link` prop.
```
rezrah, line 20:
```suggestion
The default card displays a heading and description. The link text defaults to `Read more`, but can be customized using the `label` prop.
```

apps/docs/content/components/Card.mdx
```diff
@@ -0,0 +1,89 @@
+---
+title: Card
+status: Experimental
+source: https://github.com/primer/brand/tree/main/packages/react/src/Card/Card.tsx
+storybook: '/brand/storybook/?path=/story/components-card--playground'
+description: Use the Card component to display information in a compact way and link to other pages.
```
rezrah, line 6: Could we make it clear that it links to internal pages and shouldn't be used to link to external ones?

rezrah:
```suggestion
description: Use the Card component to display information in a compact way and link to other internal pages.
```

rezrah (approved):
@josepmartins - this looks great ✨.

Few things:

- Looks like we're missing unit test coverage for this. Are you planning to add this after? 
- When are we adding support for the Labels? I think the `label` prop may need to change back given those requirements came in, to avoid naming confusion.

Also cc. @egim as our Marketing Eng reviewer this quarter. If you have any thoughts or feedback on this component before it ship, please let @josepmartins know.

egim:
@rezrah - I think the following discussion might be more useful somewhere else, but thought I would ask here.

Question about an existing design within Landing Pages:

Screenshot 2023-02-27 at 8 22 10 AM

Currently, the background "card" above is implemented as a custom component wrapping around a testimonial object in the Resources Hub application. This edge case would not be related to this component, correct?

Screenshot 2023-02-27 at 8 21 50 AM

Is it worth discussing in the future about making the background of the card optional and having an optional image (like above)? Maybe this is another component altogether?

Regardless, I'm thinking custom components for now and I don't want to hold up this PR 👍🏼.

rezrah:
@josepmartins - I'm merging this so it gets into the next release, which is time sensitive. Hope that's okay. 

For posterity, some things to follow-up on at a later date:

- adding support for Labels
- backfilling test coverate

PR #163, opened 2023-01-19
CTAForm Component

JoshBowdenConcepts:
## Summary

Creating a CTAForm component that most commonly will be used to create subscribe CTA blocks for marketing pages. It can be applied for other use cases.

## Deployments:
[MDX Documentation](https://primer-7a98f9a2bb-26139705.drafts.github.io/components/CTAForm)
[Storybook](https://primer-7a98f9a2bb-26139705.drafts.github.io/storybook/?path=/docs/components-ctaform--playground)

## List of notable changes:

- **added** CTAForm component

- Making sure the component functions and looks like it is supposed to.

## Steps to test:

1. Open the CTAForm component in CI-deployed preview environment
2. Go to CTAForm story in Storybook
3. Verify that the CTAForm behaves as described in the following issue.

- [Figma Design](https://www.figma.com/file/BJ95AjraesmRCWsKA013GS/branch/LSoPddDUuzfy4RLRh4JyPv/Primer-Brand?node-id=4482%3A29680&t=rolx6DbuNYsjYd8h-0)

## Contributor checklist:

## Reviewer checklist:

## Screenshots:

image


packages/react/src/CTAForm/CTAForm.tsx
```diff
@@ -0,0 +1,38 @@
...
+  checkboxLabel: React.ReactNode[] | React.ReactNode | string
+} & React.HTMLAttributes<HTMLFormElement>
+
+export const CTAForm = ({
+  inputLabel,
+  inputType = 'text',
+  placeholder,
+  buttonLabel,
+  checkboxLabel,
+  ...args
```
rezrah, line 21: is `args` a storybook hangover? Can we change to `rest` here?

```suggestion
  ...rest
```

packages/react/src/CTAForm/CTAForm.tsx
```diff
@@ -0,0 +1,38 @@
...
+}: CTAFormProps) => {
+  return (
+    <form {...args} className={clsx(styles.CTAForm, args.className)}>
```
rezrah, line 24:
```suggestion
    <form {...args} className={clsx(styles.CTAForm, rest.className)}>
```

packages/react/src/CTAForm/CTAForm.stories.tsx
```diff
@@ -0,0 +1,47 @@
...
+  title: 'Components/CTAForm',
+  component: CTAForm,
+  args: {
+    inputLabel: 'Your work email address',
+    inputType: 'text',
+    placeholder: 'me@github.com',
+    buttonLabel: 'Subscribe',
+    checkboxLabel: (
+      <>
+        I agree to the <a href="www.github.com">terms and conditions</a> and <a href="www.github.com">privacy policy</a>
```
rezrah, line 16: Can we use InlineLink here, even though it's an example? Link colors are off here:

Screenshot 2023-01-23 at 15 00 50

packages/react/src/CTAForm/CTAForm.tsx
```diff
@@ -0,0 +1,38 @@
...
+
+import styles from './CTAForm.module.css'
+
+type CTAFormProps = {
+  inputLabel: TextInputProps['leadingText']
+  inputType?: TextInputProps['type']
+  placeholder?: TextInputProps['placeholder']
+  buttonLabel: ButtonProps<'button'>['children']
+  checkboxLabel: React.ReactNode[] | React.ReactNode | string
+} & React.HTMLAttributes<HTMLFormElement>
```
rezrah, line 13: This is quite a closed API. Was there a reason for not going with a composable API here? I can't find the API exploration to see the wider conversation.

rezrah, line 13: This is quite a closed API. Was there a reason for not going with a composable API here? I can't find the API exploration to see the wider conversation.

rezrah: E.g. How would I use this form in either controlled or uncontrolled modes? I can't pass a `ref`, or at least it's not clear to me right now how I would grab this from the form submission event. Also the `method` and `action` attributes are forwarded from what I can see, but wondering if something like this might make this clearer.

```jsx
  <CTAForm method="post" action="...">
    <CTAForm.TextInput ref={inputRef} />
    {needsCheckbox && <CTAForm.Confirm label={checkboxLabel} />}
  <CTAForm>
``` 
In the current format, I'm not sure how I can conditionally render the checkbox field, which is important because it's not always required.

apps/docs/content/components/CTAForm.mdx
```diff
@@ -0,0 +1,45 @@
+---
+title: CTAForm
+status: Experimental
+source: https://github.com/primer/brand/tree/main/package/src/CTAForm/CTAForm.tsx
```
rezrah, line 4:
```suggestion
source: https://github.com/primer/brand/tree/main/packages/react/src/CTAForm/CTAForm.tsx
```

apps/docs/content/components/CTAForm.mdx
```diff
@@ -0,0 +1,45 @@
...
+figma: 'https://www.figma.com/file/BJ95AjraesmRCWsKA013GS/Primer-Brand?node-id=1377%3A30754'
```
rezrah, line 5: This doesn't seem to go anywhere? I think it's also linking to an old branch that's since been merged.

apps/docs/content/components/CTAForm.mdx
```diff
@@ -0,0 +1,45 @@
...
+      I agree to the <a href="www.github.com">terms and conditions</a> and{' '}
+      <a href="www.github.com">privacy policy</a>
+    </>
+  }
+  inputLabel="Your work email address"
+  inputType="email"
+  placeholder="me@github.com"
+/>
+```
+
```
rezrah, line 34: Can we have some more examples here please? Like how to conditionally load the checkbox, and remove it from the default state.

rezrah (commented):
This a cool start. I left a few comments to consider, but I'm also lacking API exploration context.

Main feedback is:

- Legal bit should be optional
- The API right now is quite closed, whereas I'd expect it to be compose-able so that it's easy for me to figure out where to add refs and `form` attributes.
- We should have a few more docs examples showing how to do this ☝️ 
- Pr seems to be missing tests (unit and visual)?

packages/react/src/CTAForm/CTAForm.stories.tsx
JoshBowdenConcepts: InlineLink might have a bug because when the screen size is full size then the text grows larger than its surrounding text. It goes to 20px after being 16px on smaller screen sizes. I can't figure out quite why yet.

JoshBowdenConcepts: image
image

packages/react/src/CTAForm/CTAForm.tsx
JoshBowdenConcepts: in next push

packages/react/src/CTAForm/CTAForm.tsx
JoshBowdenConcepts: in next push

JoshBowdenConcepts:
We are currently working through the API exploration to make sure we don't miss any implementation details.

apps/docs/content/components/CTAForm.mdx
```diff
@@ -0,0 +1,121 @@
...
+## Examples
+
+### Default
+
+```jsx live
+<CTAForm>
+  <CTAForm.Input>
+    <FormControl required size="large" fullWidth>
```
rezrah, line 29: I love the composable API 👍. I'm just wondering if some of these critical default values should be baked into the internal logic, while also allowing users to override them if necessary. 

If for example, I remove some of the optional attributes the appearance of the component changes quite dramatically. 

docs example

While we should trust the user to copy the snippet verbatim, some of these feel so crucial to the overarching pattern that the component should be handling them internally for convenience. 

Maybe this is as simple as doing a `find` for a `FormControl` instance in the function for `CTAForm.Input`, and adding default props in.

What do you think?

apps/docs/content/components/CTAForm.mdx
```diff
@@ -0,0 +1,121 @@
...
+### CTAForm
+
+| name        | type     | default     | required | description                                                        |
+| ----------- | -------- | ----------- | -------- | ------------------------------------------------------------------ |
+| `action`    | `string` | `undefined` | `false`  | The URL that processes the form submission                         |
+| `method`    | `string` | `undefined` | `false`  | The HTTP method to submit the form with                            |
+| `target`    | `string` | `undefined` | `false`  | Indicates where to display the response after submitting the form. |
+| `className` | `string` | `undefined` | `false`  | Sets a custom class                                                |
+
+```jsx live
```
rezrah, line 58: Are these snippets supposed to be here? Can we hoist them up to go alongside the other live examples please? This one seems to be showing the component without confirmation child, so guessing it should have it's own header too?

apps/docs/content/components/CTAForm.mdx
```diff
@@ -0,0 +1,121 @@
...
+</CTAForm>
+```
+
+### CTAForm.Action
+
+| name        | type     | default     | required | description         |
+| ----------- | -------- | ----------- | -------- | ------------------- |
+| `className` | `string` | `undefined` | `false`  | Sets a custom class |
+
+{/* TODO: Add all button properties here */}
```
rezrah, line 121: Just a reminder to remove this, it's currently in the docs. Should this be a HTML comment instead?
Screenshot 2023-02-13 at 13 42 29

apps/docs/content/components/CTAForm.mdx
```diff
@@ -0,0 +1,121 @@
...
+<CTAForm>
+  <CTAForm.Input>
+    <FormControl required size="large" fullWidth>
+      <FormControl.Label>Your Work Email Address</FormControl.Label>
+      <TextInput placeholder="name" />
+    </FormControl>
+  </CTAForm.Input>
+  <CTAForm.Action>Subscribe</CTAForm.Action>
+</CTAForm>
+```
```
rezrah, line 113: Think this can be removed. Or was it intentional?

packages/react/src/CTAForm/CTAForm.module.css
```diff
@@ -0,0 +1,47 @@
+.CTAForm {
+  display: grid;
+  gap: var(--base-size-16);
+  grid-template-columns: 2fr auto;
+  align-items: flex-end;
+}
+
+.CTAForm > * {
```
rezrah, line 8: Not sure if things have changed now, but I recall wildcard selectors not being recommended. Can we do this with comma separated class names instead?

packages/react/src/CTAForm/CTAForm.module.css
```diff
@@ -0,0 +1,47 @@
...
+.CTAForm-button {
+  order: 4;
+}
+
+.CTAForm > *:not(.CTAForm-input, .CTAForm-button, .CTAForm-confirm) {
+  order: 3;
+  grid-column: span 2;
+}
+
+@media screen and (min-width: 450px) {
```
rezrah, line 31: Can we use `rem` units that map to our supported breakpoints like we do elsewhere? Or is this custom value intentional?

.changeset/poor-boats-walk.md
```diff
@@ -0,0 +1,27 @@
...
+
+CTAForm Component
+
+Example Usage
+
+```jsx live
+<CTAForm>
+    <CTAForm.Input>
+      <FormControl required size="large" fullWidth>
+        <FormControl.Label>Your Work Email Address</FormControl.Label>
```
rezrah, line 13:
```suggestion
        <FormControl.Label>Your work email address</FormControl.Label>
```

apps/docs/content/components/CTAForm.mdx
```diff
@@ -0,0 +1,121 @@
...
+
+## Examples
+
+### Default
+
+```jsx live
+<CTAForm>
+  <CTAForm.Input>
+    <FormControl required size="large" fullWidth>
+      <FormControl.Label>Your Work Email Address</FormControl.Label>
```
rezrah, line 30:
```suggestion
      <FormControl.Label>Your work email address</FormControl.Label>
```

apps/docs/content/components/CTAForm.mdx
```diff
@@ -0,0 +1,121 @@
...
+        I agree to the <InlineLink href="#">Privacy Policy</InlineLink> and{' '}
+        <InlineLink href="#">Terms of Service</InlineLink>
+      </FormControl.Label>
+      <Checkbox name="confirm" />
+    </FormControl>
+  </CTAForm.Confirm>
+  <CTAForm.Action>Subscribe</CTAForm.Action>
+</CTAForm>
+```
+
```
rezrah, line 46: Nice to have: some additional examples for controlled / uncontrolled mode, so it's super clear and easy how to use this pattern.

packages/react/src/CTAForm/CTAForm.stories.tsx
```diff
@@ -0,0 +1,46 @@
...
+export default {
+  title: 'Components/CTAForm',
+  component: CTAForm
+} as ComponentMeta<typeof CTAForm>
+
+export const Playground: ComponentStory<typeof CTAForm> = _args => (
+  <CTAForm>
+    <CTAForm.Input>
+      <FormControl required size="large" fullWidth>
+        <FormControl.Label>Your Work Email Address</FormControl.Label>
```
rezrah, line 16:
```suggestion
        <FormControl.Label>Your work email address</FormControl.Label>
```

.changeset/poor-boats-walk.md
```diff
@@ -0,0 +1,27 @@
...
+        <TextInput placeholder="name" />
+      </FormControl>
+    </CTAForm.Input>
+    <CTAForm.Confirm>
+      <FormControl required size="large" fullWidth>
+        <FormControl.Label>I agree to the Privacy Policy and Terms of Service</FormControl.Label>
```
rezrah, line 19: More a design question, but this is rendering quite heavy compared to how we use this in practice.

E.g.

Screenshot 2023-02-13 at 14 12 57

cc. @danielguillan

packages/react/src/CTAForm/CTAForm.stories.tsx
```diff
@@ -0,0 +1,46 @@
...
+    </CTAForm.Confirm>
+    <CTAForm.Action>Subscribe</CTAForm.Action>
+  </CTAForm>
+)
+
+export const Default: ComponentStory<typeof CTAForm> = _args => (
+  <CTAForm>
+    <CTAForm.Input>
+      <FormControl required size="large" fullWidth>
+        <FormControl.Label>Your Work Email Address</FormControl.Label>
```
rezrah, line 34:
```suggestion
        <FormControl.Label>Your work email address</FormControl.Label>
```

packages/react/src/CTAForm/CTAForm.test.tsx
```diff
@@ -0,0 +1,167 @@
...
+    const {getByTestId} = render(<CTAForm.Confirm data-testid={mockTestId} className={expectedClass}></CTAForm.Confirm>)
+    const ctaConfirmEl = getByTestId(mockTestId)
+    expect(ctaConfirmEl.classList).toContain(expectedClass)
+  })
+})
+
+describe('CTAForm.Action', () => {
+  afterEach(cleanup)
+
+  it('no a11y violations', async () => {
```
rezrah, line 119: Do we need to test each child individually? Wonder if these tests can be rolled into the main suite at the top?

packages/react/src/CTAForm/CTAForm.test.tsx
```diff
@@ -0,0 +1,167 @@
...
+    const mockClick = jest.fn(e => e.preventDefault())
+
+    const {getByTestId} = render(
+      <CTAForm.Action data-testid={mockTestId} className={expectedClass} onClick={mockClick}>
+        Submit
+      </CTAForm.Action>
+    )
+    const ctaButtonEl = getByTestId(mockTestId)
+    expect(ctaButtonEl).toHaveAttribute('type', 'submit')
+    fireEvent(
```
rezrah, line 138: 🙌

packages/react/src/CTAForm/CTAForm.tsx
```diff
@@ -0,0 +1,50 @@
+import React, {forwardRef, type Ref} from 'react'
+import clsx from 'clsx'
+import {Button} from '../'
+
+import styles from './CTAForm.module.css'
+
+const Root = forwardRef(({...rest}: React.FormHTMLAttributes<HTMLFormElement>, ref: Ref<HTMLFormElement>) => {
+  return <form {...rest} ref={ref} className={clsx(styles.CTAForm, rest.className)} />
```
rezrah, line 8: nit: can we destructure the className in the fn arguments instead of using `rest.className`? Also moving the spread to the end?


```suggestion
const Root = forwardRef(({className, ...rest}: React.FormHTMLAttributes<HTMLFormElement>, ref: Ref<HTMLFormElement>) => {
  return <form ref={ref} className={clsx(styles.CTAForm, className)} {...rest}  />
```

apps/docs/content/components/CTAForm.mdx
```diff
@@ -0,0 +1,121 @@
...
+    <FormControl required size="large" fullWidth>
+      <FormControl.Label>Your Work Email Address</FormControl.Label>
+      <TextInput placeholder="name" />
+    </FormControl>
+  </CTAForm.Input>
+  <CTAForm.Confirm>
+    <FormControl required size="large" fullWidth>
+      <FormControl.Label>
+        I agree to the <InlineLink href="#">Privacy Policy</InlineLink> and{' '}
+        <InlineLink href="#">Terms of Service</InlineLink>
```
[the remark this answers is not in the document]
JoshBowdenConcepts: I believe this is an issue with the InlineLink component. Should I just remove it from the docs for the CTAForm?

apps/docs/content/components/CTAForm.mdx
JoshBowdenConcepts: I think that is a great idea! I would imagine setting `fullWidth` and `size='large'` by default is there anything else I am missing there?

apps/docs/content/components/CTAForm.mdx
JoshBowdenConcepts: I think that was the intention I will make that update.

apps/docs/content/components/CTAForm.mdx
JoshBowdenConcepts: This is a good note. I wanted to document somewhere that when the Button docs get filled out this should be updated. Maybe I could include it on the issue for that work. I believe Tyler will be taking that on correct?

apps/docs/content/components/CTAForm.mdx
```diff
@@ -0,0 +1,121 @@
...
+
+### CTAForm
+
+| name        | type     | default     | required | description                                                        |
+| ----------- | -------- | ----------- | -------- | ------------------------------------------------------------------ |
+| `action`    | `string` | `undefined` | `false`  | The URL that processes the form submission                         |
+| `method`    | `string` | `undefined` | `false`  | The HTTP method to submit the form with                            |
+| `target`    | `string` | `undefined` | `false`  | Indicates where to display the response after submitting the form. |
+| `className` | `string` | `undefined` | `false`  | Sets a custom class                                                |
+
```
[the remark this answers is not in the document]
JoshBowdenConcepts: Definitely I like that approach.

packages/react/src/CTAForm/CTAForm.module.css
JoshBowdenConcepts: I hear what you are saying and I would agree that wildcard selectors could be dangerous in some cases, but for this I believe that they would be both safe and effective so that users could include their own children without issue. What this specific declaration is doing is setting all children to span the full size of the grid. When a user adds and additional input that would be added with this css and the size would be as they might expect it to be. Does that make sense and what would your recommendation be?

packages/react/src/CTAForm/CTAForm.test.tsx
JoshBowdenConcepts: It depends on how logging works. Would it roll up all of the accessibility issues into the main test block? Would it be easier for a user to view them this way or individually by component piece?

apps/docs/content/components/CTAForm.mdx
JoshBowdenConcepts: What were you imagining here for this.

apps/docs/content/components/CTAForm.mdx
rezrah: nope, just those two based on the code snippet. I'd be surprised if `required` wasn't always necessary on the form input too 🤔.

apps/docs/content/components/CTAForm.mdx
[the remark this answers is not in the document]
rezrah: I think we should fix it if we can. InlineLink seems to take a `size` prop, can we match the text size that way?

apps/docs/content/components/CTAForm.mdx
rezrah: Adding to issue sounds good. We shouldn't render these as markup in the user-facing docs. If you can turn it into a comment that'd be great 👍

apps/docs/content/components/CTAForm.mdx
rezrah: Similar to your test example, where you provide a submission handler that can trigger an alert or similar. E.g from PRC for controlled mode: https://primer.style/react/CheckboxGroup#using-onchange-handlers

packages/react/src/CTAForm/CTAForm.test.tsx
rezrah: > It depends on how logging works. 

sorry, which logging? 

I was referring to the approach of using different describe blocks for testing each child independently. I haven't seen that done elsewhere, so wondering what the motivation was for doing it here. I don't think it necessarily has a user impact 🤔, but it could IMO make it harder to maintain those test fails given how much longer they are. If we scale it out to other components also, it could make our suites slower to complete, so wondering what the benefits were.

Another reason for asking, is that the child components aren't meant to be used in isolation, but rather part of the wider CTAForm. This is usually important context for axe runners.

packages/react/src/CTAForm/CTAForm.module.css
rezrah: Thanks for clarifying, I can see how custom css classes won't serve the use-case for unknown children.

I can't think of another option besides using the ~ sibling selector for tighter specificity, which I'm guessing you've already considering. It ultimately would achieve the same, so happy to also stick with this 👍

apps/docs/content/components/CTAForm.mdx
```diff
@@ -0,0 +1,91 @@
...
+```jsx live
+<CTAForm>
+  <CTAForm.Input>
+    <FormControl required size="large" fullWidth>
+      <FormControl.Label>Your work email address</FormControl.Label>
+      <TextInput placeholder="name" />
+    </FormControl>
+  </CTAForm.Input>
+  <CTAForm.Confirm>
```
rezrah, line 34: Just an FYI, that the button and text input aren't quite aligned.. probably something we need to address with the tokens so if it turns into a big deal we can always create a separate issue for it 

image

packages/react/src/CTAForm/CTAForm.features.stories.tsx
```diff
@@ -0,0 +1,24 @@
...
+
+import {CTAForm} from './CTAForm'
+import {FormControl, TextInput} from '../index'
+
+export default {
+  title: 'Components/CTAForm/Features',
+  component: CTAForm
+} as ComponentMeta<typeof CTAForm>
+
+// TODO: When there is no confirm the gap should disappear
```
rezrah, line 12: Think this can be removed now? I can't see the gap.

.changeset/poor-boats-walk.md
josepmartins: It looks good! and that seems to be aligned with the latest Galaxy designs as well. 

Screenshot 2023-02-14 at 12 18 11

apps/docs/content/components/CTAForm.mdx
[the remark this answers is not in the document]
JoshBowdenConcepts: I tried that and it seems like it scales in different view ports which is what broke the code. In mobile it looked correct but in desktop it was an issue.

apps/docs/content/components/CTAForm.mdx
JoshBowdenConcepts: Thanks for the example I will add that in.

packages/react/src/CTAForm/CTAForm.test.tsx
JoshBowdenConcepts: I would be happy to consolidate for consistency!

packages/react/src/CTAForm/CTAForm.module.css
JoshBowdenConcepts: I am happy to update this to the rem values. I think I noticed pixels being used in the SubdomainNavBar where 768px should be 48rem. Would you like me to update these as well?

.changeset/poor-boats-walk.md
rezrah: @josepmartins - sorry perhaps wasn't clear. The text in Galaxy is using a non-bold font weight, vs the Primer Brand version, which uses bold. Do we need to ask for Galaxy to be changed? IMO the bold should be limited to shorter phrases that require emphasis, rather than paragraphs like the typical use-case for this component.

packages/react/src/CTAForm/CTAForm.module.css
JoshBowdenConcepts: We should add a style-lint rule that looks for our breakpoints and suggests the right rem values. That could be a really cool addition. Maybe a passion week kind of task.

apps/docs/content/components/CTAForm.mdx
JoshBowdenConcepts: It looks like it is off because the button includes the border in its size and the input doesn't so while the height is the  "same" in the inspection panel the actual rendered hight is different. How do we determine which is the one that needs to be updated?

apps/docs/content/components/CTAForm.mdx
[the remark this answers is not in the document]
rezrah: The viewport scaling is coming from the underlying Text component, if the non-link parts of this label aren't behaving in a similar way, I recommend making them do just that. As a last resort, apply a custom style via additional class? I don't think we can go ahead with this example as it is right now, given the sizes are noticeably different.

packages/react/src/CTAForm/CTAForm.module.css
rezrah: > SubdomainNavBar where 768px should be 48rem

This is an oversight, feel free to fix it in this PR or it can be done separately. Either way, 48rem (768px) is one of our standard breakpoints, which we have a design token for. I don't think 450px is a standard breakpoint? It converts to `28.125rem`.

.changeset/poor-boats-walk.md
rezrah: Update: @josepmartins and I spoke about this separately and agreed to move forward with non-bold font. @josepmartins is planning to update Figma. Reason for this change is that the label text is typically going to be multiline, so emphasising the entire paragraph can make it harder to read.

apps/docs/content/components/CTAForm.mdx
rezrah: I believe the text input is the one that should change. That said, I also think this should be a separate issue, which we can discuss with Dani and Josep too.

JoshBowdenConcepts: Ok I will leave it out of this PR for now but it will be good to note.

packages/react/src/CTAForm/CTAForm.module.css
JoshBowdenConcepts: It wasn't and I updated that because it also was supposed to be 48rem. I can include those updates to this PR just to help keep the repo consistant. Some of the values in the SubdomainNavBar seem to be custom but I can check as I update.

apps/docs/content/components/CTAForm.mdx
rezrah: I think it's trickier for required, because it's a functional change, vs aesthetic. It may not be immediately clear how to turn off required, if anyone doesn't want it. 

I'm happy to add it in. Can we add a note in the docs to indicate that we set required on the input by default, and that to turn it off they'll need to pass required={false}?

packages/react/src/CTAForm/CTAForm.module.css
rezrah: Cool, sometimes we may need to use custom values for spacing / component-specific reasons. That's why I was asking about 450px in case there was a reason we were using non-specific values. If there isn't, best to default to our breakpoint tokens.

.changeset/poor-boats-walk.md
josepmartins: Yep! Just updated the Figma files in case you want to check them out, but in short:

- Font-weight is now using 400 or `--base-text-weight-normal`
- Color is now `--brand-color-text-muted`

JoshBowdenConcepts: Is this something that should be updated in the FormControl component or in this example specifically?

josepmartins: I'd say only for this example specifically. `FormControl` uses a semibold weight as we assume that labels will be shorter there + it allows a `FormControl.Hint` to get more control over styles and long-length descriptive texts.

Screenshot 2023-02-15 at 08 52 26

JoshBowdenConcepts: Luckily it seems anything we want to create is possible because the label is composable.

Does this mean that my current example is correct in term of its look and use in our documentation?

josepmartins: I'd say it's missing the variant="muted" in the Text component

```jsx
<Text size="300" variant="muted">....</Text>
```

JoshBowdenConcepts: Perfect I the docs and they should align now.

apps/docs/content/components/CTAForm.mdx
```diff
@@ -0,0 +1,110 @@
...
+  </CTAForm.Input>
+  <CTAForm.Action>Subscribe</CTAForm.Action>
+</CTAForm>
+```
+
+### Using onChange & onSubmit handlers
+
+```jsx live
+<CTAForm onSubmit={() => alert('onSubmit event triggered')}>
+  <CTAForm.Input onChange={() => alert('onChange event triggered')}>
```
rezrah, line 69: I think using a console.log or on-screen counter would be more ideal here, as it's currently firing the alert on every key press and breaking momentum of the user input. The alert would perhaps be more suitable for the submit handler only?

https://user-images.githubusercontent.com/13340707/219355159-e6b43c77-4fb5-4e8d-a2f8-8b5df7287465.mov

packages/react/src/CTAForm/CTAForm.stories.tsx
```diff
@@ -0,0 +1,56 @@
...
+      <FormControl required>
+        <FormControl.Label>Your Work Email Address</FormControl.Label>
+        <TextInput placeholder="name" />
+      </FormControl>
+    </CTAForm.Input>
+    <CTAForm.Confirm>
+      <FormControl required>
+        <FormControl.Label>
+          <Text size="300" variant="muted">
+            I agree to the <InlineLink href="www.github.com">Privacy Policy</InlineLink> and{' '}
```
rezrah, line 24: Will 404 for anyone that tries to click this currently.

```suggestion
            I agree to the <InlineLink href="https://github.com">Privacy Policy</InlineLink> and{' '}
```

packages/react/src/CTAForm/CTAForm.stories.tsx
```diff
@@ -0,0 +1,56 @@
...
+            <InlineLink href="www.github.com">Terms of Service</InlineLink>
```
rezrah, line 25: <InlineLink href="https://www.github.com">Terms of Service</InlineLink>

rezrah (approved):
Looks great, thanks for going through all of the feedback @JoshBowdenConcepts ✨ 

- Left a few more super-minor items of feedback for you
- FYI, you're visual tests seem to be stale, so I've re-triggered it for you. Should hopefully be done by time you see this.

PR #170, opened 2023-02-01
Brand - Button accessibility updates

TylerJDev:
## Summary

Adds accessibility fixes related to [Brand-Button accessibility review](https://github.com/github/primer/issues/1682).

## List of notable changes:

- Added outline on focus when only `aria-disabled` is applied. 
- Added focus indicator and 'disabled' style for Windows High Contrast Mode (WHCM).
- Added use of 'disabled' and 'aria-disabled' separately.
- Hid SVG from screen readers. 

- `aria-disabled` and `disabled` now being separated.
- Styles for button for when `aria-disabled` is applied (focus indication, WHCM support)

## Steps to test:

1. Open up component and use `aria-disabled` attribute control on button. Ensure element is focusable with only `aria-disabled` being applied.
2. Test the same as above, but with the "`disabled`" control, ensure that element is not focusable.
3. See if WHCM styles are applied. 
  (This can be tested within Chrome without Windows through [Chrome Devtools via the "Rendering" tab](https://developer.chrome.com/docs/devtools/rendering/emulate-css/#emulate-css-media-feature-forced-colors). Within the "Emulate CSS media feature forced-colors" section, select the "forced-colors: active" option in the select menu).

- [Accessibility engineer review for Primer/Brand Button](https://github.com/github/primer/issues/1682)
- [Windows High Contrast Mode & development](https://blogs.windows.com/msedgedev/2020/09/17/styling-for-windows-high-contrast-with-new-standards-for-forced-colors/)

## Contributor checklist:

## Reviewer checklist:

## Screenshots:

<table>
<tr>
<th> Before</th> <th> After </th> <th> Why </th>
</tr>
<tr>
<td valign="top">

Primer Brand button with "disabled" and "aria-disabled" applied, shows a dimmed button without a focus outline.

`disabled` and `aria-disabled` applied, no focus indication is present (due to `disabled` attribute, and existing CSS rule).
 </td>

<td valign="top">

Primer Brand button with only "aria-disabled" applied, shows a dimmed button with a focus outline on focus.

Only `aria-disabled` is applied in this example, retains focus indicator due to element still being focusable.
</td>

<td valign="middle">

We'd want to keep the focus outline on the button only when `aria-disabled` is applied. This is because `aria-disabled` will keep the button focusable.

This keeps us in check with [WCAG 2.4.7 Focus visible](https://www.w3.org/WAI/WCAG22/Understanding/focus-Visible.html).
</td>
</tr>

<tr>
<td valign="top">

Primer Brand button within Windows High Contrast Mode (emulated), showing a button with no focus indicator while being actively focused.

Focused button (`:focus`) within WHCM.

</td>
<td valign="top">

Primer Brand button within Windows High Contrast Mode (emulated), showing a button with a visible focus indicator while being focused.

Focused button (`:focus`), within WHCM. Focus outline is present via `outline` CSS rule.

</td>
<td valign="middle">

WHCM by default will strip `box-shadow` from elements visually, meaning that our current focus indicator that uses `box-shadow` needs to be supplemented with an `outline` style. This new rule will only be applied within WHCM (via `@media (forced-colors: active)`).
</td>
</tr>
</table>


.changeset/odd-penguins-film.md
```diff
@@ -0,0 +1,10 @@
+---
+'@primer/react-brand': patch
+---
+
+Adds accessibility fixes.
```
rezrah, line 5: Suggestion for adding some clarity on which component this update will impact:

```suggestion
Added accessibility fixes for `Button`
```

packages/react/src/Button/Button.features.stories.tsx
```diff
@@ -16,7 +16,7 @@ export const Primary = () => (
 )
 
 export const PrimaryDisabled = () => (
-  <Button as="a" disabled variant="primary" href="#">
+  <Button as="button" disabled variant="primary" href="#">
```
rezrah, line 19: Can we remove the `href` in this example too if we're using `button`?

```suggestion
  <Button as="button" disabled variant="primary">
```

packages/react/src/Button/Button.module.css
```diff
@@ -113,3 +113,17 @@
 .Button--disabled:not(.Button--secondary) {
   box-shadow: none;
 }
+
+@media (forced-colors: active) {
```
rezrah, line 117: TIL. Thanks for adding this, I suspect we'll probably need to do the same in a few other components 😓

packages/react/src/Button/Button.module.css
```diff
@@ -113,3 +113,17 @@
...
+  /* Retains focus indicator when in "Windows High Contrast Mode". */
+  .Button:focus:not(.Button--disabled[disabled]) {
+    outline: 3px solid;
+    outline-color: var(--brand-color-focus);
+    outline-offset: 2px;
+  }
+
+  /* Indicates disabled state in WHCM */
+  .Button--label.Button-label--disabled {
+    color: GrayText;
```
rezrah, line 127: Question: Are our design tokens not suitable here? Or if `GrayText` is required, should we have a design token that maps to this?

packages/react/src/Button/Button.module.css
```diff
@@ -113,3 +113,17 @@
 .Button--disabled:not(.Button--secondary) {
   box-shadow: none;
 }
+
+@media (forced-colors: active) {
+  /* Retains focus indicator when in "Windows High Contrast Mode". */
+  .Button:focus:not(.Button--disabled[disabled]) {
+    outline: 3px solid;
```
rezrah, line 120: FYI, we have some tokens that could perhaps replace the `px` variant?

E.g. 

```css
--brand-borderWidth-thick
--brand-borderWidth-thicker
```

packages/react/src/Button/Button.tsx
```diff
@@ -16,6 +16,7 @@ export type ButtonBaseProps = {
   variant?: typeof ButtonVariants[number]
   size?: typeof ButtonSizes[number]
   hasArrow?: boolean
+  ariaDisabled?: boolean
```
rezrah, line 19: Given that Button is a wrapper around HTMLElement, could it be confusing that we're redeclaring this rather than relying on the inherited attribute?

rezrah (commented):
This is great, thank you @TylerJDev 🙌 

I've left some super minor feedback for you. Specifically the ones around removing the leftover `href` in the code snippets.

.changeset/odd-penguins-film.md
TylerJDev: Good call! Should be adjusted now.

packages/react/src/Button/Button.features.stories.tsx
TylerJDev: Definitely, good call! I removed the `href` if the component was either the default (button) or `as="button"`.

packages/react/src/Button/Button.module.css
TylerJDev: No worries! I think it'll be a case by case basis for now. I'll have to do some research to see if there's a more global way we can apply these rules.

packages/react/src/Button/Button.module.css
TylerJDev: I'm not too sure. `GrayText` here only tells WHCM that the text should be styled as if it's disabled. I wonder if we'd benefit from adding this as a token or not? As of right now, I don't think we'll be using it for anything other than color for text that should be disabled. 

Would love to know your thoughts on this.

packages/react/src/Button/Button.module.css
TylerJDev: Ahh good to know! I replaced this with `--brand-borderWidth-thick` and made note of our tokens.

packages/react/src/Button/Button.tsx
TylerJDev: That makes sense. I've adjusted it so that we rely on `aria-disabled` instead!

packages/react/src/Button/Button.tsx
```diff
@@ -45,44 +47,47 @@ export const Button = forwardRef(
     const [isFocused, setIsFocused] = React.useState(false)
     const Component = as || 'button'
 
+    const isDisabled =
```
TylerJDev, line 50: Added a new variable `isDisabled`. Since `aria-disabled` can take a value of `string` or `boolean` (e.g. `aria-disabled="true"` === `aria-disabled={true}`) I made a conditional to verify that it either has a string value of `"true"` or a boolean value of `true`.

TylerJDev:
@rezrah - Thanks a bunch for checking out this PR! I addressed all your comments, and adjusted the PR slightly. I've included a test for `Button` too. Let me know if you have any questions, or more feedback!

packages/react/src/Button/Button.module.css
rezrah: That's good to know. I don't think this needs to be token in that case. It will create unnecessary obfuscation and also isn't something we'd want users to customise from sounds of it.

rezrah (approved):
Looks great, ship it :shipit:

PR #171, opened 2023-02-01
Test FAQ with anchored links

josepmartins:
## Summary

https://github.com/github/primer/issues/1749

## List of notable changes:

-
-

-
-

## Steps to test:

1.
1.
1.

-
-

## Contributor checklist:

## Reviewer checklist:

## Screenshots:

<table>
<tr>
</tr>
<tr>
<td valign="top">

 </td>
<td valign="top">

</td>
</tr>
</table>

### DIFF
--- a/packages/react/src/FAQ/FAQ.stories.tsx
+++ b/packages/react/src/FAQ/FAQ.stories.tsx
@@ -82,5 +82,111 @@ export const Default = () => (
         </p>
       </FAQ.Answer>
     </FAQ.Item>
+    <FAQ.Item>
+      <FAQ.Question>What&apos;s included in the GitHub for Startups offer?</FAQ.Question>
+      <FAQ.Answer>
+        <p>
+          All GitHub for Startups companies receive up to 20 seats of GitHub Enterprise for free for year one and 50%
+          off year two. Learn more about the features and capabilities of GitHub Enterprise{' '}
+          <InlineLink href="https://copilot.github.com/" target="_blank" rel="noreferrer">
+            here
+          </InlineLink>
+          .
+        </p>
+      </FAQ.Answer>
+    </FAQ.Item>
+    <FAQ.Item>
+      <FAQ.Question>Who is eligible to apply?</FAQ.Question>
+      <FAQ.Answer>
+        <p>Startups who meet the following criteria are eligible to apply for the program:</p>
+        <ol>
+          <li>
+            <Text size="300" variant="muted">
+              Must be associated with a current GitHub for Startups partner.
+            </Text>
+          </li>
+          <li>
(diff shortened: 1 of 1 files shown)


packages/react/src/FAQ/FAQ.stories.tsx
```diff
@@ -82,5 +82,111 @@ export const Default = () => (
...
+          If you’re not currently eligible for the GitHub for Startups but would like to try GitHub Enterprise, please
+          feel to sign up for a trial
+          <InlineLink href="https://copilot.github.com/" target="_blank" rel="noreferrer">
+            here
+          </InlineLink>
+          .
+        </p>
+      </FAQ.Answer>
+    </FAQ.Item>
+    <FAQ.Item open id="test">
```
rezrah, line 139: Sorry for the drive-by comment @josepmartins - this looked interesting and I was also thinking about this yesterday so hope you don't mind me dropping some feedback here? 😬 

I think in practice we'd typically not want to set `open` attribute by default in the markup, but rather only when required via a "deeplink" url. What do you think about passing a query param, which FAQ component reads, parses and then adds the `open` attribute to the id of the relevant question?

We can then trigger the opening of the question via the URL like:

`mywebsite.com?faq=1` which opens `id="1"`

josepmartins: I totally agree on not having the `open` attribute by default. This was more of a test to showcase how the feature "should look", that's why I forced the open state. 

Not sure about the need for a query param here (or maybe I didn't understand your proposal lol). Wouldn't it be better to use normal anchor links instead (`mywebsite.com#whats-the-meaning-of-life`)  that scroll down to the id of the question I and adds the `open` attribute to it?

josepmartins: To illustrate my comment above, made a commit as an example using the `location.hash` to adjust the open boolean in FAQ: 
https://github.com/primer/brand/pull/171/commits/fdd3e2a204d6d1b349190cf91bc4c50ac864aa7a

rezrah: 🙌 I love that idea. Glad we're aligned on the optional attribute. I think your approach of doing it without query params makes total sense too. Just bear in mind that the component will need to run any `window` operations in a `useEffect` to mitigate SSR constraints, otherwise works for me 👍

josepmartins: I made a `useHash` webhook to handle the `window.hash` operations, which works pretty fine


https://user-images.githubusercontent.com/912236/216101315-b97389a9-d9c6-46c7-b761-b7f748cb9f83.mov

PR #194, opened 2023-03-01
Add arrowDirection prop to Link component

rezrah:
## Summary

Allows reversal of the arrow in `Link` for use-cases where we need to send users back to a previous page.

🔗 [Storybook](https://primer-69e32cb976-26139705.drafts.github.io/storybook/?path=/story/components-link--reversed)

- `arrowDirection` prop naming
- review implementation

- Required for https://github.com/github/galaxy.github.com/issues/123

## Contributor checklist:

## Reviewer checklist:

## Screenshots:
Back to schedule Link with reversed arrow

https://user-images.githubusercontent.com/13340707/222214654-5cec0703-141d-49e9-8e45-46316538e67d.mov


packages/react/src/Link/Link.tsx
```diff
@@ -6,7 +6,14 @@ import {Text} from '../Text'
...
 
 export type LinkProps = {
+  /**
+   * The size variations available in Link
+   */
   size?: 'medium' | 'large'
+  /**
+   * Displays the arrow icon in the opposite direction to the default
+   */
+  reverseArrow?: boolean
```
colebemis, line 16: How do you feel about calling this `arrowDirection` and the value could be either be `left` or `right`?

colebemis (approved):
Implementation looks good. Left small comment about the prop naming

packages/react/src/Link/Link.tsx
rezrah: I like that idea! It's much better for internationalisation too, as we're not implying one is more correct than the other, particularly in languages that read RTL.

I've changed to arrowDirection 👍

packages/react/src/Link/Link.tsx
```diff
@@ -65,17 +73,22 @@ export function Link({
...
-      className={clsx(styles.Link, styles[`Link--${size}`], className)}
+      className={clsx(styles.Link, styles[`Link--${size}`], styles[`Link--arrow-${arrowDirection}`], className)}
       onMouseEnter={handleMouseEnter}
       onMouseLeave={handleMouseLeave}
       onFocus={handleOnFocus}
       onBlur={handleOnBlur}
       {...props}
     >
+      {arrowDirection === 'start' && (
+        <ExpandableArrow className={styles['Link-arrow']} expanded={isHovered || isFocused} reverse />
```
TylerJDev, line 84: Could we add the `hidden` prop to both `ExpandableArrow` SVGs? We'd hide them since they are mostly decorative.

TylerJDev (approved):
Looks good for accessibility! I added one small comment on something we can add, but it's non-blocking.

packages/react/src/Link/Link.tsx
rezrah: Yep, added via the Link prop. Thanks Tyler.

PR #205, opened 2023-03-13
Add new ActionMenu component

rezrah:
## Summary

Adds first-pass for an accessible `ActionMenu` component.

**Try it out:**

🔗 [Storybook](https://primer-0a563ca881-26139705.drafts.github.io/storybook/?path=/story/components-actionmenu-features--single-selection)
🔗 [Documentation](https://primer-0a563ca881-26139705.drafts.github.io/components/ActionMenu)

- Design accuracy
- React implementation

## Steps to test:

1. Review Storybook examples
1. Use keyboard navigation
1. Review code

- [Figma specifications](https://www.figma.com/file/BJ95AjraesmRCWsKA013GS/Primer-Brand?node-id=4912%3A32799)
- [API exploration](https://www.figma.com/file/HcwslO0t21VmPcnWyHLeYH/Primer-Brand---ActionMenu-API-exploration?node-id=0%3A1&t=yTFTTMzT5ffEXjMG-0)

## Contributor checklist:

## Reviewer checklist:

## Screenshots:

> **Note**
> This ActionMenu supports `single`  and `menu`-like selection only at this time. Support for `multiple` selection will be added at a future date.

menu open first item focussed

WAI-ARIA compliant markup and keyboard navigation, verified through E2E tests:

https://user-images.githubusercontent.com/13340707/224698900-4a634d11-7a28-4a46-b4fc-dff406cf56ec.mov


danielguillan:
From a design standpoint, there two minor things to fix:

- The height of the trigger should be `48px`. It's currently `49px` in its default state and `54px` when active. It should match the secondary default Button.
- An extra `24px` of padding is missing at the right of the items to balance the single selection checkmark visually.

When testing accessibility and keyboard navigation I found the following problems:
- As per the a11y design review: 
  - We should probably not enable the spacebar for making selections, as it might increase the chance of accidental activation.
  - The trigger doesn't need `aria-expanded="true" when not activated.
  - I believe we're missing an accessible name for the `ul` that does not contain the selected value.
- Keyboard navigation isn't working for me on the documentation site (Chrome, Safari), and several storybook examples, such as the "In Stack" one. I can focus the trigger but pressing `enter` or `space` doesn't activate the menu. If I activated those using a mouse click, I cannot navigate with the keyboard either.

We might need to introduce an option to align the menu to the trigger's `start`, `center`, and `end`. So we can have right-aligned control as in the galaxy website.

Finally, when the viewport is shorter than the menu, the overlay gets positioned to the side of the trigger. Do you think this is intentional? In some situations, the content overflows the viewport. 

The menu is overflowing the viewport, and some options are not visible anymore

apps/docs/content/components/ActionMenu.mdx
```diff
@@ -0,0 +1,103 @@
...
+render(<App />)
+```
+
+## Component props
+
+### ActionMenu <Label>Required</Label>
+
+| name               | type                                                          | default | required | description                                          |
+| ------------------ | ------------------------------------------------------------- | ------- | -------- | ---------------------------------------------------- |
+| `disabled`         | `boolean`                                                     | `false  | `false`  | Controls the ActionMenu active/inactive state        |
```
josepmartins, line 73:
```suggestion
| `disabled`         | `boolean`                                                     | `false`  | `false`  | Controls the ActionMenu active/inactive state        |
```

apps/docs/content/components/ActionMenu.mdx
```diff
@@ -0,0 +1,103 @@
...
+| `open`             | `boolean`                                                     | `false  | `false`  | Determines whether the ActionMenu is open by default |
```
josepmartins, line 74:
```suggestion
| `open`             | `boolean`                                                     | `false`  | `false`  | Determines whether the ActionMenu is open by default |
```

apps/docs/content/components/ActionMenu.mdx
```diff
@@ -0,0 +1,103 @@
...
+| `onSelect`         | `(newValue: string) => void`                                  |         | `false`  | Callback that is called when an item is selected     |
+| `selectionVariant` | <PropTableValues values={['single', 'none']} addLineBreaks /> |         | `single` | The selection variant of the ActionMenu              |
```
josepmartins, line 76:
```suggestion
| `selectionVariant` | <PropTableValues values={['single', 'none']} addLineBreaks /> | `single` |   | The selection variant of the ActionMenu              |
```

packages/react/src/ActionMenu/ActionMenu.tsx
```diff
@@ -0,0 +1,357 @@
...
+   * Callback that is called when an item is selected.
+   * The value of the selected item is passed as an argument.
+   */
+  onSelect?: (newValue: string) => void
+  /**
+   * The selection variant of the ActionMenu.
+   * - `single`: Only one item can be selected at a time. Menu items as a menuitemradio.
+   * - `none`: No items can be selected. Menu items as a menuitem.
+   */
+  selectionVariant?: 'single' | 'none'
```
josepmartins, line 79: should this be `none` by default?

packages/react/src/ActionMenu/ActionMenu.tsx
```diff
@@ -0,0 +1,357 @@
...
+  handler?: (newValue: string) => void
+  selected?: boolean
+  type?: 'none' | 'single'
+  disabled?: boolean
+} & PropsWithChildren<React.HTMLProps<HTMLLIElement>>
+
+const roleTypeMap = {
+  none: 'menuitem',
+  single: 'menuitemradio'
+  // multiple: 'menuitemcheckbox'
```
josepmartins, line 281: 👋

packages/react/src/ActionMenu/ActionMenu.tsx
```diff
@@ -0,0 +1,357 @@
...
+              className={clsx(styles['ActionMenu__item-icon'], disabled && styles['ActionMenu__item-icon--disabled'])}
+              size={16}
+            />
+          )}
+        </span>
+      )}
+
+      <span className={styles['ActionMenu__item-text']}>
+        <Text
+          variant={disabled ? 'muted' : 'default'}
```
josepmartins, line 319: are `disabled` items necessary for Brand purposes?

packages/react/src/ActionMenu/ActionMenu.tsx
rezrah: Good question and thanks for bringing this up. I'm inclined to agree, because this is now an ActionList. The original intent for this component however, was to behave as a SelectMenu first, which is why it's still defaulting to single.

packages/react/src/ActionMenu/ActionMenu.tsx
rezrah: I can see this being useful in future. Particularly with dynamic data, where we would otherwise need to re-render a filtered set of data which is worse for accessibility and ux. Can we support this in Figma too?

rezrah:
@danielguillan - thanks for the feedback ✨ .. left some notes for you below.

> * An extra `24px` of padding is missing at the right of the items to balance the single selection checkmark visually.

✅ Added 


>   * We should probably not enable the spacebar for making selections, as it might increase the chance of accidental activation.

✅ Now only enter selection is valid

>   * The trigger doesn't need `aria-expanded="true" when not activated.

Not changed, following our discussion on Slack to keep functionality as-is, which is still valid.

>   * I believe we're missing an accessible name for the `ul` that does not contain the selected value.

`aria-label` is now required for ActionMenu.Button. This value is then propagated down to the menu to `aria-labelledby`.

> * Keyboard navigation isn't working for me on the documentation site (Chrome, Safari), and several storybook examples, such as the "In Stack" one. I can focus the trigger but pressing `enter` or `space` doesn't activate the menu. If I activated those using a mouse click, I cannot navigate with the keyboard either.

✅ Fixed, there was a ref issue that's been resolved. 

> We might need to introduce an option to align the menu to the trigger's `start`, `center`, and `end`. So we can have right-aligned control as in the galaxy website.

✅ New prop added called `menuAlignment`. Available options are `start` and `end`. No `center`, because it's not supported in Primer behaviors. Also think we're unlikely to use that much. 

> 
> Finally, when the viewport is shorter than the menu, the overlay gets positioned to the side of the trigger. Do you think this is intentional? In some situations, the content overflows the viewport.

✅ Locked out of bounds, which is the most practical option and consistent with other ActionMenu implementations.

> * The height of the trigger should be `48px`. It's currently `49px` in its default state and `54px` when active. It should match the secondary default Button.

I'm seeing this as 48px in both Chrome and Safari. Are you able to point me at where this is happening and how I can reproduce this please? Safari 👇 

Screenshot 2023-03-14 at 17 50 32

JoshBowdenConcepts:
Is there a plan for ActionMenu's to live next to one another? If so I would imagine there would be a lot of jumping on the page. If this is the plan is there possibly some animation or action menu list refresh idea so make the change less jaring?

JoshBowdenConcepts (approved):
Looks really cool! I am excited to see how you handle multiple selections

rezrah:
> Is there a plan for ActionMenu's to live next to one another? If so I would imagine there would be a lot of jumping on the page. If this is the plan is there possibly some animation or action menu list refresh idea so make the change less jaring?

Yep! @JoshBowdenConcepts you can see an example of ActionMenu is horizontal stack configuration [in Storybook](https://primer-0a563ca881-26139705.drafts.github.io/storybook/?path=/story/components-actionmenu-features--in-stack).

You can also see production usage [in Galaxy today.](https://galaxy.github.com/#schedule)

packages/react/src/ActionMenu/ActionMenu.tsx
```diff
@@ -0,0 +1,387 @@
...
+        styles.ActionMenu__item,
+        type === 'single' && styles[`ActionMenu__item--selection-type-${type}`],
+        className
+      )}
+      // eslint-disable-next-line jsx-a11y/no-noninteractive-element-to-interactive-role
+      role={roleTypeMap[type || 'single']}
+      aria-checked={type === 'none' ? undefined : selected ? 'true' : 'false'}
+      aria-disabled={disabled ? 'true' : 'false'}
+      onClick={handler && !disabled ? () => handler(value) : undefined}
+      tabIndex={disabled ? undefined : 0}
```
TylerJDev, line 330: When a menu item is disabled, it loses the ability to be focused. Could we allow the `aria-disabled` items to still be focusable like the other menu items? When using a screen reader the amount of items in the menu will be announced, and this will include the `aria-disabled` items. If we allow the items to retain focus, a user knows what exactly is disabled.

TylerJDev (commented):
Left a few comments in regards to accessibility! I see that some of the comments I made are referencing recent suggested changes, so let me know what you think, or if you have questions regarding them!

danielguillan:
> I'm seeing this as 48px in both Chrome and Safari. Are you able to point me at where this is happening and how I can reproduce this please?

Screenshot showing the button node dimension with a height of 49 pixels

@rezrah this is what I was seeing yesterday, but I can't reproduce it anymore. So either it's been fixed somehow or it was a temporary bug om my browser. 🤷

packages/react/src/ActionMenu/ActionMenu.stories.tsx
```diff
@@ -0,0 +1,23 @@
...
+import {ActionMenu} from './ActionMenu'
+
+export default {
+  title: 'Components/ActionMenu',
+  component: ActionMenu
+} as ComponentMeta<typeof ActionMenu>
+
+const Template: ComponentStory<typeof ActionMenu> = args => (
+  <ActionMenu {...args} onSelect={newValue => alert(newValue)}>
+    <ActionMenu.Button aria-label="Actions">Open menu</ActionMenu.Button>
```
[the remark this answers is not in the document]
rezrah: 👍 Thanks Tyler, as we discussed in DMs, I've moved `aria-label` logic into the Overlay now.

packages/react/src/ActionMenu/ActionMenu.tsx
```diff
@@ -0,0 +1,387 @@
...
+      >
+        {SelectButton}
+        {showMenu ? SelectOverlay : null}
+      </div>
+    )
+  }
+)
+
+type ActionMenuButtonProps = PropsWithChildren<Ref<HTMLButtonElement>> & {
+  'aria-label': string
```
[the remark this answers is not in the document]
rezrah: Moved over to Overlay, but made it required for reasons we discussed in DMs. 

For posterity, these reasons were:

- aria-label is important for screen reader users. 
- there's a change it could be forgotten if we leave users to add this in
- this is therefore now required at the API-level to use this component

packages/react/src/ActionMenu/ActionMenu.tsx
```diff
@@ -0,0 +1,387 @@
...
+        } else if (child.type === ActionMenuOverlay) {
+          acc['Overlay'] = cloneElement(child as ReactElement<ActionMenuOverlayProps>, {
+            ref: floatingElementRef as React.RefObject<HTMLUListElement>,
+            className: clsx(styles.ActionMenu__menu, position && styles['ActionMenu__menu--visible']),
+            style: {
+              top: `${position?.top ?? 0}px`,
+              left: `${position?.left ?? 0}px`
+            },
+            id: `${instanceId}-menu`,
+            'aria-labelledby': ariaLabel,
```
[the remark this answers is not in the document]
rezrah: 👍 Thanks Tyler, as we discussed in DMs, I've moved `aria-label` logic into the Overlay now.

packages/react/src/ActionMenu/ActionMenu.tsx
rezrah: Good to know 👍 I've added back the ability to focus on this.

I initially set this to not focus on disabled items given that it's not interactive. Could this be a little misleading now that you can focus on it but not select the item though?

packages/react/src/ActionMenu/ActionMenu.tsx
```diff
@@ -0,0 +1,387 @@
...
+    const floatingElementRef = useRef<HTMLUListElement>(null)
+    const anchorElementRef = useRef<HTMLButtonElement>(null)
+    const instanceId = useId(id)
+
+    const ariaLabel = useMemo(() => {
+      const menuButton = Children.toArray(children).find(
+        child => isValidElement(child) && child.type === ActionMenuButton
+      ) as React.ReactElement<{'aria-label'?: string}> | undefined
+      return menuButton?.props['aria-label']
+    }, [children])
```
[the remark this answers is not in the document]
rezrah: That's correct. But based on the other feedback received, this functionality has been removed in favour of adding aria-label to overlay instead.

packages/react/src/ActionMenu/ActionMenu.tsx
TylerJDev: Good question! Here's a bit [more context on why they should be focusable](https://github.com/github/accessibility/blob/main/docs/office-hours-notes/2022-08-03-menu-disabled-items-segmented-control-tree-view.md#menus-with-disabled-items). Mainly we'd want to make them focusable because a screen reader will announce the amount of items in a menu including the disabled ones, so we want to make sure that they can reach all of them. 

Without the disabled item being focusable, a user would hear that a menu has 3 items, but can only find 2 of them as the third is disabled. We want to make sure each item is discoverable, even if they can't trigger them. The good thing is, the state of the item (disabled) will be announced to the user as we're using `aria-disabled` here.

TylerJDev (approved):
Looks good! Superb work :tada:!

packages/react/src/ActionMenu/ActionMenu.module.css
```diff
@@ -0,0 +1,139 @@
...
+    opacity: 0;
+    transform: translateY(-10px);
+  }
+  100% {
+    opacity: 1;
+    transform: translateY(0);
+  }
+}
+.ActionMenu__menu::-webkit-scrollbar {
+  width: 12px;
```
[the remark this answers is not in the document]
rezrah: 🙈 thanks @langermank

packages/react/src/ActionMenu/ActionMenu.tsx
rezrah: Great info, thanks Tyler 🙏. I guess the key here is aria-disabled being set. We'll need to make sure this has adequate coverage against regressions.

PR #209, opened 2023-03-13
Fix box-shadow transition bug

### LINKED ISSUE #206: 🐛 [BUG] - Button box shadow problem during OS alerts

TylerJDev:
## Summary

Fixes #206.

Adds `[disabled]` to `:hover` style to ensure it retains priority over `:focus` box-shadow rule. This also fixes the issue where the `:focus` box-shadow transitions into the `:hover` box-shadow, causing the effect in #206.

## List of notable changes:

- Adds `[disabled]` rule to `:hover` style.

- Ensure that `:focus` box-shadow doesn't overrule `:hover` box-shadow.

## Steps to test:

1. [ To Add ]

 
## Contributor checklist:

## Reviewer checklist:

## Screenshots:

<table>
<tr>
</tr>
<tr>
<td valign="top">

https://user-images.githubusercontent.com/26746305/224808953-f850cbf8-d73e-4f1b-b075-4f3ef89bc5e1.mov

Video of cursor clicking Brand Primary `<Button>` which shows the initial `:focus` box-shadow style being applied as the outline for the button. Clicking it triggers a form error which moves focus away from the button, causing the `:focus` box-shadow style to transition to the `:hover` box-shadow style.

 </td>
<td valign="top">

https://user-images.githubusercontent.com/26746305/224812759-116f45c1-7ce7-4d69-8a5e-a10aff1540f2.mov

Video of cursor clicking Brand Primary `<Button>` which shows the `:hover` box-shadow style initially. Clicking it triggers a form error which causes focus to move away from the button. No longer transitions from `:focus` style to `:hover` style, as it now retains the `:hover` styles. 
</td>
</tr>
</table>

### DIFF
--- a/packages/react/src/Button/Button.module.css
+++ b/packages/react/src/Button/Button.module.css
@@ -74,7 +74,7 @@
   background: var(--brand-Button-background-overlay);
 }
 
-.Button--primary:not(.Button--disabled):hover {
+.Button--primary:not(.Button--disabled[disabled]):hover {
   box-shadow: var(--brand-Button-shadow-primary-hover);
 }
 


packages/react/src/Button/Button.module.css
```diff
@@ -74,7 +74,7 @@
   background: var(--brand-Button-background-overlay);
 }
 
-.Button--primary:not(.Button--disabled):hover {
+.Button--primary:not(.Button--disabled[disabled]):hover {
```
rezrah, line 77: Quite odd that `.Button--disabled` isn't doing its job here. Do you know why?

Does this have the same effect? Wonder if it's worth just dropping the `.Button--disabled` class if attribute approach works just as well.

```jsx
.Button--primary:not(.Button[disabled]):hover {
```

Also, just an FYI, that in the example you shared the secondary button retains its focus on press. 


https://user-images.githubusercontent.com/13340707/225414637-e8270d07-1abe-41d8-b5fd-845e61cff4fa.mov



Can we make the effect consistent between all variants?

TylerJDev: Ah good callout, and thanks for the suggestion! I added the same rules for the secondary button variant. I also slimmed down the selector as recommended.

As for `.Button--disabled` not working as expected, I believe it's due to one selector being more specific than the other. It seems if I add `[disabled]` in one place, I need to make sure the other styles that'll apply in those scenarios also have the same `[disabled]` or `[aria-disabled]` selectors, as to keep the same priority in which styles overwrite others 🤔.

I also added `[disabled]` in some cases, because I only wanted those styles to apply when `[disabled]` and not `[aria-disabled]`.

TylerJDev: Added it to the `subtle` variant! 

I think the original button styles had the same focus/button press interactions, where the hover style took priority over the focus style, but I'm not entirely sure.

TylerJDev:
> Few things before you merge...
> 
> * Snapshots are out of date, I've triggered a CI update for you, but you may need to manually verify the snapshots after they're regenerated
> * There's a prettier formatting error thats failing in the Button file

Thank you! I'll get on this before I merge.

PR #215, opened 2023-03-17
Add landmark to `<Hero>` component

TylerJDev:
## Summary

This will resolve the [accessibility engineering review for the `<Hero>` component](https://github.com/github/primer/issues/903). A labelled `<section>`element can help users navigate the page, as the label will describe the section's contents. 

## List of notable changes:

- Swaps out `<div>` for `<section>`, ties landmark with heading
- Adds prop `headingLevel` to change heading level (defaults to `h1`)

## Steps to test:

1. Inspect `<section>` region in DevTools
2. Ensure that `<section>` is tied to heading through `aria-labelledby`

- [Region landmark example](https://www.w3.org/WAI/ARIA/apg/patterns/landmarks/examples/region.html)

## Contributor checklist:

## Reviewer checklist:


packages/react/src/Hero/Hero.tsx
```diff
@@ -15,17 +15,26 @@ type Action = {
 
 export type HeroProps = BaseProps<HTMLDivElement> & {
   heading: string | React.ReactElement
+  headingLevel?: 'h1' | 'h2' | 'h3'
```
rezrah, line 18: AFAIK the hero pattern should only ever use H1 and appear at the top of the page. We have other components like SectionIntro and CTABanner, which allow configurability for the other heading levels.

Is the configurability of the heading level super important right now? If not, wondering if we could defer this until the component rewrite has taken place. That way we can do this by simply forwarding the `as` in Heading.

packages/react/src/Hero/Hero.tsx
```diff
@@ -24,8 +24,13 @@ export type HeroProps = BaseProps<HTMLDivElement> & {
 export const Hero = forwardRef<HTMLDivElement, HeroProps>(
   ({className, heading, description, primaryAction, secondaryAction, align = 'start', ...rest}, ref) => {
     return (
-      <div className={clsx(styles.Hero, styles[`Hero--align-${align}`], className)} ref={ref} {...rest}>
-        <Heading className={styles['Hero-heading']} as="h1">
+      <section
+        aria-labelledby="hero-section-brand-heading"
```
rezrah, line 28: if someone wants to customise this, could we instead make this a prop value, with `hero-section-brand-heading` being the default value?

rezrah (approved):
Looks good, but also left some additional suggestions on improving customisation of the aria attribute value and making the change log entry more specific. Let me know if you want to talk through any of those points.

.changeset/gold-planes-call.md
```diff
@@ -0,0 +1,5 @@
+---
+'@primer/react-brand': patch
+---
+
+- Replaces `<div>` container with landmark element `<section>`
```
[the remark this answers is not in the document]
TylerJDev: Love this! Thanks a bunch for the suggestion, will be more descriptive in the change logs!

packages/react/src/Hero/Hero.tsx
TylerJDev: That's a good call. I added a new prop as suggested, with the default.

PR #220, opened 2023-03-24
Add label component

josepmartins:
## Summary

The `Label` component will be used to display status or add metadata to other elements. The component is non-interactive, can be displayed in different sizes, and feature lead visuals. 

## List of notable changes:

- **added** label component and color tokens
- **added** stories and tests
- **added** documentation
- **update** `Card` component and documentation to use `Label` and the `color` properties.

- Take a look at the CSS implementation, especially the part about the color gradient

## Steps to test:

- Storybook: https://primer-90dbc9b46e-26139705.drafts.github.io/storybook/?path=/story/components-label--default
- Documentation: https://primer-90dbc9b46e-26139705.drafts.github.io/components/Label

- Figma component: https://www.figma.com/file/BJ95AjraesmRCWsKA013GS/Primer-Brand?node-id=6084-28401
- Design issue: https://github.com/github/primer/issues/1921
- a11y design issue: https://github.com/github/primer/issues/1938
- API draft: https://www.figma.com/file/bcuH9lmlc0ZdY7QiZu6lvw/Primer-Brand---Label-API?node-id=0%3A1&t=wZ2sp2lHbch7vHFU-1

## Contributor checklist:

- [x] New visual snapshots have been generated/updated for any UI changes

## Reviewer checklist:


packages/react/src/Label/Label.module.css
```diff
@@ -0,0 +1,168 @@
...
+  height: var(--base-size-16);
+  width: auto;
+}
+
+.Label__leading-visual {
+  display: flex;
+  margin-right: var(--base-size-8);
+}
+
+.Label--label {
```
danielguillan, line 166: Nitpick: Not sure about this class name, as it is not used as a modifier for the `Label` block.

danielguillan (approved):
Looks great! I left some small non-blocking comments and suggestions.

packages/react/src/Label/Label.module.css
josepmartins: Good point. I'll rename it to `Label-label`

.changeset/moody-kings-hope.md
```diff
@@ -0,0 +1,25 @@
...
+
+Update Card component to use Label component and its color variants
+
+```jsx
+<Card href="https://github.com">
+  <Card.Label color="green">New feature</Card.Label>
+  <Card.Heading>Collaboration is the key to DevOps success</Card.Heading>
+  <Card.Description>Everything you need to know about getting started with GitHub Actions.</Card.Description>
+</Card>
+```
```
rezrah, line 25: Could we also add a separate changeset for `@primer/brand-primitives` please? 

E.g of how it's previously been done:

https://github.com/primer/brand/blob/main/packages/design-tokens/CHANGELOG.md#0150

So should contain a list a tokens specific to the Label

packages/react/src/Label/Label.features.stories.tsx
```diff
@@ -0,0 +1,67 @@
...
+    <Label color="red">Red</Label>
+    <Label color="red-orange">Red orange</Label>
+    <Label color="teal">Teal</Label>
+    <Label color="yellow">Yellow</Label>
+  </>
+)
+
+export const WithLeadingVisualSVG = () => (
+  <Label
+    leadingVisual={
```
rezrah, line 41: can we add a fill to the svg here? it's not visible in dark color mode
Screenshot 2023-04-06 at 09 30 22

packages/react/src/Label/Label.tsx
```diff
@@ -0,0 +1,104 @@
...
+  'lime',
+  'orange',
+  'pink',
+  'pink-blue',
+  'purple',
+  'purple-red',
+  'red',
+  'red-orange',
+  'teal',
+  'yellow'
```
rezrah, line 27: If none of these values were suitable for users and someone needed an escape-hatch, what's the recommended approach to customising the color?

I see we have tokens `--brand-Label-color-start` and `--brand-Label-color-end`, that get reassigned based on the color value mapping in css, but to avoid app devs overriding the other values here, or introducing global overrides, should we consider introducing a `custom` value here, which maps to something like `--brand-Label-color-custom`?

packages/react/src/Label/Label.tsx
```diff
@@ -0,0 +1,104 @@
...
+export const defaultLabelColor = LabelColors[0]
+export const defaultLabelSize = LabelSizes[1]
+
+export type LabelProps = BaseProps<HTMLSpanElement> & {
+  /**
+   * The leading visual appears before the Label content
+   */
+  leadingVisual?: React.ReactNode
+  /**
+   * The styling variations available in Label
```
rezrah, line 40:
```suggestion
   * The color variations available in Label
```

packages/react/src/Label/Label.tsx
```diff
@@ -0,0 +1,104 @@
...
+          >
+            {children}
+          </Text>
+        </span>
+      </span>
+    )
+  }
+)
+
+export const Label = Object.assign(_Label, {testIds})
```
rezrah, line 104: QoL improvement for VisualCode users. This will provide context when they hover over the component name:

```suggestion
/**
 * Use a Label to indicate the status of an object or to categorize it with metadata that describes it.
 * @see https://primer.style/brand/components/Label
 */
export const Label = Object.assign(_Label, {testIds})
````

File: packages/react/src/Label/Label.tsx [hunk 7]
```diff
@@ -0,0 +1,104 @@
...
+                ['aria-hidden']: 'true'
+              })
+            )}
+          </span>
+        )}
+        <span className={styles['Label__text']}>
+          <Text
+            as="span"
+            size={size === 'small' ? '100' : size === 'medium' ? '200' : '300'}
+            className={clsx(styles['Label-label'])}
```
rezrah, line 94: For consistency with BEM pattern used elsewhere in this component

```suggestion
            className={clsx(styles['Label__label'])}
```

packages/react/src/Label/Label.tsx
```diff
@@ -0,0 +1,104 @@
+import clsx from 'clsx'
+import React, {forwardRef, type Ref} from 'react'
+import {Text} from '../Text'
+import '@primer/brand-primitives/lib/design-tokens/css/tokens/functional/components/label/colors-with-modes.css'
```
rezrah, line 4: Nit: Could we move stylesheets below JS imports and put annotations above them please?

Example: https://github.com/primer/brand/blob/7da349d21d7cc4bf77b105186a9a1aae14ae464b/packages/react/src/ActionMenu/ActionMenu.tsx#L27

apps/docs/content/components/Label.mdx
```diff
@@ -0,0 +1,78 @@
...
+  </Label>
+  <Label color="blue-purple" leadingVisual={CopilotIcon}>
+    Copilot
+  </Label>
+</Stack>
+```
+
+## Component props
+
+### Button
```
rezrah, line 72:
```suggestion
### Label
```

apps/docs/content/components/Label.mdx
```diff
@@ -0,0 +1,78 @@
+---
+title: Label
+status: Experimental
+figma: 'https://www.figma.com/file/BJ95AjraesmRCWsKA013GS/Primer-Brand?node-id=6084-28401'
```
rezrah, line 4: This goes to Button page in Figma. Can we repoint to Label?

apps/docs/content/components/Label.mdx
```diff
@@ -0,0 +1,78 @@
...
+source: https://github.com/primer/brand/blob/main/packages/react/src/Label/Label.tsx
+storybook: '/brand/storybook/?path=/story/components-label--default'
+description: Use a Label to indicate the status of an object or to categorize it with metadata that describes it.
```
rezrah, line 7: Could we use a similar description to PCSS? 

```suggestion
Use Labels to add metadata or indicate status of items and navigational elements.
```

packages/react/src/Label/Label.features.stories.tsx
josepmartins: Fixed 👍

apps/docs/content/components/Label.mdx
josepmartins: I think the usage in Brand is a bit different as we don't put labels for navigational items, but headings/titles mainly.

What about something in between and simpler?

```suggestion
description: Use Labels to add metadata or indicate an object's status.
```

apps/docs/content/components/Label.mdx
josepmartins: Do you have the Figma branch updated? I just double-checked that and it goes to the Label page.

apps/docs/content/components/Label.mdx
josepmartins: oopsie!

packages/react/src/Label/Label.tsx
rezrah: Okay, so just to be clear.. if i want to have one new color variation (say pink-orange) of a Label in an app, purely as a one-off.. am i still going to be able to use all the other built-in color values as normal? or will that custom color override prevent the originals from working correctly?

apps/docs/content/components/Label.mdx
rezrah: Huh, yeah it goes there when i open in browser, but in app it loaded button. 🤷 .. guess it's fine and maybe my app is buggy.

packages/react/src/Label/Label.tsx
josepmartins: You could define that one-off color by either overriding the  `--brand-Label-color-start` and `--brand-Label-color-end` in selecting the specific Label class or creating a custom one with those changes and adding it to the Label,  and that won't affect the rest of the Label colors on a page

Example modifying the first Label colors in the playground to pink-orange:

Screenshot 2023-04-13 at 12 08 13

rezrah: > and that won't affect the rest of the Label colors on a page

👍 works for me. thanks for clarifying @josepmartins

packages/react/src/Label/Label.module.css
```diff
@@ -0,0 +1,168 @@
...
+  align-items: center;
+  justify-content: center;
+  border-radius: var(--brand-borderRadius-full);
+  display: inline-flex;
+  color: var(--brand-Label-color-start, var(--brand-Label-color, currentColor));
+  position: relative;
+  background: none;
+}
+
+.Label:before {
```
rezrah, line 19: Super minor nit

```suggestion
.Label::before {
```

rezrah (approved):
🚀 

I think there's still some aliasing issues in dark mode, but I'm also seeing the same issue in PCSS and PRC. Think it's worth looking at separately as a follow up maybe?

Also FYI that your visual snapshots will need to be updated again it seems.

josepmartins:
> I think there's still some aliasing issues in dark mode, but I'm also seeing the same issue in PCSS and PRC. Think it's worth looking at separately as a follow up maybe?

I saw those issues in PCSS and PRC as well. I'll create an issue for that 👍

PR #225, opened 2023-04-04
Add `combobox` support to `SubdomainNavBar`

TylerJDev:
## Summary

Adds keyboard support for the search region in `SubdomainNavBar`. This PR aims to implement the [`combobox` pattern](https://www.w3.org/WAI/ARIA/apg/patterns/combobox/).

Resolves `combobox` issue from https://github.com/github/primer/issues/1668#issuecomment-1490677547. 

## List of notable changes:

- Adds `combobox` support
- Adds live region to announce amount of results to users of assistive technologies (AT)

- Keyboard usage within `search` region
- Added live region

## Steps to test:

1. Go to [`SubdomainNavBar` "Search results visible" example in Storybook](https://primer-6f41389c5d-26139705.drafts.github.io/storybook/?path=/story/components-subdomainnavbar--search-results-visible)
2. Focus input and using `Up` & `Down` arrow keys to navigate through the options.
3. Test other key support (`Enter` to select option, `Escape` to close listbox)

- [`Combobox` pattern](https://www.w3.org/WAI/ARIA/apg/patterns/combobox/) (non-normative)
- [`auto-complete-element`](https://github.com/github/auto-complete-element) (used as reference)
- [`role='combobox'` WAI-ARIA 1.2](https://www.w3.org/TR/wai-aria-1.2/#combobox)

## Contributor checklist:

## Reviewer checklist:


TylerJDev:
Currently working on unblocking the current test failure, but for the most part the work is at a state where it can be reviewed.

@rezrah, @JoshBowdenConcepts - If you have capacity, I'd love a review for this PR. This one is a bit more bulky than the rest, so I want to make sure I'm following the best practices within Brand. 

None of the changes here should consist of any style change, the only change is the functionality of the search region (i.e. additional keyboard navigation).

packages/react/src/SubdomainNavBar/SubdomainNavBar.module.css
```diff
@@ -593,6 +593,10 @@
     var(--brand-SubdomainNavBar-search-results-border-default, #b7bfc7);
 }
 
+.SubdomainNavBar-search-result-item[aria-selected='true'] a {
+  outline: var(--brand-color-focus) auto 1px;
```
rezrah, line 597:
```suggestion
  outline: var(--brand-color-focus) auto var(--brand-borderWidth-thick);
```

packages/react/src/SubdomainNavBar/SubdomainNavBar.tsx
```diff
@@ -369,6 +450,15 @@ const _SearchInternal = (
                 </ul>
               </div>
             )}
+            <div
+              aria-live="polite"
+              aria-atomic="true"
+              data-testid="search-live-region"
+              className={styles['SubdomainNavBar-search-result-item-suggestion']}
```
rezrah, line 457: Nice! As this seems to be visually hidden, can we use that global util class here instead?

```suggestion
              className="visually-hidden"
```

packages/react/src/SubdomainNavBar/SubdomainNavBar.module.css
```diff
@@ -609,6 +613,18 @@
...
+  width: 1px;
+  height: 1px;
+  padding: 0;
+  margin: -1px;
+  overflow: hidden;
+  clip: rect(0, 0, 0, 0);
+  white-space: nowrap;
+  border-width: 0;
+}
+
```
rezrah, line 627: Think this can be removed as we have a `visually-hidden` global style ([link](https://github.com/primer/brand/blob/ce969048bd332a00f3f8d70deba15b44edcc4fd8/packages/react/src/css/utilities.css#L6))

packages/react/src/SubdomainNavBar/SubdomainNavBar.tsx
```diff
@@ -266,7 +267,75 @@ const _SearchInternal = (
...
+    const dialog = dialogRef.current
+    let count
+
+    // Prevent any other keys outside of supported from being prevented.
+    // Only prevent "Enter" if activeDescendant is greater than -1.
+    if (!supportedKeys.includes(event.key) || (event.key === 'Enter' && activeDescendant === -1) || !dialog) {
+      return false
+    }
+
+    event.preventDefault()
```
rezrah, line 298: Could this be hoisted to the top of the handler, or are you partially relying on some default behavior?

packages/react/src/SubdomainNavBar/SubdomainNavBar.tsx
```diff
@@ -266,7 +267,75 @@ const _SearchInternal = (
...
+      dialog.querySelector(`#subdomainnavbar-search-result-${count}`)?.scrollIntoView()
+    }
+
+    if (event.key === 'Enter') {
+      const link = dialog.querySelector(`#subdomainnavbar-search-result-${activeDescendant} a`) as HTMLAnchorElement
+      link.click()
+    }
+  }
+
+  const searchLiveRegion = () => {
```
rezrah, line 320: can we memoize this in a `useCallback` please? Minor perf optimisation.

packages/react/src/SubdomainNavBar/SubdomainNavBar.tsx
```diff
@@ -266,7 +267,75 @@ const _SearchInternal = (
...
+    setActiveDescendant(-1)
+  }
+
+  useOnClickOutside(dialogRef, handleClose)
+  useKeyboardEscape(() => {
+    setListboxActive(false)
+    setActiveDescendant(-1)
+  })
+
+  const handleAriaFocus = event => {
```
rezrah, line 285: Can this go in a useCallback also please?

packages/react/src/SubdomainNavBar/SubdomainNavBar.tsx
```diff
@@ -369,6 +450,15 @@ const _SearchInternal = (
                 </ul>
               </div>
             )}
+            <div
+              aria-live="polite"
+              aria-atomic="true"
```
rezrah, line 455: Just for my own understanding, is `true` preferred to `false` for any particular reason? If search results were particularly large for example, and a user is trying to optimise / refine a search query, would it not be better to read out the changes rather than the entire result again? The top result for example, may not be any different.

packages/react/src/SubdomainNavBar/SubdomainNavBar.tsx
```diff
@@ -369,6 +450,15 @@ const _SearchInternal = (
...
+              data-testid="search-live-region"
+              className={styles['SubdomainNavBar-search-result-item-suggestion']}
+            >
+              {`${searchResults?.length} suggestions.`}
```
rezrah, line 459: I like this feature, and wonder if we can make this work for sighted users also..

@danielguillan thoughts on putting the results count in the results heading?

Before:
Screenshot 2023-04-11 at 12 08 20

After:
Screenshot 2023-04-11 at 12 11 19

@TylerJDev - if we did this ☝️ - would that make this visually hidden code redundant?

rezrah (approved):
Looks and works great @TylerJDev - thank you.

Left some minor feedback for you, let me know if you'd like another review later.

packages/react/src/SubdomainNavBar/SubdomainNavBar.tsx
TylerJDev: This is perfect, thank you for the suggestion!

packages/react/src/SubdomainNavBar/SubdomainNavBar.tsx
TylerJDev: I put it below so that it wouldn't restrict regular key presses outside of arrow keys, and the enter key in some scenarios. Mainly so users can still type without it being prevented.

packages/react/src/SubdomainNavBar/SubdomainNavBar.tsx
TylerJDev: Good question! The intention was so that the user would be notified of changes even if their query amounted in the same number of results. `aria-atomic` will ensure that it announces on each change regardless of if the value inside the live region is the same as any prior announcements. (e.g. user types "devops" AT announces "**8 suggestions**", user types "devo", amount of results haven't changed so AT announces "**8 suggestions**" again), without the `aria-atomic="true"`, the announcements would most likely cut out after the initial "8 suggestions" announcement.


There definitely is some redundancy here, but the `aria-live="polite"` should make it not interrupt anything else the user might be doing.

packages/react/src/SubdomainNavBar/SubdomainNavBar.tsx
TylerJDev: Yup! We could definitely remove the current visually hidden live region and utilize this instead if we went this route. This might even be a more accessible way of going about it too!

rezrah: cc. @josepmartins @danielguillan .. would you both be okay to move forward with this?

rezrah (approved):
Looks great 👍

PR #227, opened 2023-04-06
Encapsulate focus within `dialog`

TylerJDev:
## Summary

Adds a focus-trap to the search `dialog`. 

Resolves `dialog` issue from https://github.com/github/primer/issues/1668 (internal).

## List of notable changes:

- Pulls in Primer `useFocusTrap` hook to encapsulate user focus within `dialog`.

- Ensure that focus does not leave dialog unless dismissed manually (currently through the close button)

## Steps to test:

1. Open up Storybook example of `SubdomainNavbar` with search feature present.
2. See if focus is trapped within the `dialog` after activation of the search button.

- [`useFocusTrap`](https://github.com/primer/react/blob/main/src/hooks/useFocusTrap.ts)

## Contributor checklist:

## Reviewer checklist:


TylerJDev:
@rezrah - I pulled in our focusTrap hook from PRC as a new hook entirely. Let me know if this isn't the right way to go about it!

.changeset/thirty-snails-sell.md
```diff
@@ -0,0 +1,6 @@
+---
+'@primer/brand-storybook': patch
```
rezrah, line 2: This can be removed, as we don't publish this package ☝️

packages/react/src/hooks/useProvidedRefOrCreate.ts
```diff
@@ -0,0 +1,14 @@
...
+
+/**
+ * There are some situations where we only want to create a new ref if one is not provided to a component
+ * or hook as a prop. However, due to the `rules-of-hooks`, we cannot conditionally make a call to `React.useRef`
+ * only in the situations where the ref is not provided as a prop.
+ * This hook aims to encapsulate that logic, so the consumer doesn't need to be concerned with violating `rules-of-hooks`.
+ * @param providedRef The ref to use - if undefined, will use the ref from a call to React.useRef
+ * @type TRef The type of the RefObject which should be created.
+ */
+export function useProvidedRefOrCreate<TRef>(providedRef?: React.RefObject<TRef>): React.RefObject<TRef> {
```
rezrah, line 11: I debated pulling this in for ActionMenu too, but decided against it because in most cases you'll almost always need to provide the ref upfront, so there's no need to create one. Looking at where you're using it, perhaps that's worth considering here too? The less custom code we need to maintain the better IMO.

packages/react/src/hooks/useFocusTrap.ts
```diff
@@ -0,0 +1,80 @@
...
+   */
+  restoreFocusOnCleanUp?: boolean
+}
+
+/**
+ * Hook used to trap focus inside a container. Returns a ref that can be added to the container
+ * that should trap focus.
+ * @param settings {FocusTrapHookSettings}
+ */
+export function useFocusTrap(
```
rezrah, line 36: Are there any tests for this hook that we could also copy over here?

packages/react/src/hooks/useFocusTrap.ts
```diff
@@ -0,0 +1,80 @@
...
+  /**
+   * Set to true to disable the focus trap and clean up listeners. Can be re-enabled at
+   * any time.
+   */
+  disabled?: boolean
+
+  /**
+   * If true, when this focus trap is cleaned up, restore focus to the element that had
+   * focus immediately before the focus trap was enabled. (Default: false)
+   */
```
rezrah, line 27:
```suggestion
   * If true, when this focus trap is cleaned up, restore focus to the element that had focus immediately before the focus trap was enabled. (Default: false)
   */
```

rezrah (approved):

packages/react/src/hooks/useProvidedRefOrCreate.ts
TylerJDev: Yup, that makes sense. I removed it, doesn't seem like we need it at all.

packages/react/src/hooks/useFocusTrap.ts
TylerJDev: Doesn't seem like there is. Only a story which is dependent on some PRC components. I can definitely follow up with a new test for this hook after this PR if that's okay?

rezrah: Yeah that'd be great, thanks. No rush.

PR #229, opened 2023-04-11
Image Component

JoshBowdenConcepts:
## Summary

Creating an image component!

## List of notable changes:

- **Added** Image Component

- Making sure all test cases are thought through
- Making sure the rendered output is as expected

## Steps to test:

1. Open the Image component in CI-deployed preview environment
2. Go to Image story in Storybook
3. Verify that it behaves as described.

- [API Exploration](https://www.figma.com/file/pYV9ln5Bkvh07lbzczaTkF/Primer-Brand---Image-Component-Exploration?node-id=0-1&t=CoiGn3tOAfI0QP9D-0)
- [Tracking Issue](https://github.com/github/primer/issues/2048)

## Contributor checklist:

## Reviewer checklist:


.changeset/swift-clouds-tell.md
```diff
@@ -0,0 +1,14 @@
+---
+'@primer/brand-storybook': patch
+'@primer/react-brand': patch
```
rezrah, line 3: Can we make this a minor please, as it's a new component.

```suggestion
'@primer/react-brand': minor
```

apps/docs/content/components/Image.mdx
```diff
@@ -0,0 +1,90 @@
+---
+title: Image
+status: Experimental
+# figma: https://www.figma.com/file/BJ95AjraesmRCWsKA013GS/Primer-Brand?node-id=539%3A14285&t=RJEWZhVUtrRQ0pKh-0
```
rezrah, line 4: This can be removed altogether if it's not applicable

```suggestion
```

apps/docs/content/components/Image.mdx
```diff
@@ -0,0 +1,90 @@
...
+## Component props
+
+### Image <Label>Required</Label>
+
+The first row in a table is automatically styled as a table header row.
+
+| name          | type                                    | default     | required | description                                                                                                                                                               |
+| ------------- | --------------------------------------- | ----------- | -------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| `src`         | `string`                                |             | `true`   | Specifies the path to the image                                                                                                                                           |
+| `alt`         | `string`                                |             | `true`   | Specifies an alternate text for the image                                                                                                                                 |
```
rezrah, line 83:
```suggestion
| `alt`         | `string`                                |             | `true`   | Specifies a text value explaining the nature of the image for users of Assistive Technolog                                                                                                                         y
```

apps/docs/content/components/Image.mdx
```diff
@@ -0,0 +1,90 @@
...
+  width={200}
+  aspectRatio="16:9"
+/>
+```
+
+## Component props
+
+### Image <Label>Required</Label>
+
+The first row in a table is automatically styled as a table header row.
```
rezrah, line 78: ☝️ Is this intentional? Looks like it's from [ComparisonTable](https://primer-9770a149af-26139705.drafts.github.io/components/ComparisonTable#comparisontablerow-required). If so, can we remove this please?

```suggestion
```

apps/docs/content/components/Image.mdx
```diff
@@ -0,0 +1,90 @@
...
+
+### Default
+
+```jsx live
+<Image src="https://source.unsplash.com/random" alt="Random Image" />
+```
+
+### Picture
+
+The `as` prop can be used to set the container of the image to be a picture component.
```
rezrah, line 26:
```suggestion
The `as` prop can be used to set the container of the image to be use `picture`.
```

apps/docs/content/components/Image.mdx
```diff
@@ -0,0 +1,90 @@
...
+
+```jsx live
+<Image
+  src="https://source.unsplash.com/random"
+  alt="Random Image"
+  as="picture"
```
rezrah, line 32: could we hoist as="picture" to the top as it's most relevant to this example?

apps/docs/content/components/Image.mdx
```diff
@@ -0,0 +1,90 @@
...
+
+### Height
+
+The `height` prop can be used to set the height of the image. This can be used along side the `aspectRatio` prop to create a responsive image the same size as other images.
+
+```jsx live
+<Image
+  src="https://source.unsplash.com/random"
+  alt="Random Image"
+  height={200}
```
rezrah, line 56: Not sure if this example is working 🤔 ... changing the height value isn't updating the live code example, it's always rendering full-height.

apps/docs/content/components/Image.mdx
```diff
@@ -0,0 +1,90 @@
...
+### Width
+
+The `width` prop can be used to set the width of the image. This can be used along side the `aspectRatio` prop to create a responsive image the same size as other images.
+
+```jsx live
+<Image
+  src="https://source.unsplash.com/random"
+  alt="Random Image"
+  width={200}
+  aspectRatio="16:9"
```
rezrah, line 70: The aspect ratio here doesn't seem to be 16:9, is it working correctly? It looks more square than rectangular.

Screenshot 2023-04-14 at 11 13 10

apps/docs/content/components/Image.mdx
```diff
@@ -0,0 +1,90 @@
...
+| `src`         | `string`                                |             | `true`   | Specifies the path to the image                                                                                                                                           |
+| `alt`         | `string`                                |             | `true`   | Specifies an alternate text for the image                                                                                                                                 |
+| `as`          | `img`, `picture`                        | `img`       | `false`  | Specification to create a picture component                                                                                                                               |
+| `aspectRatio` | `1:1`, `16:9`, `16:10`, `4:3`, `custom` | `undefined` | `false`  | Sets the image aspect ratio. A custom ratio can be provided in the design tokens.                                                                                         |
```
rezrah, line 85: Is `custom` still a valid value, or were we dropping it. If it's valid, can we add an example of how users should use it please?

rezrah: Could we add quotes around these to convey these are strings?

```suggestion
| `aspectRatio` | `'1:1'`, `'16:9'`, `'16:10'`, `'4:3'`, `'custom'` | `undefined` | `false`  | Sets the image aspect ratio. A custom ratio can be provided in the design tokens.                                                                                         |

````

File: packages/design-tokens/src/tokens/functional/components/image/base.json [hunk 10]
```diff
@@ -0,0 +1,11 @@
+{
+  "brand": {
+    "Image": {
+      "ratio": {
+        "custom": {
+          "value": "3 / 2"
```
rezrah, line 6: Could this be `none` to avoid being opinionated about this by default?

packages/react/src/Image/Image.tsx
```diff
@@ -0,0 +1,64 @@
+import React from 'react'
+import clsx from 'clsx'
+import styles from './Image.module.css'
+
+type AspectRatio = '1:1' | '16:9' | '16:10' | '4:3' | 'custom'
+
+export interface ImageProps extends React.ImgHTMLAttributes<HTMLImageElement> {
+  ref?: React.Ref<HTMLImageElement>
```
rezrah, line 8: Could you add `src` to this list please? I think that's supposed to be required, but isn't here. 

Example from Storybook:
Screenshot 2023-04-14 at 11 22 28

rezrah, line 8: Could we please use BaseProps here with a forwardRef? Should be examples in other components.

packages/react/src/Image/Image.tsx
```diff
@@ -0,0 +1,64 @@
+import React from 'react'
+import clsx from 'clsx'
+import styles from './Image.module.css'
+
+type AspectRatio = '1:1' | '16:9' | '16:10' | '4:3' | 'custom'
+
+export interface ImageProps extends React.ImgHTMLAttributes<HTMLImageElement> {
```
rezrah, line 7: Could we change this to a `type` for consistency. We don't typically use interfaces for component prop definitions and type is easier for intersections.

packages/react/src/Image/Image.tsx
```diff
@@ -0,0 +1,64 @@
...
+    const [width, height] = ratio.split(':').map(Number)
+    return `${width}-${height}`
+  }
+}
+
+export const Image = ({className, aspectRatio, as = 'img', ref, alt, width, height, ...rest}: ImageProps) => {
+  if (as === 'picture') {
+    return (
+      <picture
+        className={clsx(styles['Image-container'], aspectRatio && styles[`Image--${aspectRatioResolver(aspectRatio)}`])}
```
rezrah, line 27: Could we make the modifier describe the aspect ratio context here please?

e.g. `Image--aspect-ratio-16-9`?

packages/react/src/Image/Image.tsx
```diff
@@ -0,0 +1,64 @@
...
+
+export interface ImageProps extends React.ImgHTMLAttributes<HTMLImageElement> {
+  ref?: React.Ref<HTMLImageElement>
+  alt: string
+  as?: 'picture' | 'img'
+  aspectRatio?: AspectRatio
+}
+
+const aspectRatioResolver = (ratio?: AspectRatio) => {
+  if (ratio === undefined) return ''
```
rezrah, line 15: Is this part necessary? As you're checking for truthy occurrence of `aspectRatio`,  so this may be redundant.

packages/react/src/Image/Image.tsx
```diff
@@ -0,0 +1,64 @@
...
+          height={width ? width : '100%'}
+          className={clsx(styles.Image, className)}
+          {...rest}
+        />
+      </picture>
+    )
+  }
+  if (aspectRatio) {
+    return (
+      <span className={clsx(styles['Image-container'], styles[`Image--${aspectRatioResolver(aspectRatio)}`])}>
```
rezrah, line 42:
```suggestion
      <span className={clsx(styles['Image__container'], styles[`Image--${aspectRatioResolver(aspectRatio)}`])}>
```

packages/react/src/Image/Image.stories.tsx
```diff
@@ -0,0 +1,79 @@
...
+    }
+  }
+} as ComponentMeta<typeof Image>
+
+export const Playground: ComponentStory<typeof Image> = args => <Image {...args} />
+
+export const Default = Playground.bind({
+  args: {
+    aspectRatio: undefined
+  }
```
rezrah, line 78: Would this be undefined by default? Or is it getting overridden somewhere?

packages/react/src/Image/Image.test.tsx
```diff
@@ -0,0 +1,141 @@
...
+
+import {Image} from './'
+
+expect.extend(toHaveNoViolations)
+
+describe('Image', () => {
+  afterEach(cleanup)
+
+  it('no a11y violations', async () => {
+    const {container} = render(<Image src="https://source.unsplash.com/random" alt="" aspectRatio={undefined} />)
```
rezrah, line 14: Noticing `aspectRatio={undefined}` on quite a few of these tests, could we remove them as it's not a required prop.

also instead of `""` for the alt text, could we put something like 'mock alt text' please?

```suggestion
    const {container} = render(<Image src="https://source.unsplash.com/random" alt="mock alt text" />)
```

packages/react/src/Image/Image.test.tsx
```diff
@@ -0,0 +1,141 @@
...
+      <Image src="https://source.unsplash.com/random" alt="" className={testClass} aspectRatio={undefined} />
+    )
+
+    expect(container.querySelector(`img.${testClass}`)).toBeInTheDocument()
+  })
+
+  it('should return only an img element if no aspectRatio is provided and as is not picture', async () => {
+    const {container} = render(<Image src="https://source.unsplash.com/random" alt="" aspectRatio={undefined} />)
+
+    expect(container.querySelector('.Image-container')).not.toBeInTheDocument()
```
rezrah, line 33: Rather than querySelector, could we use either `getByAltText`, which serves dual advantage of checking for alt text, or use `getByRole('img')`?

packages/react/src/Image/Image.test.tsx
```diff
@@ -0,0 +1,141 @@
...
+    const testClass = 'test'
+
+    const {container} = render(
+      <Image src="https://source.unsplash.com/random" alt="" className={testClass} aspectRatio={undefined} />
+    )
+
+    expect(container.querySelector(`img.${testClass}`)).toBeInTheDocument()
+  })
+
+  it('should return only an img element if no aspectRatio is provided and as is not picture', async () => {
```
rezrah, line 30:
```suggestion
  it('should return only an img element if no aspectRatio is provided and is not a picture', async () => {
```

packages/react/src/Image/Image.test.tsx
```diff
@@ -0,0 +1,141 @@
...
+    const {container} = render(<Image src="https://source.unsplash.com/random" alt="" aspectRatio={undefined} />)
+
+    expect(container.querySelector('.Image-container')).not.toBeInTheDocument()
+    expect(container.querySelector('span')).not.toBeInTheDocument()
+    expect(container.querySelector('picture')).not.toBeInTheDocument()
+    expect(container.querySelector('.Image')).toBeInTheDocument()
+    expect(container.querySelector('img')).toBeInTheDocument()
+  })
+
+  it('should return an img element in a span tag if aspectRatio is provided and as is equal to picture', async () => {
```
rezrah, line 40:
```suggestion
  it('should return an img element in a span tag if aspectRatio is provided and is not set to picture', async () => {
```

packages/react/src/Image/Image.test.tsx
```diff
@@ -0,0 +1,141 @@
...
+    const {container} = render(<Image src="https://source.unsplash.com/random" alt="" aspectRatio={'1:1'} />)
+
+    expect(container.querySelector('.Image-container')).toBeInTheDocument()
+    expect(container.querySelector('span')).toBeInTheDocument()
+    expect(container.querySelector('picture')).not.toBeInTheDocument()
+    expect(container.querySelector('.Image')).toBeInTheDocument()
+    expect(container.querySelector('img')).toBeInTheDocument()
+  })
+
+  it('should return an img element in a picture tag if aspectRatio is provided and as is equal to picture is true', async () => {
```
rezrah, line 50:
```suggestion
  it('should return an img element in a picture tag if aspectRatio is provided and picture is set', async () => {
```

packages/react/src/Image/Image.tsx
```diff
@@ -0,0 +1,64 @@
...
+        />
+      </span>
+    )
+  }
+  return (
+    <img
+      ref={ref}
+      alt={alt}
+      className={clsx(styles.Image, className)}
+      width={width ? width : 'initial'}
```
rezrah, line 59: Out of curiousity why do we need to set `initial` in normal usage of this component? If it were to behave like a normal `img` out-of-box, could we make this undefined instead?

Screenshot 2023-04-14 at 12 43 58

packages/react/src/Image/ImageStory.module.css
```diff
@@ -0,0 +1,5 @@
+.custom-image {
```
rezrah, line 1: Could we rename this file to `Image.stories.module.css` please? We do this in `River`, and could be useful later if we need to pattern match any story specific css files.

packages/react/src/Image/Image.tsx
JoshBowdenConcepts: BaseProps gives issues related to ref className and id because they are duplicated. I would need to Omit them from the orignal class. I think instead we should remove BaseProps or use it for something non standard in general HTML interfaces.

apps/docs/content/components/Image.mdx
```diff
@@ -0,0 +1,89 @@
...
+```jsx live
+<Image src="https://source.unsplash.com/random" alt="Random Image" />
+```
+
+### Picture
+
+The `as` prop can be used to set the container of the image to be use `picture`.
+
+```jsx live
+<Image
```
rezrah, line 28: If I wanted to generate this markup:

```html
<picture>
  <source srcset="logo-768.png, logo-768-1.5x.png 1.5x" />
  <img src="logo-320.png" alt="logo" />
</picture>
```

How would I do that with as="picture"? I can't see a way to set `sources`.

apps/docs/content/components/Image.mdx
```diff
@@ -0,0 +1,89 @@
...
+| `alt`         | `string`                                |             | `true`   | Specifies a text value explaining the nature of the image for users of Assistive Technolog                                                                                                                         y
+| `as`          | `img`, `picture`                        | `img`       | `false`  | Specification to create a picture component                                                                                                                               |
+| `aspectRatio` | `'1:1'`, `'16:9'`, `'16:10'`, `'4:3'`, `'custom'` | `undefined` | `false`  | Sets the image aspect ratio. A custom ratio can be provided in the design tokens.                                                                                         |
+
+| `height`      | `number`                                |             | `false`  | The height of the image element or its container if it has an aspect ratio                                                                                                |
```
rezrah, line 85: FYI The table seems to be breaking around here somewhere...

Screenshot 2023-04-17 at 09 31 50

packages/design-tokens/src/tokens/functional/components/image/base.json
JoshBowdenConcepts: I believe that auto is the default so I will set it to that. If that works for you.

apps/docs/content/components/Image.mdx
```diff
@@ -0,0 +1,90 @@
...
+/>
+```
+
+### Width
+
+The `width` prop can be used to set the width of the image. This can be used along side the `aspectRatio` prop to create a responsive image the same size as other images.
+
+```jsx live
+<Image
+  src="https://source.unsplash.com/random"
```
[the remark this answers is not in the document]
JoshBowdenConcepts: I think we discussed this before but I would like to use a different color for this image because I can't see this on my monitor. So I will pick the next darker grey available.

packages/react/src/Image/Image.tsx
JoshBowdenConcepts: I don't believe so I will remove it.

apps/docs/content/components/Image.mdx
JoshBowdenConcepts: Currently it sets on the image tag but I could include the children prop. Do you think that would be beneficial? I wasn't sure if that was really more performant.

apps/docs/content/components/Image.mdx
JoshBowdenConcepts: this should be resolved in this build

apps/docs/content/components/Image.mdx
rezrah: We'd deffo need a way to produce the `<source>` elements when setting `as="picture"`. I don't think we need to do it via JSX children, but it should render markup like the example above. Could you show me me how you saw it working currently without doing that?

apps/docs/content/components/Image.mdx
rezrah: I'm still seeing it unfortunately and can't see any upcoming new deployments.. when you say 'this build', has it been pushed up?

packages/react/src/Image/Image.tsx
rezrah: Reopening this as it still seems to be happening.. can those be removed @JoshBowdenConcepts?
Screenshot 2023-04-17 at 16 51 33

apps/docs/content/components/Image.mdx
JoshBowdenConcepts: Basically you would apply the property to the image. What I can do it parse the property from the img component and then take each value to create a source component. Would that work for you?

packages/react/src/Image/Image.tsx
JoshBowdenConcepts: Whoops! I didn't think I closed it but it should be updated now!

packages/react/src/Image/Image.tsx
rezrah: > they are duplicated and extending different types

BaseProps accept a generic type, where you can pass in the HTMLImageElement. As it's an intersection, it should combine any duplicates too. Using BaseProps is preferred because it's also future proofing it, where we can easily deploy global props in one-place, which we're thinking about doing for spacing.

apps/docs/content/components/Image.mdx
```diff
@@ -0,0 +1,94 @@
...
+```jsx live
+<Image
+  src="https://via.placeholder.com/600x400/d3d9df/d3d9df.png"
+  alt="placeholder, blank area with an off-white background color"
+/>
+```
+
+### Picture
+
+The `as` prop can be used to set the container of the image to be use `picture`.
```
rezrah, line 30: Typo

```suggestion
The `as` prop can be used to set the container of the image to use `picture`.
```

apps/docs/content/components/Image.mdx
```diff
@@ -0,0 +1,94 @@
...
+
+```jsx live
+<Image
+  as="picture"
```
rezrah, line 34: This example is setting an empty `source`. Could we add a test to guard this please? Should only appear if a `source` is defined.

Screenshot 2023-04-18 at 14 32 23

I assume most users of `as="picture"` would want to set some sources manually. Could we update this example to show how you would do this? I'm not quite clear on that myself right now either.

apps/docs/content/components/Image.mdx
```diff
@@ -0,0 +1,94 @@
...
+
+This component uses the `img` element by default.
+
+```jsx live
+<Image
+  src="https://via.placeholder.com/600x400/d3d9df/d3d9df.png"
+  alt="placeholder, blank area with an off-white background color"
+/>
+```
+
```
rezrah, line 27: Could we add an additional example using `srcSet` on an `img` please? That would be a super-helpful for copy-pasting.

apps/docs/content/components/Image.mdx
JoshBowdenConcepts: This would require access to children if we want to do that simply. I can include that but it would overcomplicate the component a bit possibly.

JoshBowdenConcepts: I think we should try to avoid doing this because it is going to complicate things quite a bit possibly.

packages/react/src/Image/Image.tsx
```diff
@@ -0,0 +1,82 @@
...
+        />
+      </span>
+    )
+  }
+  return (
+    <img
+      ref={ref}
+      alt={alt}
+      className={clsx(styles.Image, className)}
+      width={width ? width : undefined}
```
rezrah, line 77: is width undefined by default? wondering if you could skip the ternary.

apps/docs/content/components/Image.mdx
JoshBowdenConcepts: This has been added

apps/docs/content/components/Image.mdx
JoshBowdenConcepts: This has been updated to work when defined but not a custom setting.

apps/docs/content/components/Image.mdx
JoshBowdenConcepts: This has been applied

packages/react/src/Image/Image.tsx
JoshBowdenConcepts: This has been updated

apps/docs/content/components/Image.mdx
JoshBowdenConcepts: "This build" means the current build pushed to GitHub for me. It seems by local prettier is adding a space so it keeps adding and removing. It should be fixed now in the newest build.

packages/react/src/Image/Image.tsx
```diff
@@ -0,0 +1,82 @@
...
+
+type AspectRatio = '1:1' | '16:9' | '16:10' | '4:3' | 'custom'
+
+export type ImageProps = React.ImgHTMLAttributes<HTMLImageElement> & {
+  src: string
+  alt: string
+  ref?: React.Ref<HTMLImageElement>
+  as?: 'picture' | 'img'
+  aspectRatio?: AspectRatio
+  media?: string
```
rezrah: How would we handle multiple `media` declarations in usage? I think this works for one source but not more?

Take this example from w3:

```html
<picture>
  <source media="(min-width:650px)" srcset="img_pink_flowers.jpg">
  <source media="(min-width:465px)" srcset="img_white_flower.jpg">
  <img src="img_orange_flowers.jpg" alt="Flowers" style="width:auto;">
</picture>
```

Also this work for instances where you set `media` but not `srcSet` against it? I think they need to be more tightly coupled. Wonder if this could be a `sources` array, which takes in objects, which are later transformed into individual`<source>` elements internally. 

Alternatively, make children more composable and read them from the `children`.

cc. @danielguillan @josepmartins for additional perspectives.

JoshBowdenConcepts: I like the sources array here. I think adding children unnecessarily would overcomplicate the component. Would still love to hear what others think.

JoshBowdenConcepts: I wonder if we should make it strict to either apply the sources array or the srcSet and media attributes. What do you all think? If we allow both someone could overset it. Although I tested it and it doesn't break anything.

JoshBowdenConcepts: We have agreed to use sources as the property to set when creating these source elements. These will only be possible when picture is the as type. I have updated the component to reflect this.

apps/docs/content/components/Image.mdx
JoshBowdenConcepts: This has be discussed and resolved here: https://github.com/primer/brand/pull/229#discussion_r1171768591

TylerJDev (commented):
Just wanted to add that this looks good from an accessibility perspective! 👍

packages/react/src/Image/Image.tsx
```diff
@@ -0,0 +1,98 @@
...
+  if (as === 'picture') {
+    return (
+      <picture
+        className={clsx(
+          styles['Image__container'],
+          aspectRatio && styles[`Image--aspect-ratio-${aspectRatioResolver(aspectRatio)}`]
+        )}
+      >
+        {rest['sources'] &&
+          rest['sources'].map((source, index) => <source key={index} srcSet={source.srcset} media={source.media} />)}
```
joshblack, line 57: When we're reading `sources` like this, should we pull this out of `rest` so that it doesn't get spread onto `` below? Otherwise I think we end up with it stringifying into `sources="[object Object],[objectObject]"` like in the docs example

image

packages/react/src/Image/Image.tsx
```diff
@@ -0,0 +1,98 @@
...
+        {srcSet && <source srcSet={srcSet} media={media} />}
+        <img
+          ref={ref}
+          alt={alt}
+          width={width ? width : '100%'}
+          height={width ? width : '100%'}
```
joshblack, line 63: Totally a style thing so not important to change at all, just wanted to share another syntax for this that I think/hope is equivalent:

```suggestion
          width={width ?? '100%'}
          height={width ?? '100%'}
```

packages/react/src/Image/Image.test.tsx
```diff
@@ -0,0 +1,242 @@
...
+import {render, cleanup} from '@testing-library/react'
+import '@testing-library/jest-dom'
+import {axe, toHaveNoViolations} from 'jest-axe'
+
+import {Image} from './'
+
+expect.extend(toHaveNoViolations)
+
+describe('Image', () => {
+  afterEach(cleanup)
```
joshblack, line 11: I _think_ this is unnecessary in later versions of `@testing-library/react` 👀 

```suggestion
```

PR #237, opened 2023-04-19
Add new Grid component

rezrah:
## Summary

Adds a 12 column, responsive grid component for flexible layout.

🔗 [See documentation](https://primer-89832c6e10-26139705.drafts.github.io/components/Grid)
🔗 [See Storybook examples](https://primer-89832c6e10-26139705.drafts.github.io/storybook/?path=/story/components-grid--default)

## List of notable changes:

- **added** base Grid component to match Figma counterpart
- **added** prop for `enableOverlays` for visual aid

- Review [documentation](https://primer-89832c6e10-26139705.drafts.github.io/components/Grid) and verify the examples are clear and comprehensive
- Review API 
- Try it out in the docs sandbox

## Steps to test:

1. Open the Grid component documentation and/or storybook
2. Verify that Grid works as expected

- [Initial API exploration](https://www.figma.com/file/M4gYzqZSDHVBZNSzRcx9Ej/Primer-Brand-Grid-API-exploration?node-id=0%3A1&t=ahUTqblluwUODzmH-1)

## Contributor checklist:

## Reviewer checklist:

## Screenshots:

Grid in 12 column arrangement

Grid in asymmetrical arrangement


apps/docs/content/components/Grid.mdx
```diff
@@ -0,0 +1,274 @@
...
+```
+
+### Responsive behavior
+
+Use `start` with an `Object` of breakpoint-specific keys and `number` values to enable responsive behavior.
+
+```jsx live
+<Grid enableOverlay>
+  <Grid.Column
+    span={{xsmall: 12, small: 12, medium: 6, large: 4, xlarge: 3, xxlarge: 2}}
```
enstyled, line 68: Is/Will it be possible to set only a e.g. middle point where the span changes:

In Primer, `col-md-6` would set the column to be wide 12 units until the `md` breakpoint and upwards, so you don't have to specify `col-12 col-sm-12 col-md6 col-lg-6 col-xl-6`.

.changeset/blue-mugs-unite.md
```diff
@@ -0,0 +1,17 @@
...
+'@primer/react-brand': minor
+---
+
+Added `Grid` component
+
+Use `Grid` to create flexible and responsive grid-based layouts
+
+```jsx
+<Grid>
+  <Grid.Column span={4}></Grid.Column>
```
enstyled, line 11: Is it required for the name to be `Grid.Column` instead of just `Column`?

apps/docs/content/components/Grid.mdx
```diff
@@ -0,0 +1,274 @@
...
+```
+
+### Column offset
+
+Use `start` to provide a positioning offset.
+
+```jsx live
+<Grid enableOverlay>
+  <Grid.Column span={4}></Grid.Column>
+  <Grid.Column start={8}></Grid.Column>
```
enstyled, line 113: Nothing against the prop name `start`, but is there a reason for not keeping it consistent with the widely adopted `offset`?

apps/docs/content/components/Grid.mdx
```diff
@@ -0,0 +1,274 @@
...
+  ></Grid.Column>
+  <Grid.Column
+    span={{xsmall: 6, small: 6, medium: 6, large: 4, xlarge: 3, xxlarge: 2}}
+  ></Grid.Column>
+</Grid>
+```
+
+### Column offset
+
+Use `start` to provide a positioning offset.
```
enstyled, line 108: In the provided example, it's unclear why specifying only a `start` prop, makes the second column a span=6. AFAICT, the default span is 12?

apps/docs/content/components/Grid.mdx
```diff
@@ -0,0 +1,274 @@
...
+</Stack>
+```
+
+## Component props
+
+### Grid <Label>Required</Label>
+
+| name            | type                                                                             | default    | required | description                                           |
+| --------------- | -------------------------------------------------------------------------------- | ---------- | -------- | ----------------------------------------------------- |
+| `as`            | <PropTableValues values={['div', 'section', 'span']} addLineBreaks />            | `'div'`    | `false`  | The HTML element used to render the grid.             |
```
enstyled, line 257: Should we also add `ul` to this list? Which makes the columns as `li`?

enstyled:
Should we consider this behaviour expected: 

Screenshot 2023-04-19 at 13 46 38

in small browser size, at https://primer-89832c6e10-26139705.drafts.github.io/storybook/?path=/story/components-grid-features--nesting

apps/docs/content/components/Grid.mdx
rezrah: Is that a use-case you run into often @enstyled? 🤔 I wonder if there's an accessibility element to using list markup for this, which could result in unintentionally verbose callouts if we did it. cc. @TylerJDev.

apps/docs/content/components/Grid.mdx
rezrah: Yes, that's correct it's 12. It's documented in the prop table, but I could update this example to add span={12} and it would do the same thing.

apps/docs/content/components/Grid.mdx
rezrah: Yeah, this is a reallly verbose example 😀. It's intended to work the way you described because we use the min-width media query. When you change the examples in docs, does it not behave that way right now for you?

rezrah:
> Should we consider this behaviour expected? More specifically the overlapping columns:
> 
> in small browser size, at https://primer-89832c6e10-26139705.drafts.github.io/storybook/?path=/story/components-grid-features--nesting

The example is admittedly not great, but in real-world usage you'd use the responsive api to adjust the columns to fit. I'll update that example to fit better at narrow viewports 👍. The purpose of that story however, was to showcase that nesting is possible.

apps/docs/content/components/Grid.mdx
enstyled: I would not say I run into it often, but for semantic reasons, we might want to display a 3-4 items list in a responsive grid. Interested in hearing about the accessibility details.

apps/docs/content/components/Grid.mdx
enstyled: Perhaps my confusion is coming from `12` meaning "fill the available space". From my experience it means "full-width column, breaking on a new line if necessary"

apps/docs/content/components/Grid.mdx
enstyled: Great! It actually behaves as described when I change it in the docs.

Maybe we can update the examples with the shortest form in order to promote this practice?

packages/design-tokens/src/tokens/functional/components/grid/colors.json
```diff
@@ -0,0 +1,12 @@
...
+    "column": {
+      "overlay": {
+        "color": {
+          "value": "hsl(var(--base-color-scale-red-4-hsl) / 30%)",
+          "dark": "hsl(var(--base-color-scale-pink-2-hsl) / 30%)"
+        }
+      }
+    }
+  }
+}
```
langermank, line 12:
```suggestion
{
  "grid": {
    "column": {
      "bgColor": {
        "overlay": {
          "value": "hsl(var(--base-color-scale-red-4-hsl) / 30%)",
          "dark": "hsl(var(--base-color-scale-pink-2-hsl) / 30%)"
        }
      }
    }
  }
}
```

pattern-variant-property-scale

langermank: I guess this is just used for debugging though? I'm not sure it makes sense to expose them as CSS vars if its just used for docs/demo 🤔

packages/react/src/Grid/Grid.module.css
```diff
@@ -0,0 +1,738 @@
...
+  grid-gap: var(--grid-spacing-spacious);
+}
+
+.Grid__column {
+  grid-column: auto / span 12;
+}
+
+.Grid--gutter-condensed {
+  padding-left: var(--grid-spacing-condensed);
+  padding-right: var(--grid-spacing-condensed);
```
langermank, line 29:
```suggestion
  padding-inline: var(--grid-spacing-condensed);
```

packages/react/src/Grid/Grid.module.css
```diff
@@ -0,0 +1,738 @@
...
+}
+
+.Grid--gutter-normal {
+  padding-left: var(--grid-spacing-normal);
+  padding-right: var(--grid-spacing-normal);
```
langermank, line 34:
```suggestion
  padding-inline: var(--grid-spacing-normal);
```

packages/react/src/Grid/Grid.module.css
```diff
@@ -0,0 +1,738 @@
...
+}
+
+.Grid--gutter-spacious {
+  padding-left: var(--grid-spacing-spacious);
+  padding-right: var(--grid-spacing-spacious);
```
langermank, line 39:
```suggestion
  padding-inline: var(--grid-spacing-spacious);
```

apps/docs/content/components/Grid.mdx
rezrah: 👍 .. Happy to update it, but wanted to clarify that it's intentionally verbose to show the full-breadth of supported breakpoints. The only other way to know the values, is to review the source code. I could add a note above this example, to indicate that we use min-width?

apps/docs/content/components/Grid.mdx
rezrah: > we might want to display a 3-4 items list in a responsive grid

I'm slightly struggling to visualise this, do you have an example I can refer to?

I think a [Stack](https://primer.style/brand/components/Stack/) in responsive configuration could suit this use-case better.

packages/design-tokens/src/tokens/functional/components/grid/colors.json
rezrah: Thanks, will adjust. 

Re. using tokens, it's just for enabling color modes overrides, which the tokens do more easily without the need for custom css overrides in the stylesheet.

apps/docs/content/components/Grid.mdx
TylerJDev: Taking a look at it, having `<ul>` as an option might cause some unintended use. Looking at the `Stack` component mentioned by @rezrah, I think that might be a better choice when needing to utilize a list. I'd say, if we truly needed a list with this component then we could utilize ARIA roles (e.g. `role="list"` and `role="listitem"`), as these roles would convey the list to assistive technologies (screen readers, etc). 

I might not be thinking of all scenarios though, I'm only imagining this grid as primarily handling layout, so correct me if I'm wrong! :grin:

apps/docs/content/components/Grid.mdx
enstyled: Yes, fair point! Maybe we can add a note about the shorthand syntax or vice-verca?

apps/docs/content/components/Grid.mdx
enstyled: Yes, perhaps a Stack might be better suited for this case 👍🏼

JoshBowdenConcepts:
More of a general question is this grid only intended to handle one row?

apps/docs/content/components/Grid.mdx
josepmartins: I agree with @enstyled here, without looking at the prop table, I'd say that 12 is a "full-width column" example

packages/react/src/Grid/Grid.module.css
```diff
@@ -0,0 +1,736 @@
...
+.Grid__column--start-12 {
+  grid-column-start: 12;
+}
+
+/**
+ * Responsive style overrides
+ */
+
+/* XSMALL breakpoint */
+@media screen and (max-width: 20rem) {
```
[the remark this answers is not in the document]
rezrah: @josepmartins these aren't new, they are the existing ones that are already loaded into the stylesheet 👇 

Screenshot 2023-04-20 at 18 34 36

packages/design-tokens/src/tokens/functional/components/grid/grid-responsive.json
```diff
@@ -0,0 +1,39 @@
...
+    "spacing": {
+      "normal": {
+        "value": "var(--base-size-24)",
+        "responsive": {
+          "768px": {
+            "value": "var(--base-size-24)"
+          },
+          "1012px": {
+            "value": "var(--base-size-24)"
+          }
```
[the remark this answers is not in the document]
rezrah: Sorry, could you elaborate on this one please, not sure I'm following. This config generates this 👇 

Screenshot 2023-04-20 at 18 40 28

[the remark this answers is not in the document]
rezrah: The media queries are only there for future-proofing, as they aren't doing much right now.

packages/react/src/Grid/Grid.tsx
```diff
@@ -0,0 +1,146 @@
+import React, {PropsWithChildren, memo} from 'react'
+
+import {default as clsx} from 'clsx'
+
+import type {BaseProps} from '../component-helpers'
+
+/**
+ * Design tokens
+ */
+import '@primer/brand-primitives/lib/design-tokens/css/tokens/functional/components/grid/grid-responsive.css'
```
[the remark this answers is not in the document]
rezrah: 👍 fixed it now

apps/docs/content/components/Grid.mdx
rezrah: Updated the example to include a span value.

packages/react/src/Grid/Grid.module.css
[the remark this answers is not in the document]
josepmartins: Didn't know those existed 😶 🙏

packages/design-tokens/src/tokens/functional/components/grid/grid-responsive.json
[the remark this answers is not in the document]
josepmartins: That's a bit my point, if the actual design doesn't contemplate changes in the grid-spacing between viewports, should we provide/create these CSS tokens for consumers to modify it?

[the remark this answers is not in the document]
rezrah: Gotcha.. fair point. Thinking back, I think I made these responsive as a talking point. Do we want the values to change? I wasn't sure if the Figma one uses different spacing at different breakpoints, are you able to confirm @josepmartins? If not, I can remove the media queries altogether.

rezrah:
> More of a general question is this grid only intended to handle one row?


@JoshBowdenConcepts Grid can handle multiple rows. You can also stack multiple Grids. Up to the user how they prefer to use it, but latter creates a little more markup.

Screenshot 2023-04-21 at 11 51 44

packages/react/src/Grid/Grid.features.stories.tsx
```diff
@@ -0,0 +1,289 @@
...
+  return (
+    <Grid {...args}>
+      <Grid.Column span={6}></Grid.Column>
+      <Grid.Column span={6}></Grid.Column>
+    </Grid>
+  )
+}
+FiftyFifty.storyName = 'Ratio: 50:50'
+
+export const SixtyForty = args => {
```
JoshBowdenConcepts, line 168: I was just curious why this was named SixtyForty? Is it just because it is approximate? Will there be a need for a 60:40 grid that doesn't conform to our 12 column grid?
image

rezrah: It is approximate, yes. Many of our site designs use this column config, which is referred to as 60:40, but is actually like 58% and 42% in practice.

JoshBowdenConcepts (approved):
Looks great! Excited to start using this more!

packages/design-tokens/src/tokens/functional/components/grid/grid-responsive.json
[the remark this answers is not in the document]
rezrah: @josepmartins - have removed it for now, but question outstanding on whether the Figma grid has different column gap paddings at different breakpoints.. let me know if you find out.

apps/docs/content/components/Grid.mdx
rezrah: Added short blurb about this.

apps/docs/content/components/Grid.mdx
rezrah: Added some additional copy about this.

enstyled (approved):

packages/react/src/Grid/Grid.stories.tsx
```diff
@@ -0,0 +1,65 @@
...
+  component: Grid,
+  args: {
+    as: 'div',
+    gap: 'normal',
+    enableOverlay: true,
+    columns: [6, 6]
+  },
+  argTypes: {
+    gap: {control: 'radio', options: [undefined, 'condensed', 'normal', 'spacious'], defaultValue: undefined},
+    gutters: {control: 'radio', options: [undefined, 'condensed', 'normal', 'spacious'], defaultValue: undefined},
```
josepmartins, line 16: we can remove these two from the playground

packages/react/src/Grid/Grid.stories.tsx
```diff
@@ -0,0 +1,65 @@
+import {ComponentMeta, ComponentStory} from '@storybook/react'
+import React from 'react'
+import {Grid, GridProps, GridColumnIndex} from './Grid'
+
+export default {
+  title: 'Components/Grid',
+  component: Grid,
+  args: {
+    as: 'div',
+    gap: 'normal',
```
josepmartins, line 10: we can remove this option too

josepmartins (approved):
Left two non-blocking comments. Looks good to me 👍

packages/react/src/Grid/Grid.stories.tsx
rezrah: thanks 🙏 , removed

danielguillan (approved):
This is great! ✨

rfearing (commented):
claire-dancing

TylerJDev (approved):
Looks great! Couldn't find any concerns in regards to accessibility.

Issue #63, opened 2022-08-11
Next.js Link component incompatible with Hero CTAs


rezrah:
### Problem statement
Next.js encourage using the `Link` component declaratively to wrap any `<a>` tags.

E.g. 

```jsx
<Link href="/about">
  <a>About Us</a>
</Link>
```

The Hero doesn't use a composable API, which makes composition of Next.js link impossible.

```jsx
<Hero
  heading="..."
  description="..."
  primaryAction={{
    text: 'Primary action',
    href: '#',
  }}
  secondaryAction={{
    text: 'Secondary action',
    href: '#',
  }}
  align="..."
/>
```

### Suggestion
Rewrite the API for consistency with other composable API's.

```jsx
<Hero align="...">
  <Hero.Heading></Hero.Heading>
  <Hero.Description></Hero.Description>
  <Hero.Description></Hero.Description>
  <Link href="#">
    <Hero.PrimaryAction></Hero.PrimaryAction>
  </Link>
  <Link href="#">
    <Hero.SecondaryAction></Hero.SecondaryAction>
  </Link>
</Hero>
```

### References
[Next.js Link component docs](https://nextjs.org/docs/api-reference/next/link)

rfearing:
FYI: I spoke with Max and we don't recall any issues with `Next/link`. (Context: I mentioned we had some speedbumps with Next's link, but what I was thinking of is actually Next's image component.

stefankp:
Wondering if we should remove the forced vertical padding on Hero 🤔 None of the other components have that.

rfearing:
> Wondering if we should remove the forced vertical padding on Hero 🤔 None of the other components have that.

That makes sense to me

PR #239, opened 2023-04-19
`Hero` component refactor

TylerJDev:
## Summary

Provides the existing `Hero` component with a composable API. 

- [`Hero` documentation](https://primer-cfabcb7d9c-26139705.drafts.github.io/components/Hero)
- [`Hero` storybook](https://primer-cfabcb7d9c-26139705.drafts.github.io/storybook/?path=/story/components-hero--default) 

## List of notable changes:

- Refactors `Hero` to use composable API.

- The ability to extend `Hero` components (i.e. `Hero.Heading`, `Hero.Description`, etc).
- No visual changes being introduced.

## Steps to test:

1. Open `Hero` component in Storybook instance provided by this PR
2. Compare against existing `Hero` in production, there should be **no** visual change

- https://github.com/github/primer/issues/2064
- https://github.com/primer/brand/issues/63
- [API exploration](https://www.figma.com/file/8PgBjRPhhF7IGs4ZMCwUnH/Primer-Brand---Hero-refactor-API?node-id=0-1&t=wvuayIEaVHRYFMPO-0)

## Contributor checklist:

## Reviewer checklist:


packages/react/src/Hero/Hero.tsx
```diff
@@ -1,69 +1,90 @@
...
+      {children}
+    </Heading>
+  )
+})
+
+type HeroDescriptionProps = {
+  children: React.ReactNode | React.ReactNode[]
+}
+
+function HeroDescription({children}: PropsWithChildren<HeroDescriptionProps>) {
```
TylerJDev, line 56: Should we allow any props to be passed down to the `<Text>` component like we do with `HeroHeading`, or should we keep this locked down?

josepmartins: For descriptions, we keep them locked down atm (see Card, SectionIntro or the upcoming [Pillar](https://github.com/primer/brand/pull/262/files#diff-338f2497ae157e09ae2127ab08a052c0bae3161ab0cda467a64451efb5a1945fR90) examples).

packages/react/src/Hero/Hero.tsx
```diff
@@ -1,69 +1,94 @@
...
+import {Button, ButtonBaseProps} from '../Button'
+import {Heading, HeadingProps} from '../Heading'
 import {Text} from '../Text'
 
 import type {BaseProps} from '../component-helpers'
 import '@primer/brand-primitives/lib/design-tokens/css/tokens/functional/components/hero/base.css'
 
-type Action = {
-  text: string
+export type HeroProps = BaseProps<HTMLDivElement> & {
```
josepmartins, line 11: as this is a `<section>` wrapper, should we use `HTMLElement` instead?

packages/react/src/Hero/Hero.tsx
```diff
@@ -1,69 +1,94 @@
...
+  align?: 'start' | 'center'
+  children: React.ReactNode | React.ReactNode[] // Maybe use this from `FAQ`? type FAQRootProps = PropsWithChildren<BaseProps<HTMLElement>>
```
josepmartins, line 13: I'd say the way to go here using ReactNode is correct, but cc @rezrah

apps/docs/content/components/Hero.mdx
```diff
@@ -18,53 +27,73 @@ import {Hero} from '@primer/react-brand'
...
+| `id`            | `string`                                                                                           |         | Sets a custom id                               |
+| `ref`           | `React.RefObject`                                                                                  |         | Forward a Ref to the underlying DOM node       |
+| `size`          | <PropTableValues values={HeadingSizes} addLineBreaks />                                            |         | Override the default visual text size          |
+| `weight`        | <PropTableValues values={[...HeadingWeights, 'ResponsiveWeightMap']} addLineBreaks />              |         | Override the default visual text weight        |
+| `stretch`       | <PropTableValues values={[...HeadingStretch, 'ResponsiveStretchMap']} addLineBreaks />             |         | Override the default visual text stretch value |
+| `letterSpacing` | <PropTableValues values={[...HeadingLetterSpacing, 'ResponsiveLetterSpacingMap']} addLineBreaks /> |         | Override the default visual character spacing  |
```
josepmartins, line 79: For maintainability reasons, I like what we do in `SectionIntro` by adding a description and pointing to the `Heading` documentation. 

https://primer.style/brand/components/SectionIntro#sectionintroheading-required

josepmartins (commented):
As this will be considered a breaking change, I just have some concerns about removing the old `Hero` documentation for consumers in old versions. 

One temporary solution would be to keep the old docs in a separate `/Hero-deprecated` page and add a note on top of the updated page that link to these (example: https://primer.style/react/Radio)

packages/react/src/Hero/Hero.tsx
rezrah: I think for Hero we may need to be a bit more flexible @josepmartins. The previous Hero implementation was too restrictive, because you couldn't granularly control aspects of the text to meet requirements around brand expression. Hero's are unique in this sense. 

I think we should forward the `weight` and `size` props as an escape hatch. These are really useful, but we don't need to encourage it in documentation. It's there only if needed.

packages/react/src/Hero/Hero.tsx
rezrah: If you use PropsWithChildren, you shouldn't need to do this 🤔. Otherwise this is fine because it also accounts for multiple children without a fragment. Can we remove the comment after this is fixed too please.

apps/docs/content/components/Hero.mdx
```diff
@@ -18,53 +27,73 @@ import {Hero} from '@primer/react-brand'
...
 ```
 
 ### Alignment
 
 `Hero` text alignment can be adjusted by changing the `align` prop.
 
 ```jsx live
-<Hero
-  heading={
+<Hero>
```
rezrah, line 47: This example seems to be the same as the default one. Could we please update to use center alignment?

```suggestion
<Hero align="center">
```

apps/docs/content/components/Hero.mdx
rezrah: +1 to this☝️🙏

apps/docs/content/components/Hero.mdx
```diff
@@ -18,53 +27,73 @@ import {Hero} from '@primer/react-brand'
...
+| `size`          | <PropTableValues values={HeadingSizes} addLineBreaks />                                            |         | Override the default visual text size          |
+| `weight`        | <PropTableValues values={[...HeadingWeights, 'ResponsiveWeightMap']} addLineBreaks />              |         | Override the default visual text weight        |
+| `stretch`       | <PropTableValues values={[...HeadingStretch, 'ResponsiveStretchMap']} addLineBreaks />             |         | Override the default visual text stretch value |
+| `letterSpacing` | <PropTableValues values={[...HeadingLetterSpacing, 'ResponsiveLetterSpacingMap']} addLineBreaks /> |         | Override the default visual character spacing  |
+
+### Hero.Description
+
+| name        | type     | default | description       |
+| ----------- | -------- | ------- | ----------------- |
+| `className` | `string` |         | Text custom class |
```
rezrah, line 85:
```suggestion
| `className` | `string` |         | Sets a custom CSS class |
```

packages/react/src/Hero/Hero.stories.tsx
```diff
@@ -14,125 +15,105 @@ export default {
...
-Default.args = {
-  heading: 'This is my super sweet hero heading',
-  description:
-    'Lorem ipsum dolor sit amet, consectetur adipiscing elit. In sapien sit ullamcorper id. Aliquam luctus sed turpis felis nam pulvinar risus elementum.',
-  primaryAction: {
-    text: 'Primary action',
-    href: '#'
-  }
-}
+export const Centered: ComponentStory<typeof Hero> = _args => (
```
rezrah, line 29: Could we move all stories besides the default and playground into a new file called `Hero.features.stories.tsx` please? This is happening for all new components per an ADR. [Here's an example for ActionMenu](https://primer-cfabcb7d9c-26139705.drafts.github.io/storybook/?path=/story/components-actionmenu--default).

packages/react/src/Hero/Hero.tsx
```diff
@@ -1,69 +1,94 @@
...
+  )
+
+  return (
+    <section
+      className={clsx(styles.Hero, styles[`Hero--align-${align}`], className)}
+      ref={ref}
+      aria-labelledby="hero-section-brand-heading"
+      {...rest}
+    >
+      {React.Children.map(children, child => {
```
rezrah, line 32: could we hoist these outside the render and memoize them like you did for buttons? And when that happens, wonder if you could do both filters in one loop rather than 2. That would be a neat optimisation, as they are otherwise doing the same thing but with inverted conditions.

packages/react/src/Hero/Hero.tsx
```diff
@@ -1,69 +1,94 @@
...
+          }
+        }
+      })}
+      <div className={styles['Hero-actions']}>{HeroActions}</div>
+    </section>
+  )
+})
+
+type HeroHeadingProps = {
+  children: React.ReactNode | React.ReactNode[]
```
rezrah, line 45: You should be able to use PropsWithChildren to avoid needing to declare children each time.

packages/react/src/Hero/Hero.tsx
TylerJDev: Thanks for the feedback y'all! I implemented both `weight` and `size`, while keeping the other props locked down. I added a small note in the documentation for it under the props table only explaining that it's available.

packages/react/src/Hero/Hero.tsx
TylerJDev: Good catch! Swapped this for `HTMLElement`.

packages/react/src/Hero/Hero.tsx
TylerJDev: Swapped this out with `PropsWithChildren`!

apps/docs/content/components/Hero.mdx
TylerJDev: Great suggestion! Did this for `Hero.Description` too.

packages/react/src/Hero/Hero.tsx
TylerJDev: Yup! Replaced this with a `.reduce` so that we'd only run one loop. Let me know if this isn't the right way to go about it!

PR #262, opened 2023-05-04
Add Pillar component

josepmartins:
## Summary

The `Pillar` component is an informational block of text (heading + description), with an optional Octicon.

## List of notable changes:

- **added** `Pillar` component, stories, tests and documentation
- **updated** the `Card` component and docs to use `HeadingProps` on `Card.Heading` as well as we do in `Pillar` or `SectionIntro`

- Storybook: https://primer-24f973e4ee-26139705.drafts.github.io/storybook/?path=/story/components-pillar--default
- Documentation: https://primer-24f973e4ee-26139705.drafts.github.io/components/Pillar

- Figma API exploration: https://www.figma.com/file/Wj54hOacJWuohfvtfzYf1p/Primer-Brand-Pillar-API-exploration?node-id=0%3A1&t=22FwxsPJFSyXO1KT-1
- Figma component: https://www.figma.com/file/BJ95AjraesmRCWsKA013GS/Primer-Brand?node-id=6849-31364&mode=design
- Tracking issue: https://github.com/github/primer/issues/1968

## Contributor checklist:

## Reviewer checklist:

- [x] Check that pull request and proposed changes adhere to our [contribution guidelines](../../CONTRIBUTING.md) and [code of conduct](../../CODE_OF_CONDUCT.md)
- [x] Check that tests prove the feature works and covers both happy and unhappy paths
- [x] Check that there aren't other open Pull Requests for the same update/change

## Screenshots:

<table>
<tr>
</tr>
<tr>
<td valign="top">

 </td>
<td valign="top">

</td>
</tr>
</table>


apps/docs/content/components/Pillar.mdx
```diff
@@ -0,0 +1,131 @@
...
+</Stack>
+```
+
+## Component props
+
+### Pillar
+
+| name    | type              | default | required | description               |
+| ------- | ----------------- | ------- | -------- | ------------------------- |
+| `align` | `start`, `center` | `start` | `true`   | Aligns the pillar content |
```
danielguillan, line 104:
```suggestion
| `align` | `start`, `center` | `start` | `false`   | Aligns the pillar content |
```

packages/design-tokens/src/tokens/functional/components/pillar/colors.json
danielguillan, line 1: Is there a _simple_ way these colors could be shared with the Card Icon colors, so we don't have to update these in two different places?

packages/react/src/Pillar/Pillar.tsx
```diff
@@ -0,0 +1,119 @@
...
+export const defaultPillarIconColor = PillarIconColors[0]
+export type PillarProps = {
+  /**
+   * Valid children include Pillar.Image, Pillar.Heading, and Pillar.Description
+   */
+  children:
+    | React.ReactNode
+    | React.ReactElement<PillarIconProps>
+    | React.ReactElement<PillarHeadingProps>
+    | React.ReactElement<PillarDescriptionProps>
```
danielguillan, line 31:
```suggestion
  children:
    | React.ReactElement<PillarIconProps>
    | React.ReactElement<PillarHeadingProps>
    | React.ReactElement<PillarDescriptionProps>
```
I guess we don't want to allow any arbitrary React nodes?

packages/react/src/Pillar/Pillar.tsx
```diff
@@ -0,0 +1,119 @@
...
+
+  /**
+   * Aligns the pillar content
+   */
+  align?: 'start' | 'center'
+} & BaseProps<HTMLDivElement> &
+  React.ComponentPropsWithoutRef<'div'>
```
danielguillan, line 38: Should we maybe support an isomorphic `as` prop so pillars can be set to a more semantic element such as `article`?

packages/react/src/Pillar/Pillar.tsx
```diff
@@ -0,0 +1,119 @@
...
+
+const PillarRoot = forwardRef<HTMLDivElement, PillarProps>(({children, className, align = 'start', ...props}, ref) => {
+  const filteredChildren = React.Children.toArray(children).filter(child => {
+    if (React.isValidElement(child) && typeof child.type !== 'string') {
+      if (
+        isFragment(child) ||
```
danielguillan, line 44: I might be reading this incorrectly, but does this mean that the following code would be accepted?

```jsx
<Pillar>
  <><Heading>Some arbitrary heading</Heading></>
</Pillar>
```

I understand why we need to support fragments, but we might want to iterate over its children to ensure they only contain valid elements?

danielguillan (approved):
Looks great! Just left a few minor comments.

packages/design-tokens/src/tokens/functional/components/pillar/colors.json
josepmartins: I can think of creating a shared `Icon` component to retrofit `Card` and `Pillar`, to avoid multiple repeated styles across the library

packages/react/src/Pillar/Pillar.features.stories.tsx
```diff
@@ -0,0 +1,163 @@
...
+            </g>
+            <path d="m8.01306152 30v2" fillRule="nonzero" stroke="#79b8ff" />
+            <g stroke="#2088ff" strokeLinejoin="round">
+              <path d="m10.5599344 15-3.06918743 3.0651125-1.49074697-1.4854753" />
+              <path d="m10.5599344 6-3.06918743 3.06511252-1.49074697-1.4854753" />
+            </g>
+          </g>
+        </svg>
+      }
+      color="purple"
```
rezrah, line 86: Think this is redundant, as it won't apply to custom icons? ☝️

packages/react/src/Pillar/Pillar.tsx
```diff
@@ -0,0 +1,130 @@
...
+import {Icon as IconProps} from '@primer/octicons-react'
+
+export const PillarIconColors = Colors
+
+export const defaultPillarIconColor = PillarIconColors[0]
+export type PillarProps<C extends keyof JSX.IntrinsicElements = 'div'> = React.HTMLAttributes<C> & {
+  /**
+   * The HTML element used to render the grid.
+   */
+  as?: C | 'div' | 'span' | 'article'
```
rezrah, line 25: Do we need `span`? Feels semantically incorrect, as the contents of the pillar would be block elements. I can imagine some static analysis tools warning about this.

packages/react/src/Pillar/Pillar.tsx
```diff
@@ -0,0 +1,130 @@
...
+ */
+import styles from './Pillar.module.css'
+import {Icon as IconProps} from '@primer/octicons-react'
+
+export const PillarIconColors = Colors
+
+export const defaultPillarIconColor = PillarIconColors[0]
+export type PillarProps<C extends keyof JSX.IntrinsicElements = 'div'> = React.HTMLAttributes<C> & {
+  /**
+   * The HTML element used to render the grid.
```
rezrah, line 23:
```suggestion
   * The HTML element used to render the Pillar.
```

packages/react/src/Pillar/Pillar.tsx
```diff
@@ -0,0 +1,130 @@
...
+   */
+  as?: C | 'div' | 'span' | 'article'
+  /**
+   * Valid children include Pillar.Image, Pillar.Heading, and Pillar.Description
+   */
+  children:
+    | React.ReactNode
+    | React.ReactElement<PillarIconProps>
+    | React.ReactElement<PillarHeadingProps>
+    | React.ReactElement<PillarDescriptionProps>
```
rezrah, line 33: Wonder if we could just use PropsWithChildren? I think declaring React.ReactNode will make the other types redundant.

packages/react/src/Card/Card.tsx
```diff
@@ -160,15 +160,7 @@ function CardIcon({
...
-      {typeof Icon === 'function' ? (
-        <Icon size={20} />
-      ) : (
-        React.isValidElement(Icon) &&
-        React.cloneElement(Icon, {
-          width: 20,
-          height: 20
-        })
-      )}
+      {typeof Icon === 'function' ? <Icon size={20} /> : React.isValidElement(Icon) && React.cloneElement(Icon)}
```
rezrah, line 163: Should our Octicons be 24px instead of 20px?

packages/react/src/Pillar/Pillar.features.stories.tsx
```diff
@@ -0,0 +1,163 @@
...
+      <Pillar.Description>
+        Enables you to rapidly search, navigate, and understand code, right from GitHub.com.
+      </Pillar.Description>
+    </Pillar>
+  )
+}
+
+export const IconColors: ComponentStory<typeof Pillar> = () => {
+  return (
+    <Stack direction="horizontal" gap={'spacious'} style={{flexWrap: 'wrap'}}>
```
rezrah, line 26: Given the inline styles here, would a Grid work better here?

packages/react/src/Pillar/Pillar.tsx
```diff
@@ -0,0 +1,130 @@
...
+    return (
+      <Heading size={size} className={clsx(styles.Pillar__heading, className)} ref={ref} as={as} {...rest}>
+        {children}
+      </Heading>
+    )
+  }
+)
+
+type PillarDescriptionProps = BaseProps<HTMLParagraphElement> & {
+  children: React.ReactNode | React.ReactNode[]
```
rezrah, line 102: Could use PropsWithChildren here to avoid needing to redeclare this

packages/react/src/Pillar/Pillar.tsx
```diff
@@ -0,0 +1,130 @@
...
+}
```
rezrah, line 103: Could we extend TextProps here too? In case someone needs to customise certain values like variant or `as`

packages/react/src/Pillar/Pillar.tsx
josepmartins: oopsie

packages/react/src/Pillar/Pillar.tsx
josepmartins: I'd like to hold on providing such props.

- I can't think of a case where we need to modify the description to an html that isn't a paragraph via `as` props. 
- For variants and other font styles this is similar to the `Card` component description, where we agreed with site designers to keep it simple to one color/weight/size variant, until a strong use case or a design need pops up that confronts that.

PR #284, opened 2023-05-22
Fix type issues in `Hero`

TylerJDev:
## Summary

Adjusts types for `HeroActions` type. 
* Remove `Omit` for `variant` prop
* Fix `leadingVisual` and `trailingVisual` type issue

## List of notable changes:

- Adjusts TS types for `HeroActions`

- The ability to use props stemming from `Button`
- Added story to showcase icon support and `variant` usage

## Steps to test:

* Ensure no TypeScript errors are present when utilizing props from the [Brand `Button` component](https://primer-5dccbf8f09-26139705.drafts.github.io/components/Button#component-props).

## Contributor checklist:

## Reviewer checklist:

### DIFF
--- a/packages/react/src/Button/Button.tsx
+++ b/packages/react/src/Button/Button.tsx
@@ -1,8 +1,10 @@
 import clsx from 'clsx'
 import React, {forwardRef, useCallback, type Ref, ReactElement} from 'react'
-import {ExpandableArrow} from '../ExpandableArrow'
-import {Text} from '../Text'
+import type {Icon} from '@primer/octicons-react'
 import '@primer/brand-primitives/lib/design-tokens/css/tokens/functional/components/button/colors-with-modes.css'
+import {Text} from '../Text'
+import {ExpandableArrow} from '../ExpandableArrow'
+
 import type {BaseProps} from '../component-helpers'
 import styles from './Button.module.css'
 
@@ -16,11 +18,11 @@ export type ButtonBaseProps = {
   /**
    * The leading visual appears before the button content
    */
-  leadingVisual?: ReactElement
+  leadingVisual?: ReactElement | Icon
   /**
    * The trailing visual appears after the button content
    */
-  trailingVisual?: ReactElement
+  trailingVisual?: ReactElement | Icon
   /**
    * The styling variations available in Button
    */
@@ -80,7 +82,7 @@ export const _Button = forwardRef(
     const isDisabled =
       disabled || ariaDisabled === 'true' || (typeof ariaDisabled === 'boolean' && ariaDisabled === true)
 
-    const returnValidComponent = useCallback((component?: ReactElement) => {
+    const returnValidComponent = useCallback((component?: ReactElement | Icon) => {
       if (React.isValidElement(component)) {
         return component
       }
--- a/packages/react/src/Hero/Hero.tsx
+++ b/packages/react/src/Hero/Hero.tsx
@@ -74,9 +74,7 @@ function HeroDescription({size = '400', weight, children}: PropsWithChildren<Her
 
 type HeroActions = {
   href: string
-  leadingVisual?: React.ReactNode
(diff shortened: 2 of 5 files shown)


packages/react/src/Hero/Hero.tsx
```diff
@@ -74,7 +74,9 @@ function HeroDescription({size = '400', weight, children}: PropsWithChildren<Her
 
 type HeroActions = {
   href: string
-} & Omit<ButtonBaseProps, 'variant'>
+  leadingVisual?: React.ReactNode
```
rezrah, line 77: Can these be picked from ButtonBaseProps instead? And any ideas why this isn't getting inherited? Feels redundant to maintain these props here too.

packages/react/src/Hero/Hero.tsx
```diff
@@ -74,7 +74,9 @@ function HeroDescription({size = '400', weight, children}: PropsWithChildren<Her
...
+  trailingVisual?: React.ReactNode
+} & Omit<ButtonBaseProps, 'trailingVisual' | 'leadingVisual'>
```
TylerJDev, line 79: This was one of the few ways I was able to make `leadingVisual` and `trailingVisual` play nice with Octicons with its `Icon` type. It seems that there's a mismatch in props against `ReactElement` that TypeScript was complaining about, which is only a problem when using `Button` here, and not directly. 

A alternative would be replacing `ReactElement` type in `Button` for `leadingVisual` and `trailingVisual` with `React.ReactNode`, but I only wanted to adjust the `Hero` component, and not change anything within `Button`.

packages/react/src/Hero/Hero.tsx
TylerJDev: I left a comment on this https://github.com/primer/brand/pull/284#discussion_r1200626396 - I'm not entirely sure why it's throwing an error here. I get the following TS error: `Type 'FunctionComponent<IconProps>' is missing the following properties from type 'ReactElement<any, string | JSXElementConstructor<any>>': type, props, key`, which isn't present when utilizing `Button` and the "visual" props directly.

Utilizing `React.ReactNode` fixed it, as it seems to be tied to `ReactElement`. I kept the change in this component via `Omit` as to not touch `Button`. I can do some more digging in the meantime though.

TylerJDev: @rezrah - Digging into this a bit deeper it seems like this is tied to the `ReactElement` type in `Button` when utilizing an Octicon. ReactElement does not appear to apply when the type is a functional component `React.FC`, which is what `Octicon` seems to return as its type.

When used directly, `Button` doesn't seem to check the types of the props that are passed to it, which I think is why it doesn't cause an issue when `Button` is used directly with `leadingVisual` and `trailingVisual`. 

Could we replace the `ReactElement` type in `Button.tsx` with `React.ReactNode` or add both, (e.g. 
 `leadingVisual?: ReactElement | React.ReactNode`)? It seems like we do this within `Label` and `TextInput` too.
 
 Let me know your thoughts!

packages/react/src/Button/Button.tsx
```diff
@@ -16,11 +18,11 @@ export type ButtonBaseProps = {
   /**
    * The leading visual appears before the button content
    */
-  leadingVisual?: ReactElement
+  leadingVisual?: ReactElement | Icon
```
rezrah, line 21: @TylerJDev - here's the change ☝️

packages/react/src/Hero/Hero.tsx
rezrah: Hey Tyler, you're right that it needs to be fixed in Button. Putting this here for posterity, as we spoke about this over slack. The type needs to extend Icon from Octicons, doing that should fix it. I've pushed the change in [this commit](https://github.com/primer/brand/pull/284/commits/2a9a062f645cf71c7d687d9d6aa9f7dfd992a9ba).

PR #317, opened 2023-06-20
Card a11y fixes

mperrotti:
## Summary

Addresses @TylerJDev's feedback in https://github.com/github/primer/issues/1809

Specifically:
- Hide native SVGs by default, similarly to the Pillar implementation
- Hide trailing SVG after the CTA link (ExpandableArrow)
- Adjust markup in documentation examples to show Card.Heading before Card.Label
- Ensure link text only contains text from Card.Heading, possibly by using pseudo-element pattern

## List of notable changes:
- The card container element is no longer `<a>`, it's `<div>`. **This might be a breaking change** because a `div` has different attributes than an `a`.

- Confirm that the feedback has been addressed
- Confirm there have been no visual regressions

## Steps to test:

1. Open the component in CI-deployed preview environment
2. Go to any of the Card stories in Storybook
3. Verify that the a11y eng feedback has been addressed
  - Hide native SVGs by default, similarly to the Pillar implementation
  - Hide trailing SVG after the CTA link (ExpandableArrow)
  - Adjust markup in documentation examples to show Card.Heading before Card.Label
  - Ensure link text only contains text from Card.Heading, possibly by using pseudo-element pattern
5. Verify that there are no visual or a11y regressions

## Contributor checklist:

## Reviewer checklist:

## Screenshots:

N/A - there should be no visual changes


packages/react/src/Card/Card.tsx
```diff
@@ -62,31 +63,31 @@ const CardRoot = forwardRef<HTMLAnchorElement, CardProps>(
         setIsHovered(!isHovered)
         onMouseEnter?.(event)
       },
-      [onMouseEnter, isHovered],
```
mperrotti, line 65: These code style changes are not intentional - this is just how Prettier auto-formatted when I saved the file.

packages/react/src/Card/Card.tsx
```diff
@@ -106,30 +107,32 @@ const CardRoot = forwardRef<HTMLAnchorElement, CardProps>(
...
-        onFocus={handleOnFocus}
-        onBlur={handleOnBlur}
-        ref={cardRef}
-        {...props}
-      >
-        {filteredChildren}
+      <div className={clsx(styles.Card, hasIcon && styles['Card--has-icon'], className)} ref={cardRef} {...props}>
+        {React.Children.map(filteredChildren, child => {
+          if (React.isValidElement(child) && typeof child.type !== 'string' && child.type === CardHeading) {
+            return React.cloneElement<CardHeadingProps>(child as React.ReactElement<CardHeadingProps>, {
```
mperrotti, line 117: Are we ok with this use of cloneElement? Or do you think it would be better to use React Context?

packages/react/src/Card/Card.tsx
```diff
@@ -106,30 +107,32 @@ const CardRoot = forwardRef<HTMLAnchorElement, CardProps>(
...
+            })
+          }
+          return child
+        })}
         <div className={styles.Card__action}>
           <Text as="span" size="300" className={clsx(stylesLink['Link--label'])}>
             {ctaText}
           </Text>
-          <ExpandableArrow className={stylesLink['Link-arrow']} expanded={isHovered || isFocused} />
+          <ExpandableArrow className={stylesLink['Link-arrow']} expanded={isHovered || isFocused} aria-hidden="true" />
```
mperrotti, line 131: Is it safe to just add `aria-hidden` directly in the ExpandableArrow component? Or are there cases where we want it to be "visible" to ATs?

mperrotti:
Keyboard navigation is broken. Will fix soon.

packages/react/src/Card/Card.tsx
TylerJDev: I would say it's okay to directly apply `aria-hidden` to `ExpandableArrow`, as it should never be without context (i.e. `ctaText`), as this would be an issue itself.

I think that `ExpandableArrow` has a `hidden` prop we could use too which should do the same thing.

packages/react/src/Card/Card.tsx
rezrah: Cloning is fine 👍, we tend to use it for these kind of use-cases. Only nit would be to hoist it out of the render and wrap in a callback.

rezrah (commented):
Looks great 👍 thanks @mperrotti 

> The card container element is no longer <a>, it's <div>. This might be a breaking change because a div has different attributes than an a.

As we're pre-v1, we don't bump the major. That said, **could we add a changeset to this PR please**, which also indicates the change from `a` to `div`? 

Tagging in @josepmartins for another set of 👀, and as the original component author.

@mperrotti - before we land this, could we update the status in docs to 'reviewed for accessibility'? 

E.g.
Screenshot 2023-06-26 at 10 23 52

One final thought: Should the scaling animation apply to focus states too? cc. @TylerJDev 

https://github.com/primer/brand/assets/13340707/0543cdab-acd9-4994-8463-72eed7916ef3

TylerJDev:
@rezrah - We could add the animations on focus, though I don't think this is a hard requirement, since the focus indicator present should be sufficient enough. The only thing I'd add is, if we do allow the animation to appear on focus, we want to ensure that it does not play when `prefers-reduced-motion`.

TylerJDev (approved):
As far as accessibility, this looks great! I really appreciate the swap to only utilizing the title as the link! All feedback from the a11y eng review has been implemented.

packages/react/src/Card/Card.module.css
```diff
@@ -19,12 +19,36 @@
...
+  outline: none;
+}
+
+.Card__link::before {
+  content: '';
+  position: absolute;
+  top: 0;
+  left: 0;
+  bottom: 0;
+  right: 0;
```
josepmartins, line 49: I think we could use `inset: 0;`  as a shorthand here.

josepmartins (approved):
LGTM! Thanks for doing it @mperrotti 🚀 

Left a nitty/non-blocker CSS comment

packages/react/src/Card/Card.module.css
mperrotti: Oh right, I always forget about that property. Thanks!!

mperrotti:
> before we land this, could we update the status in docs to 'reviewed for accessibility'?

@rezrah - Sure! How do I update that? I can't figure it out...

josepmartins:
> > before we land this, could we update the status in docs to 'reviewed for accessibility'?
> 
> @rezrah - Sure! How do I update that? I can't figure it out...

In the react.mdx file, you should add the `a11yReviewed: true` variable in the frontmatter. Heading example: 
https://github.com/primer/brand/blob/71a28fd524569dab1ed927acc406cb7034447d39/apps/docs/content/components/Heading/react.mdx#L4

PR #347, opened 2023-07-17
Added Box component

rezrah:
## Summary

Part of https://github.com/github/primer/issues/2430

Adding an experimental `Box` component, which is a helper component for fine-grained styling of elements on a page.

🔗 [React documentation](https://primer-63012c5322-26139705.drafts.github.io/components/Box/react)
🔗 [Storybook examples](https://primer-63012c5322-26139705.drafts.github.io/storybook/?path=/story/components-box-features--padding)
🔗 Interface guidelines (WIP)

### PR Readiness
- [x] Base component
- [x] React component docs
- [x] Uses logical operators
- [x] Interface guidelines
- [x] Storybook stories
- [x] Unit tests
- [x] Visual regression tests

## List of other notable changes:

- **renamed** pre-existing Box imports from PRC to Box to circumvent naming collisions with our own box
- **added** new prop for `flexWrap` to Stack
- **updated** some border colors, which were previously incorrect

- Review documentation
- Try the canary release

## Contributor checklist:

## Reviewer checklist:

### Uniform spacing 
Screenshot 2023-07-15 at 20 43 48
Screenshot 2023-07-15 at 20 43 37

### Directional spacing  
Screenshot 2023-07-15 at 20 44 22

### Border radius
Screenshot 2023-07-15 at 20 44 27

### Border widths
Screenshot 2023-07-15 at 20 44 32

### Border colors
Screenshot 2023-07-15 at 20 44 41


packages/react/src/Box/Box.tsx
```diff
@@ -0,0 +1,201 @@
+import {default as clsx} from 'clsx'
+import React, {PropsWithChildren, useMemo} from 'react'
+import {AnimateProps, useAnimation} from '../animation'
+import type {BaseProps} from '../component-helpers'
+import {BaseSizeScale} from '../constants'
```
JoshBowdenConcepts, line 5: Should this include "condensed", "normal", and "spacious"? I think we include these in most of our spacing use cases.

rezrah: The base scale is a number array, which corresponds directly to our base spacing tokens (from 4 to 128). the terms like 'condensed', 'normal' and 'spacious' are semantic / functional names so they aren't related to our size scale. The reusing of those names is coincidental.

packages/react/src/Box/Box.features.module.css
```diff
@@ -0,0 +1,48 @@
+.spacingContainer {
```
langermank, line 1: Are these styles just for documentation? I wonder if they should live in Storybook instead.

langermank (approved):
CSS looks good 👍 love the logical properties ❤️ 

Might be worth writing a PostCSS mixin to generate the output, but I don't think that's a blocker.

packages/react/src/Box/Box.features.module.css
rezrah: 👍 Those don't go into the user bundle, they're only imported into Storybook. Is that what you mean by living in Storybook?

packages/react/src/Stack/Stack.tsx
```diff
@@ -19,6 +19,9 @@ export const defaultStackSpacing = StackSpacingVariants[1]
 export const StackAlignItemVariants = ['center', 'flex-start', 'flex-end'] as const
 type StackAlignItemVariants = (typeof StackAlignItemVariants)[number]
 
+export const StackFlexWrapVariants = ['wrap', 'nowrap'] as const
```
danielguillan, line 22: A couple of questions:
- Is the `Flex` prefix really needed? We don't have it on other props that map to flex properties (i.e., `direction`).
- Could this prop be a `boolean` instead of a list of variants? I don't think we plan to support other values such as `wrap-reverse` or `initial`.

danielguillan (approved):
🚀 This is great! Just left a couple of minor questions.

apps/docs/src/@primer/gatsby-theme-doctocat/components/live-preview-wrapper.js
```diff
@@ -1,6 +1,6 @@
 import React, {useState} from 'react'
 import {
-  Box,
+  Box as PRCBox,
```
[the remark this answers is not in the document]
rezrah: Good question. So it's aliased as Container in our live previews, which are public facing. But aliased as PRCBox (which is ugly AF) in our source code. It's really just a case of calling it different things for different audiences. When you're reading and contributing to source code, reading PRCBox makes it immediately obvious what it is. PRC however, doesn't have any relevance to non-staff. Hope that makes sense. Let me know if you have a better solution though.

packages/react/src/Stack/Stack.tsx
rezrah: > Is the Flex prefix really needed? 

Yeah. It was really to avoid this: `wrap="wrap"`, which just felt weird. We also have drawn a line in the sand around this component being a flexbox wrapper, so I felt it was fine to be clear about that in the prop name, but I agree it's not consistent with the pre-existing API. Would you prefer wrap? I don't have super strong opinions on it.

> I don't think we plan to support other values such as wrap-reverse or initial.

I made it a string union just in case we did need to do that in future. It's more future-proofing than anything. I think wrap-reverse could be useful in some scenario's, so maybe worth keeping the API open for those scenario's?

danielguillan: If we keep it as a string union, then I think I'd go with flexWrap for clarity. 👍

.changeset/fresh-turtles-taste.md
```diff
@@ -0,0 +1,13 @@
+---
+'@primer/react-brand': patch
+---
+
+Added `flexWrap` prop to Stack component to enable automatic wrapping of flex items.
+
+```jsx
+<Stack direction="horizonta" flexWrap="wrap">
```
josepmartins, line 8:
```suggestion
<Stack direction="horizontal" flexWrap="wrap">
```

apps/docs/content/components/Box/index.mdx
```diff
@@ -0,0 +1,50 @@
+---
+title: Box
+description: Use a box to simplify the process of applying one-off styles to an element
+---
+
+import CustomVideoPlayer from '../../../src/components/custom-video-player'
+import ComponentLayout from '../../../src/layouts/component-layout'
+export default ComponentLayout
+
+![Three empty cards inside a grey background with rounded corners, which is also offset from the edge and including several annotations. First indicating distance between the background area and canvas edge to indicate margin controls. Second showing padding controls between the cards and background outer edge. Third pointing towards background color and lastly one pointing at the rounded corner to illustrate border radius options.](https://github.com/primer/brand/assets/13340707/211be3c8-0f33-47ba-8236-9b19780d2a86)
```
josepmartins, line 10: We normally use pointers for parts or sub elements in the component, but not for its options. I think that for this specific component we might want to skip the intro image, as it seems a bit difficult to express such an abstract component/wrapper.


```suggestion
![Three empty cards inside a grey background with rounded corners, which is also offset from the edge and including several annotations. First indicating distance between the background area and canvas edge to indicate margin controls. Second showing padding controls between the cards and background outer edge. Third pointing towards background color and lastly one pointing at the rounded corner to illustrate border radius options.](https://github.com/primer/brand/assets/13340707/211be3c8-0f33-47ba-8236-9b19780d2a86)
```

apps/docs/content/components/Box/index.mdx
```diff
@@ -0,0 +1,50 @@
...
+
+## Usage
+
+The box component is a versatile layout component that is intended to simplify the process of adding one-off styles such as padding, margin, background color, and border to (or around) an element.
+
+It provides convenient and comprehensive access to Primer Brand's underlying Primitives, such as the spacing scale and functional design tokens for color and borders.
```
josepmartins, line 16:
```suggestion
The box provides convenient and comprehensive access to Primer Brand's underlying Primitives, such as the spacing scale and functional design tokens for color and borders.
```

apps/docs/content/components/Box/index.mdx
```diff
@@ -0,0 +1,50 @@
...
+
+Inner spacing is configurable on all sides or specific directions. Box utilises the CSS padding specification, using logical operators to aid bi-directional reading support.
+
+Additionally, fine-grained control over padding at specific viewports is available to ensure optimal, responsive layout on different screen sizes.
+
+### Margin
+
+Outer spacing is also configurable on all sides or specific directions. Box utilises the CSS margin specification, similarly using logical operators to aid bi-directional reading support.
+
+### Background colors
```
josepmartins, line 32:
```suggestion
### Background color
```

apps/docs/content/components/Box/index.mdx
```diff
@@ -0,0 +1,50 @@
...
+
+## Accessibility
+
+Box is a purely decorative element by default. It sets no intrinsic roles, ARIA attributes, or keyboard shortcuts.
+
+Any interactive content within a box should be given their own appropriate accessibility treatment based on the specific workflow of your application.
+
+## Related components
+
+- [Stack](/components/Stack): to enable layout of its immediate children along the vertical or horizontal axis.
```
josepmartins, line 50: I think we should we add the `Grid` component here too, to list all the layout enablers

packages/react/src/Box/Box.tsx
```diff
@@ -0,0 +1,210 @@
...
+
+export const BoxBorderWidthOptions = ['thin', 'thick', 'thicker'] as const
+type BorderWidthOptions = (typeof BoxBorderWidthOptions)[number]
+
+export const BoxBorderColorOptions = ['default', 'muted', 'subtle'] as const
+type BorderColorOptions = (typeof BoxBorderColorOptions)[number]
+
+type BorderStyleOptions = Extract<React.CSSProperties['borderStyle'], 'solid' | 'none'>
+
+type BoxProps = {
```
josepmartins, line 33: Should we add an `as` property here? I can image box being used to wrap a group of components to add a background color, which can be a `section` or `article` html candidate. Or as a wrapper for a list of `Cards` or `Pillars`, so the Box can be a `ul` item too.

apps/docs/content/components/Box/index.mdx
```diff
@@ -0,0 +1,50 @@
...
+
+![Three empty cards inside a grey background with rounded corners, which is also offset from the edge and including several annotations. First indicating distance between the background area and canvas edge to indicate margin controls. Second showing padding controls between the cards and background outer edge. Third pointing towards background color and lastly one pointing at the rounded corner to illustrate border radius options.](https://github.com/primer/brand/assets/13340707/211be3c8-0f33-47ba-8236-9b19780d2a86)
+
+## Usage
+
+The box component is a versatile layout component that is intended to simplify the process of adding one-off styles such as padding, margin, background color, and border to (or around) an element.
+
+It provides convenient and comprehensive access to Primer Brand's underlying Primitives, such as the spacing scale and functional design tokens for color and borders.
+
+While it is highly customizable and can be used to create complex layouts, it is important to use it judiciously and only when necessary.
```
josepmartins, line 18:
```suggestion
While the box component is highly customizable and can be used to create complex layouts, it is important to use it judiciously and only when necessary.
```

josepmartins (approved):
Looks good! just left some non-blocking comments

danielguillan (approved):

apps/docs/content/components/Box/index.mdx
rezrah: 👍 I can remove it. Feels quite text heavy though, any other way we can help folks quickly visualise the usefulness of this component?

packages/react/src/Box/Box.tsx
rezrah: `ul` can significantly change the semantic behaviour, and also encourages a dependence on Box that I'm trying to avoid (it's meant for situational one-off styles). If we're trying to suggest that cards and pillars need this functionality, perhaps we need a CardGroup or PillarGroup instead to make this more obvious?

I do think adding section and article landmarks could be useful, but also feel we're bordering into potential accessibility violation waters when we weren't originally intending to. Can we see how Box is being used before committing to this suggestion @josepmartins @danielguillan?

apps/docs/content/components/Box/index.mdx
josepmartins: I could think of adding real use case examples, as I did for [Stack](https://primer.style/brand/components/Stack/). Maybe on a follow up PR? just to unblock this PR release

packages/react/src/Box/Box.tsx
josepmartins: I agree 👍  let's see how Box is used and consider this later

PR #357, opened 2023-07-21
Bump doctocat from 4.4.2 to 4.6.1 in /apps/docs

josepmartins:
Bumps [doctocat](https://github.com/primer/doctocat) from 4.4.2 to 4.6.1


rezrah (approved):
👍 Looks nice! Thanks @josepmartins ✨ 

General theme feedback: It would be nice if the sites content in the off-canvas menu wasn't pushed so far down. The Rails, PRC, etc links aren't as important for Primer Brand as the site content itself, and could go at the bottom imo.

Screenshot 2023-07-21 at 10 44 49

Wonder if we can change that in a future doctocat release?

josepmartins:
> General theme feedback: It would be nice if the sites content in the off-canvas menu wasn't pushed so far down. The Rails, PRC, etc links aren't as important for Primer Brand as the site content itself, and could go at the bottom imo.

That makes a lot of sense, I also see some styles that needs to be updated in that panel (monospace font and spacing)

I'll try to adjust them in the next Doctocat version 👍

emilybrick:
> General theme feedback: It would be nice if the sites content in the off-canvas menu wasn't pushed so far down. The Rails, PRC, etc links aren't as important for Primer Brand as the site content itself, and could go at the bottom imo.

If there's a way to do this while still communicating the clear hierarchy of the top-level navigation to in-page/contextual navigation, I'm all for it. Just want to make sure it's clear to users that the Primer, Brand, and About links will take them completely out of the current screen/view and into a new experience, whereas the contextual nav is within the boundaries of Primer Brand

emilybrick:
Quick notes here!

- If we're able to use UnderlineNav to show the selected states, the`/brand` part from the Primer Design System header link should be removed
Screenshot 2023-07-21 at 10 25 56 AM

- Brand should be underlined in this example, I think
Screenshot 2023-07-21 at 10 30 36 AM

- Search text looks like it increased 👀 
Screenshot 2023-07-21 at 10 30 25 AM


- It feels a ta
Screenshot 2023-07-21 at 10 33 27 AM
d more balanced when all the type sizes in the top header are consistent. wdyt?

PR #358, opened 2023-07-21
FAQ a11y sign off updates

JoshBowdenConcepts:
## Summary

A number of a11y changes to ensure full accessibility for the FAQ component.

## List of notable changes:

- **added** FAQ.Group to the FAQ component to provide accessible markup
- **updated** Accordion summary spacing so that the focus would not obscure the text

- Making sure each of the items within [this issue](https://github.com/github/primer/issues/1675) are addressed.

## Steps to test:

1. Open the FAQ component in CI-deployed preview environment
2. Go to FAQ story in Storybook
3. Verify that # behaves as described in the [following issue](https://github.com/github/primer/issues/1675).

- closes https://github.com/github/primer/issues/1675

## Contributor checklist:

## Reviewer checklist:

## Screenshots:

> The spacing here between the heading and content is increased to allow for the focus border to not block the details text

<table>
<tr>
</tr>
<tr>
<td valign="top">

image

 </td>
<td valign="top">

image

</td>
</tr>
</table>


packages/react/src/Accordion/Accordion.module.css
```diff
@@ -107,7 +107,7 @@ details[open] .Accordion__summary--expanded::after {
 details[open] > .Accordion__content {
   padding-left: var(--base-size-40);
   padding-bottom: var(--base-size-24);
-  margin-top: calc(var(--base-size-16) * -1); /* for 8px gap between question and answer */
```
JoshBowdenConcepts, line 110: > (sev 2) When the accordion controls are expanded using the keyboard, the focus indicator overlaps the first line of text in the panel decreasing readability.

This needed to be removed to support this a11y issue

packages/react/src/FAQ/FAQ.tsx
```diff
@@ -104,14 +113,37 @@ function FAQSubheading({children, className, as = 'h3', ...rest}: FAQSubheadingP
...
 }
 
+type FAQGroupProps = BaseProps<HTMLUListElement> & {
+  children: React.ReactNode | React.ReactNode[]
+  'aria-labelledby': string
+}
+
+function FAQGroup({className, children}: FAQGroupProps) {
+  return (
+    // eslint-disable-next-line jsx-a11y/no-redundant-roles
```
JoshBowdenConcepts, line 123: > sev 2) For all examples, the accordion controls are not using list semantics. Screen reader users will not be made aware of the grouped nature of the controls, nor how many items are in the collection.

> Solution: Use an unordered list ul to wrap the related accordion controls, with an li wrapping each individual item. Add an aria-labelledby attribute to the ul, and reference the ID of the section heading. If list-style: none is applied with CSS, be sure to add role="list" to the ul element to ensure that all screen readers report the list structure. Some screen reader/browser combinations will remove list semantics if no bullets are present, and the explicit role declaration restores them.

This needed to be added to follow this guidance. Maybe the linter can be updated so we can remove this in the future.

packages/react/src/FAQ/FAQ.features.stories.tsx
```diff
@@ -120,8 +120,8 @@ export const AllOpen: StoryFn<typeof FAQ> = () => {
   return (
     <Container>
       <FAQ>
-        <FAQ.Heading>Frequently asked&nbsp;questions</FAQ.Heading>
-        <>
+        <FAQ.Heading id="all-open-faq">Frequently asked&nbsp;questions</FAQ.Heading>
```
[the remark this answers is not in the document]
JoshBowdenConcepts: How would you feel about making these props required and leaving information in the documentation? We could do this dynamically but it would require a second loop of the children to first grab the id if a user specifies one and then set the id on the group. Of course, if they didn't supply one then this second loop wouldn't be needed but I was thinking we would want to support them passing their own id if they wanted to access that component using an id query.

PR #366, opened 2023-08-01
Add optional image to Hero

rezrah:
## Summary

Part of https://github.com/github/primer/issues/2498

Adds optional foreground image support the Hero component. 

🔗  [Storybook examples](https://primer-2a9327fdf7-26139705.drafts.github.io/storybook/?path=/story/components-hero-features--with-image-block-end-center)
🔗  [React docs](https://primer-2a9327fdf7-26139705.drafts.github.io/components/Hero/react#images)

## List of notable changes:

- **added** 2 image positioning variants: `inline-end` and `block-end` 
- **updated** tests to preserve 100% coverage

- Reviewing design decisions (no figma specs available at this time)
- API review
- Spotting any bugs in the implementation

## Contributor checklist:

## Reviewer checklist:

## Screenshots:

https://github.com/primer/brand/assets/13340707/b719342c-1513-428b-89c6-54e20531e867


apps/docs/content/components/Hero/react.mdx
```diff
@@ -43,6 +43,37 @@ import {Hero} from '@primer/react-brand'
...
+  </Hero>
+  <br />
+  <Hero align="center">
+    <Hero.Label>Actions</Hero.Label>
+    <Hero.Heading size="2">
+      Automate your workflow from idea to production
+    </Hero.Heading>
+    <Hero.PrimaryAction href="#">Get started with Actions</Hero.PrimaryAction>
+    <Hero.Image
+      imagePosition="inline-end"
```
josepmartins, line 69: shouldn't we add a prop to the hero parent to manage this instead? It feels strange that a subitem prop changes the layout of the component

packages/react/src/Hero/Hero.module.css
```diff
@@ -4,12 +4,40 @@
...
+  grid-template-columns: 1fr;
+  gap: var(--base-size-40);
+}
+
+@media screen and (min-width: 48rem) {
+  .Hero--image-pos-block-end {
+    gap: var(--base-size-80);
+  }
+}
+
```
josepmartins, line 33: for consistency across other elements in the page, shouldn't we use the Grid component instead?

apps/docs/content/components/Hero/react.mdx
rezrah: It's because the image is optional and `imagePosition` only makes sense if you're passing a child called `Hero.Image`, hence the coupling. 

I agree it's a bit of indirection in the parent, but the user won't see that. The trade-off I'm thinking about here is that the user wants to insert an image, so the configurability options for that _image_ is co-located to the child. It's a devexp feature more than anything.

Does that help contextualise it, or do you feel strongly about it moving into the parent?

packages/react/src/Hero/Hero.module.css
rezrah: Fair point. I can look into this but the spacing in the site refactor designs doesn't quite align with what we have in the grid today.

josepmartins: Looking at the design files it seems they've been using the Grid layout so spacing/gaps should be similar 🤔 

Screenshot 2023-08-02 at 15 23 50

apps/docs/content/components/Hero/react.mdx
josepmartins: thanks for the context, not strong thoughts on this.

packages/react/src/Hero/Hero.module.css
rezrah: It's the row gap that's the issue, not column...

Screenshot 2023-08-02 at 14 35 49

I've switched to Grid internally, but it needed various workarounds... you can check it out [here](https://github.com/primer/brand/pull/366/commits/cba3cd691b07306ef494671cdafc1a7fc927de0f).

josepmartins: nice! I see what you mean with the workarounds.. Would it make sense to use the `Box` and marginBlockStart options to manage this gap instead of the grid row gap values?

rezrah: We could do 🤔 .. I think I prefer the Grid approach you suggested before personally and switching out again feels redundant at this point. Given we want to use Box sparingly as maintainers, using the locally scoped Grid token here feels the the paved path approach to fixing this. Does that work for you too @josepmartins?

packages/react/src/Hero/Hero.tsx
```diff
@@ -15,36 +19,63 @@ export type HeroProps = BaseProps<HTMLElement> & {
...
       >
-        {HeroChildren}
-        <div className={styles['Hero-actions']}>{HeroActions}</div>
+        <Grid fullWidth className={clsx(styles['Hero-grid'], styles[`Hero-grid--${imagePosition}`])}>
+          <Grid.Column span={{medium: HeroImageChild && imagePosition === 'inline-end' ? 6 : 12}}>
+            <Stack
+              direction="vertical"
+              gap="none"
+              padding="none"
+              alignItems={imagePosition === 'inline-end' ? undefined : align === 'start' ? 'flex-start' : 'center'}
```
danielguillan, line 68: For `inline-end` we also need to set `align-items` to `flex-start`, otherwise it will default to `stretch` and make the Label span the full width of the Stack container.

screenshot-dDkJ87H4-000602@2x

packages/react/src/Hero/Hero.module.css
josepmartins: It works! Sorry I wasn't clear, I mentioned using Box (along with the Grid) to not modify the CSS properties of the grid row gap. But locally scoped grid tokens works too oc 👍

packages/react/src/Hero/Hero.tsx
rezrah: 👍 fixed

packages/react/src/Grid/Grid.module.css
```diff
@@ -8,6 +8,11 @@
   margin: 0 auto;
 }
 
+.Grid--full-width {
+  width: 100%;
+  max-width: 100%;
```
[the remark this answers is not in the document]
rezrah: 👍 thanks, fixed

rezrah:
> Nitpick: should the prop be simply called position? As it's applied to the Image already, imagePosition feels redundant.

Agreed. Updated to `position`

danielguillan (approved):
Looks great! ✨

josepmartins (approved):
lgtm!

PR #371, opened 2023-08-03
[Draft] Site refactor: typography (only change values)

langermank:
## Summary

This approach to updating Brand's type scale changes the values of the existing design tokens to better align with the overall design direction. 

The new scale relies on a more semantic naming convention and removes the plug-and-play nature of the previous numbering system. Before you could mix and match, and now each step in the scale has its own set of constraints for all of the values that make up a heading/text.

**New scale:**

```
brand-text-display-size-large
brand-text-display-size-medium
brand-text-heading-size-xlarge
brand-text-heading-size-large
brand-text-heading-size-medium
brand-text-heading-size-small
brand-text-heading-size-xsmall
brand-text-heading-size-xxsmall
brand-text-body-size-subhead
brand-text-body-size-medium (default)
brand-text-body-size-small
brand-text-body-size-xsmall 
```

As a first step, we can change the values of the existing scale to make sweeping visual changes to all Brand components. Later we'll need to introduce breaking changes to the Heading and Text component to fully align with the new scale.

**Changes needed later:**
- Add new design tokens using the correct naming convention
- Remove the numbering scale from `Heading` and offer semantic names
- Only allow for one alternative weight (medium) to be used for Headings
- `Text` should only support the body size options

### Consequences 

This PR has some limitations for how close we can get to the new scale without making a breaking code change.

- brand-text-heading-size-medium and brand-text-body-size-subhead have no equivalent in the existing scale and therefor are excluded
- This excludes the second option for `Heading` which uses `medium` for weight and has some more minor differences in line-height 
- Making the `Text` default body size 300 instead of 400 could lead to some confusion or inconsistency outside of the Brand repo

## Contributor checklist:

## Reviewer checklist:


rezrah:
@langermank - this is great. Some thoughts around the 'consequences'.

> Add new design tokens using the correct naming convention
I think we should go ahead and rename these now, but documenting the migration steps in a changelog as we go. Here's an [example of how we could do it](https://github.com/primer/brand/blob/rezrah/add-emphasis-support/.changeset/fair-rocks-return.md), from the [other typography PR](https://github.com/primer/brand/pull/376).

> Remove the numbering scale from Heading and offer semantic names
👍 Let's go ahead and do that..

> Only allow for one alternative weight (medium) to be used for Headings

Could we keep these please? These are escape-hatches that we added intentionally for edge-cases. The semantic names should take care of the defaults though.

> Text should only support the body size options
By this, do you mean it won't support variant like it currently does?

PR #373, opened 2023-08-03
VideoPlayer Component

JoshBowdenConcepts:
## Summary

A VideoPlayer component that will be used to render hosted video content.

- https://github.com/github/primer/issues/2448

## List of notable changes:

- **added** VideoPlayer component

- Making sure the VideoPlayer performs at all screen sizes as expected.
- Stress test the controls to make sure they are performant and don't glitch

## Steps to test:

1. Open the VideoPlayer component in CI-deployed preview environment
2. Go to VideoPlayer story in Storybook
3. Verify that the VideoPlayer behaves as described in the following issue.

- [API Exploration](https://www.figma.com/file/y0OnbQ7Zpku6Y1Bl7IHvkZ/VideoPlayer---API-exploration?type=whiteboard&node-id=0-1&t=FlsKh5FUvY1YMULQ-0)
- [Tracking Issue](https://github.com/github/primer/issues/2448)
- [Octicon Play Pause Issue](https://github.com/github/primer/issues/2572)

## Contributor checklist:

## Reviewer checklist:

## Screenshots:

image


.gitignore
```diff
@@ -15,6 +15,9 @@ playwright-report/
 .DS_Store
 *.pem
 
+# Tarballs
+*.tgz
```
JoshBowdenConcepts, line 19: I added this for testing but I think it is fine for this to be added. If anything it adds protection.

apps/docs/scripts/components-with-animation.js
```diff
@@ -1,16 +1,16 @@
...
 export const supportedComponents = [
-  'Button',
   'Box',
+  'Button',
   'ComparisonTable',
   'FAQ',
   'Heading',
   'Image',
   'Label',
-  'River',
```
JoshBowdenConcepts, line 11: This happened automatically on build so I assume this is ok to push otherwise we will have to keep reverting it.

packages/react/src/VideoPlayer/VideoPlayer.tsx
```diff
@@ -0,0 +1,180 @@
...
+  return (
+    <div
+      className={clsx(
+        styles.VideoPlayer__container,
+        styles.VideoPlayer__overlays,
+        !playing && styles.VideoPlayer__showOverlays,
+      )}
+      ref={videoWrapperRef}
+    >
+      {/* eslint-disable-next-line jsx-a11y/media-has-caption */}
```
JoshBowdenConcepts, line 100: This is needed because we can't inspect the children but the idea is that it is required and will be added on implementation.

packages/react/src/VideoPlayer/VideoPlayer.test.tsx
```diff
@@ -0,0 +1,100 @@
+import React from 'react'
+import {render, cleanup, fireEvent} from '@testing-library/react'
+import '@testing-library/jest-dom'
+
+import {VideoPlayer} from '.'
+
```
JoshBowdenConcepts, line 6: I think more tests could be included here but I haven't been able to get specific state values to update. If anyone has some ideas for improvements please let me know.

.changeset/tame-keys-provide.md
```diff
@@ -0,0 +1,18 @@
+---
+'@primer/react-brand': minor
+---
+
+Use VideoPlayer component to render hosted videos
+
+```jsx
+<VideoPlayer title="HLS On Demand Demo" crossOrigin="anonymous">
```
rezrah, line 8: As `crossOrigin` isn't indicated as a required prop, it shouldn't need to be mentioned in the usage example.

```suggestion
<VideoPlayer title="HLS On Demand Demo">
```

.changeset/tame-keys-provide.md
```diff
@@ -0,0 +1,18 @@
...
+  <VideoPlayer.Source src="https://cdn.api.video/vod/vi36GGvd6PoTqViQLxBWwHjJ/mp4/1080/source.mp4" />
+  <VideoPlayer.Track
+    src="https://raw.githubusercontent.com/JoshBowdenConcepts/Subtitles-Example/main/sample.vtt"
```
rezrah, line 11: This is linking to your personal account, could you replace it with a placeholder / example please? (doesn't need to be a real url, relative is fine).

.changeset/tame-keys-provide.md
```diff
@@ -0,0 +1,18 @@
...
+    default
+    kind="subtitles"
+    srcLang="en"
+    label="English"
+  />
```
rezrah, line 16: I think only `kind` and `src` are required for `<track>`. Could we either remove these other ones or make them required props on VideoPlayer.Track if they are necessary?

apps/docs/content/components/VideoPlayer.mdx
```diff
@@ -0,0 +1,113 @@
...
+```js
+import {VideoPlayer} from '@primer/react-brand'
+```
+
+## Examples
+
+### Default
+
+```jsx live
+<VideoPlayer title="HLS On Demand Demo" crossOrigin="anonymous">
```
rezrah, line 20: `crossOrigin` isn't a required attribute, and I don't think we should be opinionated about the value `anonymous` as end-user CORS expectations could differ. 

```suggestion
<VideoPlayer title="HLS On Demand Demo">
```

apps/docs/content/components/VideoPlayer.mdx
```diff
@@ -0,0 +1,113 @@
...
+  <VideoPlayer.Source src="https://cdn.api.video/vod/vi36GGvd6PoTqViQLxBWwHjJ/mp4/1080/source.mp4" />
+  <VideoPlayer.Track
+    src="https://raw.githubusercontent.com/JoshBowdenConcepts/Subtitles-Example/main/sample.vtt"
```
rezrah, line 23: This is pointing at your personal account right now, could we use a GitHub/Primer-hosted version instead please?

apps/docs/content/components/VideoPlayer.mdx
```diff
@@ -0,0 +1,113 @@
...
+import {VideoPlayer} from '@primer/react-brand'
+```
+
+## Examples
+
+### Default
+
+```jsx live
+<VideoPlayer title="HLS On Demand Demo" crossOrigin="anonymous">
+  <VideoPlayer.Source src="https://cdn.api.video/vod/vi36GGvd6PoTqViQLxBWwHjJ/mp4/1080/source.mp4" />
```
rezrah, line 21: Do you know who owns this placeholder video? I think we should try to source our own. Brand Studio may have something that we can use.

apps/docs/content/components/VideoPlayer.mdx
rezrah: Same also applies for all other instances 🙏

apps/docs/content/components/VideoPlayer.mdx
JoshBowdenConcepts: This was specifically to make the demo work. Ideally, I will have some demo fixtures uploaded to our public folder so we can use those without this attribute.

apps/docs/content/components/VideoPlayer.mdx
```diff
@@ -0,0 +1,113 @@
...
+</VideoPlayer>
+```
+
+## Without branding
+
+```jsx live
+<VideoPlayer
+  branding={false}
+  title="HLS On Demand Demo"
+  crossOrigin="anonymous"
```
rezrah, line 57:
```suggestion
```

apps/docs/content/components/VideoPlayer.mdx
```diff
@@ -0,0 +1,113 @@
...
+## Component props
+
+<h3>
+  VideoPlayer <Label>Required</Label>
+</h3>
+
+`VideoPlayer` provides a React alternative to the native HTML `<video />`.
+
+The component API supports all standard HTML attribute props, while providing some additional behavior as described below.
+
```
rezrah, line 79: For consistency with other docs pages, could you please drop these details to underneath its relevant prop table?

apps/docs/content/components/VideoPlayer.mdx
```diff
@@ -0,0 +1,113 @@
...
+description: Use the video player component to present videos availible via CDN or local.
+---
+
+import {Label} from '@primer/react'
+
+```js
+import {VideoPlayer} from '@primer/react-brand'
+```
+
+## Examples
```
rezrah, line 15: FYI there seems to be a strange hover / overflow related bug in the docs examples 👇 


https://github.com/primer/brand/assets/13340707/cab69536-23fd-48d5-9c61-3b43e9797f33

apps/docs/content/components/VideoPlayer.mdx
```diff
@@ -0,0 +1,113 @@
...
+  VideoPlayer <Label>Required</Label>
+</h3>
+
+`VideoPlayer` provides a React alternative to the native HTML `<video />`.
+
+The component API supports all standard HTML attribute props, while providing some additional behavior as described below.
+
+| Name        | Type                                   | Default | Required | Description                                                          |
+| :---------- | :------------------------------------- | :-----: | :------: | :------------------------------------------------------------------- |
+| `title`     | `boolean`                              |         |  `true`  | Sets the title of the video                                          |
```
rezrah, line 82: Should this be a `string` type instead of boolean?

```suggestion
| `title`     | `string`                              |         |  `true`  | Sets the title of the video                                          |
```

apps/docs/content/components/VideoPlayer.mdx
```diff
@@ -0,0 +1,113 @@
...
+| `children`  | `React.ReactNode`, `React.ReactNode[]` |         |  `true`  | Takes `<source>` and `<track>` elements to present video information |
```
rezrah, line 83: nit: I don't think <track> is required, so may need to call that out in the description.

packages/design-tokens/src/tokens/functional/components/video-player/base.json
```diff
@@ -0,0 +1,129 @@
...
+  "brand": {
+    "VideoPlayer": {
+      "border": {
+        "radius": {
+          "value": "var(--brand-borderRadius-medium)"
+        }
+      },
+      "controls": {
+        "gap": {
+          "value": "15px"
```
rezrah, line 11: Does this have to be 15px? Could we use rems, which scale better? If so, can we switch to use..

```suggestion
          "value": "var(--base-size-16)"
```

packages/design-tokens/src/tokens/functional/components/video-player/base.json
```diff
@@ -0,0 +1,129 @@
+{
+  "brand": {
+    "VideoPlayer": {
+      "border": {
+        "radius": {
+          "value": "var(--brand-borderRadius-medium)"
+        }
+      },
```
rezrah, line 8: Does this need to be a variable / should it be user customisable?

If it does, could we change the token to be `--brand-VideoPlayer-borderRadius` please? this will currently build as `--brand-VideoPlayer-border-radiud` (not camel-case).

packages/design-tokens/src/tokens/functional/components/video-player/base.json
```diff
@@ -0,0 +1,129 @@
...
+      "closedCaption": {
+        "padding": {
+          "vertical": {
+            "value": "1px"
+          },
+          "horizontal": {
+            "value": "5px"
+          }
```
rezrah, line 43: Do we need tokens for this? Feels like something we wouldn't want users to customize, so could we hard-code these into the stylesheet instead?

packages/design-tokens/src/tokens/functional/components/video-player/base.json
```diff
@@ -0,0 +1,129 @@
...
+        },
+        "icon": {
+          "height": {
+            "value": "24px"
+          }
+        },
+        "focus": {
+          "borderColor": {
+            "value": "var(--brand-color-focus)"
+          },
```
rezrah, line 30: I think you can drop this token and just use `var(--brand-color-focus)` in the stylesheet.

packages/design-tokens/src/tokens/functional/components/video-player/base.json
```diff
@@ -0,0 +1,129 @@
...
+          "horizontal": {
+            "value": "5px"
+          }
+        },
+        "backgroundColor": {
+          "value": "var(--brand-color-white)"
+        },
+        "color": {
+          "value": "black"
+        },
```
rezrah, line 50: We don't use named colors, and prefer tokens. Could we please:

a) use our base color for black instead `var(--base-color-scale-black-0)`
b) move colors into their own JSON file, and provide the dark color mode value also (should be able to reference other color source files for examples of how we do this)

packages/design-tokens/src/tokens/functional/components/video-player/base.json
```diff
@@ -0,0 +1,129 @@
...
+        "padding": {
+          "vertical": {
+            "value": "1px"
+          },
+          "horizontal": {
+            "value": "5px"
+          }
+        },
+        "backgroundColor": {
+          "value": "var(--brand-color-white)"
```
rezrah, line 46: Is `var(--brand-color-white)` a valid token? Not seen that one before. 👀

packages/react/src/VideoPlayer/VideoPlayer.tsx
```diff
@@ -0,0 +1,189 @@
...
+          onPlay && onPlay(e)
+        }}
+        onPause={e => {
+          setPlaying(false)
+          onPause && onPause(e)
+        }}
+        onPlaying={e => {
+          videoRef.current?.textTracks[0].cues && setTrackInformation(videoRef.current.textTracks[0].cues)
+          onPlaying && onPlaying(e)
+        }}
```
rezrah, line 119: Performance optimisation: Could you hoist this function to outside of the render function, so that we're not recreating it each time. You can wrap it inside a `useCallback` to memoize it after. Example from [another component](https://github.com/primer/brand/blob/551c47cf635454a5eb82b3cca1ef833e2f53c1b4/packages/react/src/ActionMenu/ActionMenu.tsx#L124).

rezrah: Could we also please apply this optimisation to all other inline function declarations?

apps/docs/content/components/VideoPlayer.mdx
JoshBowdenConcepts: Sure! It was before the prop table for Checkbox which I grabbed this language from should I update Checkbox to reflect this change as well?

apps/docs/content/components/VideoPlayer.mdx
JoshBowdenConcepts: I make this property required because it will be required for accessibility. Are you saying we should update it to not be required or is there another way you are looking for me to represent this?

packages/react/src/VideoPlayer/VideoPlayer.tsx
JoshBowdenConcepts: This has been updated in the following [SHA](https://github.com/primer/brand/pull/373/commits/6f9a9c8740844ac6d50be622f1c9aab106c564dc)

packages/design-tokens/src/tokens/functional/components/video-player/base.json
JoshBowdenConcepts: This has been updated in the following [SHA](https://github.com/primer/brand/pull/373/commits/6f9a9c8740844ac6d50be622f1c9aab106c564dc)

packages/design-tokens/src/tokens/functional/components/video-player/base.json
JoshBowdenConcepts: This has been updated in the following [SHA](https://github.com/primer/brand/pull/373/commits/6f9a9c8740844ac6d50be622f1c9aab106c564dc)

packages/design-tokens/src/tokens/functional/components/video-player/base.json
JoshBowdenConcepts: This has been updated in the following [SHA](https://github.com/primer/brand/pull/373/commits/6f9a9c8740844ac6d50be622f1c9aab106c564dc)

packages/design-tokens/src/tokens/functional/components/video-player/base.json
JoshBowdenConcepts: This has been updated to remove the token in the following [SHA](https://github.com/primer/brand/pull/373/commits/6f9a9c8740844ac6d50be622f1c9aab106c564dc)

packages/design-tokens/src/tokens/functional/components/video-player/base.json
```diff
@@ -0,0 +1,129 @@
...
+        "radius": {
+          "value": "var(--brand-borderRadius-medium)"
+        }
+      },
+      "controls": {
+        "gap": {
+          "value": "15px"
+        },
+        "background": {
+          "value": "rgba(0, 0, 0, 0.2)"
```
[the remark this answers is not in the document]
JoshBowdenConcepts: This has been updated in the following [SHA](https://github.com/primer/brand/pull/373/commits/6f9a9c8740844ac6d50be622f1c9aab106c564dc)

packages/design-tokens/src/tokens/functional/components/video-player/base.json
JoshBowdenConcepts: This has been updated in the following [SHA](https://github.com/primer/brand/pull/373/commits/6f9a9c8740844ac6d50be622f1c9aab106c564dc)

apps/docs/content/components/VideoPlayer.mdx
JoshBowdenConcepts: This has been updated in the following [SHA](https://github.com/primer/brand/pull/373/commits/6f9a9c8740844ac6d50be622f1c9aab106c564dc)

apps/docs/content/components/VideoPlayer.mdx
JoshBowdenConcepts: This has been updated in the following [SHA](https://github.com/primer/brand/pull/373/commits/6f9a9c8740844ac6d50be622f1c9aab106c564dc)

apps/docs/content/components/VideoPlayer.mdx
```diff
@@ -0,0 +1,113 @@
...
+</VideoPlayer>
+```
+
+## With poster
+
+```jsx live
+<VideoPlayer
+  poster="https://github.githubassets.com/images/icons/media-player/poster.jpg"
+  title="HLS On Demand Demo"
+  crossOrigin="anonymous"
```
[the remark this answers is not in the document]
JoshBowdenConcepts: This has been updated in the following [SHA](https://github.com/primer/brand/pull/373/commits/6f9a9c8740844ac6d50be622f1c9aab106c564dc)

apps/docs/content/components/VideoPlayer.mdx
JoshBowdenConcepts: This has been updated in the following [SHA](https://github.com/primer/brand/pull/373/commits/6f9a9c8740844ac6d50be622f1c9aab106c564dc)

apps/docs/content/components/VideoPlayer.mdx
JoshBowdenConcepts: This has been updated in the following [SHA](https://github.com/primer/brand/pull/373/commits/6f9a9c8740844ac6d50be622f1c9aab106c564dc)

.changeset/tame-keys-provide.md
JoshBowdenConcepts: This has been updated in the following [SHA](https://github.com/primer/brand/pull/373/commits/6f9a9c8740844ac6d50be622f1c9aab106c564dc)

apps/docs/content/components/VideoPlayer.mdx
JoshBowdenConcepts: This has been updated in the following [SHA](https://github.com/primer/brand/pull/373/commits/a15991caf8ad89311382ef5e27a02f6acfcc13c7)

.changeset/tame-keys-provide.md
JoshBowdenConcepts: This has been updated in the following [SHA](https://github.com/primer/brand/pull/373/commits/a15991caf8ad89311382ef5e27a02f6acfcc13c7)

.changeset/tame-keys-provide.md
JoshBowdenConcepts: This has been updated in the following [SHA](https://github.com/primer/brand/pull/373/commits/05ec5df1318a6dfeb932bbc4b8e240cf0fd0e3bf)

JoshBowdenConcepts: Kind was not required so I removed that as well.

apps/docs/content/components/VideoPlayer.mdx
```diff
@@ -0,0 +1,90 @@
...
+
+<h3>VideoPlayer.Track</h3>
+
+`VideoPlayer.Track` provides a React alternative to the native HTML `<track />`.
+
+The component API supports all standard HTML attribute props, while providing some additional behavior as described below.
+
+| Name      | Type                                                            | Default | Required | Description                                                                                                                              |
+| :-------- | :-------------------------------------------------------------- | :-----: | :------: | :--------------------------------------------------------------------------------------------------------------------------------------- |
+| `src`     | `boolean`                                                       |         |  `true`  | Sets the source of the VTT file containing captions                                                                                      |
```
danielguillan, line 86:
```suggestion
| `src`     | `string`                                                       |         |  `true`  | Sets the source of the VTT file containing captions                                                                                      |
```

.changeset/tame-keys-provide.md
rezrah: Checking in on this ☝️, it seems to be the same as before FYI. Do you need to push up the changes?

apps/docs/content/components/VideoPlayer.mdx
```diff
@@ -0,0 +1,90 @@
...
+```js
+import {VideoPlayer} from '@primer/react-brand'
+```
+
+## Examples
+
+### Default
+
+```jsx live
+<VideoPlayer title="HLS On Demand Demo">
```
rezrah, line 20: Thoughts on changing this title to be more generic?

```suggestion
title="GitHub media player"
```

apps/docs/static/example.mov
rezrah, line 1: This seems to be a personal video, could we update this please to use some GitHub stock video? Perhaps you can ask BaM  / Brand Studio to see if they have anything suitable. 

Currently appearing like this in docs 👇 

Screenshot 2023-08-10 at 14 40 31

apps/docs/static/example.vtt
```diff
@@ -0,0 +1,9 @@
+WEBVTT
+00:00:00.500 --> 00:00:10.000
+Hey there! 🍉 This is a sample subtitle. 👈
+00:00:10.500 --> 00:00:20.300
+This sub title is a bit longer and demonstrates the ability to render longer content onto multiple lines. It is good to have examples of this where there is a lot of content.
```
rezrah, line 5:
```suggestion
This subtitle is a bit longer and demonstrates how to present longer content onto multiple lines. It is good to have examples of this where there is a lot of content.
```

apps/docs/static/example.vtt
```diff
@@ -0,0 +1,9 @@
...
+00:00:21.000 --> 00:00:30.000
+We can continue to test how these subtitles might look
+00:00:30.500 --> 00:01:30.000
+This last subtitle will last until one minute and thirty seconds have passed in the video
```
rezrah, line 9: Presumably we'll need to adjust this depending on the final video file used here?

packages/design-tokens/src/tokens/functional/components/video-player/base.json
```diff
@@ -0,0 +1,81 @@
...
+    "VideoPlayer": {
+      "controls": {
+        "gap": {
+          "value": "var(--base-size-16)"
+        },
+        "background": {
+          "value": "hsl(var(--base-color-scale-black-0-hsl) / 80%)"
+        },
+        "backgroundGradient": {
+          "value": "linear-gradient(rgba(0, 0, 0, 0), hsl(var(--base-color-scale-black-0-hsl) / 40%))"
```
rezrah, line 12:
```suggestion
          "value": "linear-gradient(var(--base-color-scale-black-0), hsl(var(--base-color-scale-black-0-hsl) / 40%))"
```

apps/docs/content/components/VideoPlayer.mdx
```diff
@@ -0,0 +1,90 @@
+---
+title: VideoPlayer
```
rezrah, line 2:
```suggestion
title: Video player
```

rezrah: ICYMI all docs page headers now use sentence case

apps/docs/content/components/VideoPlayer.mdx
```diff
@@ -0,0 +1,90 @@
...
+status: Experimental
+source: https://github.com/primer/brand/blob/main/packages/react/src/VideoPlayer/VideoPlayer.tsx
+storybook: '/brand/storybook/?path=/story/components-videoplayer--playground'
+description: Use the video player component to present videos availible via CDN or local.
```
rezrah, line 6: ending at local feels a bit abrupt, how about this suggestion for conciseness...

```suggestion
description: Use the video player component to present self-hosted videos available via CDN or local.
```

packages/design-tokens/src/tokens/functional/components/video-player/base.json
```diff
@@ -0,0 +1,81 @@
...
+        },
+        "padding": {
+          "value": "var(--base-size-16) var(--base-size-24)"
+        },
+        "icon": {
+          "height": {
+            "value": "24px"
+          }
+        },
```
rezrah, line 21: I don't think this is being used? Can we remove this please? I don't think the icon size needs to be configurable.

packages/react/src/VideoPlayer/VideoPlayer.tsx
```diff
@@ -0,0 +1,194 @@
...
+          onLoadedMetadata && onLoadedMetadata(e)
+        }}
+        onTimeUpdate={() => setPlayedTime(videoRef.current?.currentTime || 0)}
+      >
+        {children}
+      </video>
+      {title && (
+        <div className={styles.VideoPlayer__title}>
+          {branding && (
+            <svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 16 16" width="48" height="48">
```
rezrah, line 137: Is this is the invertocat logo? If it is, could you replace this with the version from Octicons please? We shouldn't need to hardcode these, unless you're applying some customisations to it? Will help reduce footprint of this file too.

apps/docs/static/example.vtt
JoshBowdenConcepts: I was thinking this could remain the same as the goal is to show both large and small subtitles. Unless you think when this comes down to an a11y review that we will want to accurately represent the video.

packages/react/src/VideoPlayer/VideoPlayer.tsx
JoshBowdenConcepts: This is updated in this [SHA](https://github.com/primer/brand/pull/373/commits/c4e331d5b4f9967a0295bc43a82782ec6f61a88f)

JoshBowdenConcepts:
**Decision:** In a discussion with @rezrah we have removed the color modes as this will always remain the same to provide the best viewing on top of a dynamic background like a movie.

apps/docs/static/example.mov
JoshBowdenConcepts: This has been updated! And it was amazing 🌴

apps/docs/content/components/VideoPlayer.mdx
rezrah: Hiding the overflow seems to clip the tooltip. Is there any other workaround? Do we know why that was occurring on hover to begin with? 

Screenshot 2023-08-11 at 17 13 36

apps/docs/content/components/VideoPlayer.mdx
```diff
@@ -0,0 +1,87 @@
...
+<VideoPlayer title="GitHub media player">
+  <VideoPlayer.Source src="/example.mp4" />
+  <VideoPlayer.Track src="/example.vtt" />
+</VideoPlayer>
+```
+
+## With poster
+
+```jsx live
+<VideoPlayer poster="/example-poster.png" title="GitHub Video Demo">
```
rezrah, line 29: Update to title for consistency with previous example

```suggestion
<VideoPlayer poster="/example-poster.png" title="GitHub media player">
```

apps/docs/content/components/VideoPlayer.mdx
```diff
@@ -0,0 +1,87 @@
...
+  <VideoPlayer.Source src="/example.mp4" />
+  <VideoPlayer.Track src="/example.vtt" />
+</VideoPlayer>
+```
+
+## Without branding
+
+```jsx live
+<VideoPlayer branding={false} title="GitHub Video Demo">
```
rezrah, line 38: Update to title for consistency with previous example

```suggestion
<VideoPlayer branding={false} title="GitHub media player">
```

apps/docs/content/components/VideoPlayer.mdx
rezrah: Checkbox will eventually need to be updated, but can be done separately. They should go underneath the main table. Examples: [SectionIntro](https://primer-bb0111d97b-26139705.drafts.github.io/components/SectionIntro/react#component-props), [Card](https://primer-bb0111d97b-26139705.drafts.github.io/components/Card/react#cardheading), [ComparisonTable](https://primer-bb0111d97b-26139705.drafts.github.io/components/ComparisonTable#comparisontablefootnote), etc

apps/docs/content/components/VideoPlayer.mdx
```diff
@@ -0,0 +1,87 @@
...
+
+`VideoPlayer.Track` provides a React alternative to the native HTML `<track />`.
+
+The component API supports all standard HTML attribute props, while providing some additional behavior as described below.
+
+| Name      | Type                                                            | Default | Required | Description                                                                                                                              |
+| :-------- | :-------------------------------------------------------------- | :-----: | :------: | :--------------------------------------------------------------------------------------------------------------------------------------- |
+| `src`     | `string`                                                        |         |  `true`  | Sets the source of the VTT file containing captions                                                                                      |
+| `kind`    | `subtitles`, `captions`, `descriptions`, `chapters`, `metadata` |         | `false`  | Sets how the text track is meant to be used                                                                                              |
```
rezrah, line 84: These types should have quotes around them to indicate they are string types.

```suggestion
| `'kind'`    | `'subtitles'`, `'captions'`, `'descriptions'`, `'chapters'`, `'metadata'` |         | `undefined`  | Sets how the text track is meant to be used                                                                                              |
```

apps/docs/static/example.vtt
rezrah: Sorry, I was referring to the video length, which is 0.59s max right now but your subtitles go on until 1 minute 30s. This is a nitpick though.

packages/react/src/VideoPlayer/components/controls/Controls.tsx
```diff
@@ -0,0 +1,305 @@
...
+  /* 2. Seek                                                    */
+
+  useEffect(() => {
+    setPlayedTime(videoRef.current?.currentTime || 0)
+  }, [videoRef, videoRef.current?.currentTime])
+
+  const renderSeek = useMemo(() => {
+    return (
+      <>
+        <Range
```
rezrah, line 102: FYI this and the other form input based controls are causing AXE violations. 

Storybook 👇 
Screenshot 2023-08-11 at 17 27 17

apps/docs/static/example.vtt
JoshBowdenConcepts: This doesn't actually affect the output which is good. Do we want to have the video actually transcribed or is this ok to keep?

apps/docs/content/components/VideoPlayer.mdx
JoshBowdenConcepts: Fixed in the following [SHA](https://github.com/primer/brand/pull/373/commits/ee6ae80b15875236b720a51258e132c66f39e856)

apps/docs/content/components/VideoPlayer.mdx
JoshBowdenConcepts: Fixed in the following [SHA](https://github.com/primer/brand/pull/373/commits/ee6ae80b15875236b720a51258e132c66f39e856)

apps/docs/content/components/VideoPlayer.mdx
JoshBowdenConcepts: Fixed in the following [SHA](https://github.com/primer/brand/pull/373/commits/ee6ae80b15875236b720a51258e132c66f39e856)

apps/storybook/package.json
```diff
@@ -24,7 +24,7 @@
     "build:preview": "npm run clean && scripts/build-storybook preview",
     "build:prod": "npm run clean && scripts/build-storybook",
     "clean": "rm -rf out",
-    "start": "storybook dev -p 6006"
+    "start": "storybook dev -p 6006 -s ../../docs/public/static"
```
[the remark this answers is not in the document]
JoshBowdenConcepts: Fixed in the following [SHA](https://github.com/primer/brand/pull/373/commits/ee6ae80b15875236b720a51258e132c66f39e856)

packages/react/src/VideoPlayer/components/controls/Controls.tsx
JoshBowdenConcepts: This has been fixed in the following [SHA](https://github.com/primer/brand/pull/373/commits/6eb9240149efbbd452b8726f0fee836860ea3fa5)

.changeset/tame-keys-provide.md
rezrah: @JoshBowdenConcepts A reminder that this still needs to be updated

Screenshot 2023-08-16 at 15 30 15

JoshBowdenConcepts: This has been updated in a more recent SHA

apps/docs/content/components/VideoPlayer.mdx
JoshBowdenConcepts: It seemed to happen because of the tooltip. I think if we want to address this correctly we would need to create a Tooltip component as we have discussed. Then this can be updated.

apps/docs/content/components/VideoPlayer.mdx
JoshBowdenConcepts: This has been updated in the following [SHA](https://github.com/primer/brand/pull/373/commits/0ec9c6c62a4ab24513059c86e087fb4cd7728a33)

apps/docs/content/components/VideoPlayer.mdx
JoshBowdenConcepts: This has been updated in the following [SHA](https://github.com/primer/brand/pull/373/commits/0ec9c6c62a4ab24513059c86e087fb4cd7728a33)

apps/docs/content/components/VideoPlayer.mdx
rezrah: Creating a dedicated tooltip will require considerable effort, so can we offset this particular tooltip from the left by a few pixels in the meantime?

packages/design-tokens/src/tokens/functional/components/video-player/base.json
```diff
@@ -0,0 +1,67 @@
...
+        "gap": {
+          "value": "var(--base-size-16)"
+        },
+        "padding": {
+          "value": "var(--base-size-16) var(--base-size-24)"
+        },
+        "focus": {
+          "boxShadow": {
+            "value": "0 0 0 2px var(--brand-color-focus)"
+          }
```
rezrah, line 14: I don't think we need tokens for these, as they aren't the kind of thing we'd want users to customise. They also add some overhead that we can avoid by using the values directly in the stylesheet. 

E.g. What is the advantage a user has in changing the focus presentation here?

JoshBowdenConcepts: I was thinking that depending on the branding around the video player they might have customized this for different events. I am happy to remove this, however.

JoshBowdenConcepts: I have paired down the tokens in the [SHA](https://github.com/primer/brand/pull/373/commits/7dfa6209762e0ebe9416a9487e55bf0f77f8eac8)

rezrah: Oh interesting. Thanks for clarifying.

- Gap, padding I don't think would be event-specific values 🤔 
- focus colors are globally blue in other components, and shouldn't be customisable for a11y reasons AFAIK. Even if it were, all other focus colors in other colors would still be blue, so this would create visual inconsistencies.

On that basis, I feel we should remove these.

.changeset/tame-keys-provide.md
rezrah: 🤔 Looks the same to me. If you go to the diff view of the PR, the latest commits show this 👇 

Screenshot 2023-08-16 at 16 35 23

Here's the url where i'm seeing this:

https://github.com/primer/brand/pull/373/files

raytalks: @JoshBowdenConcepts Could you please check the PR to see if the change is visible and confirm the push rather than updating the SHA? As a reviewer, I'd like to see the pushed changes.

packages/react/src/VideoPlayer/VideoPlayer.tsx
```diff
@@ -0,0 +1,188 @@
...
+  }
+
+  /* > Hide Default Captions                                                   */
+  useEffect(() => {
+    if (videoRef.current) {
+      for (let i = 0; i < videoRef.current.textTracks.length; i++) {
+        videoRef.current.textTracks[0].mode = 'hidden'
+      }
+    }
+  }, [videoRef])
```
rezrah, line 68: I'm seeing caption-related DOM update events when captions are toggled to off. I imagine this is unintended?

https://github.com/primer/brand/assets/13340707/04077d41-dcba-492a-a350-ece446b240da

packages/react/src/VideoPlayer/VideoPlayer.tsx
```diff
@@ -0,0 +1,188 @@
...
+
+  useEffect(() => {
+    const handleResize = () => {
+      const breakpoint = videoWrapperRef.current?.getBoundingClientRect().width
+      if (breakpoint && breakpoint < 650) {
+        videoWrapperRef.current.classList.add(styles.VideoPlayer__breakpointSmall)
+      } else {
+        videoWrapperRef.current?.classList.remove(styles.VideoPlayer__breakpointSmall)
+      }
+    }
```
rezrah, line 80: I'm likely missing something crucial here, but AFAICT this is just adding some classes for visual styling. If that's the case, can this not be done with css media queries instead of in the JS runtime?

packages/react/src/VideoPlayer/VideoPlayer.tsx
```diff
@@ -0,0 +1,188 @@
...
+      )}
+      ref={videoWrapperRef}
+    >
+      {/* eslint-disable-next-line jsx-a11y/media-has-caption */}
+      <video
+        ref={videoRef}
+        title={title}
+        controls={false}
+        poster={poster}
+        {...rest}
```
rezrah, line 113: Can we put the spread at the end please?

packages/react/src/VideoPlayer/VideoPlayer.tsx
JoshBowdenConcepts: I was handling this specifically based on the size of the container which I would hope to replace with container queries that are supported but not yet available in our system. At least when I tested it out. I was thinking this might be a good candidate in case the Video player was encapsulated. The controls really need to understand the size of the component rather than the size of the page.

packages/react/src/VideoPlayer/VideoPlayer.tsx
JoshBowdenConcepts: Updated in the following [SHA](https://github.com/primer/brand/pull/373/commits/c739ebc20c8a616c696f069f9ba2cc9843488637)

packages/react/src/VideoPlayer/VideoPlayer.tsx
JoshBowdenConcepts: This is actually intended. So to get access to them completely as well as the ability for them to interact with our other controls they are a DOM element.

rezrah: Sorry, to clarify I'm referring to the the caption DOM elements being updated when **captions are off.** That doesn't feel intended, as the user has turned them off and it's creating additional, and unnecessary computation. Could you add a guard check to see if captions are enabled/disabled before running that effect?

JoshBowdenConcepts: I can make this change the tradeoff would be that if a user turns on captions when the update is not running then they won't have access to the captions at that moment. Is this a tradeoff that we want to accept? I would think that more immediate captions would be preferred from an accessibility standpoint. @ericwbailey do you know what is preferred in this case?

ericwbailey: From an end user perspective we'll want to ensure that the end effect of invoking captions leads to them appearing immediately on request and synchronized to the currently spoken audio.

I'm unsure about the performance implication particulars, or if a WebVTT/caption file is smart enough to jump immediately to the timestamp. I don't want to crash the browser or AT, either.

JoshBowdenConcepts: I don't believe this has any heavy browser implications but it does supply the captions immediately when the other approach would wait for the next captions to be sent once they are turned on. So I think the current approach is preferred from a11y but has the tradeoff of slightly negative performance. @rezrah I think this is an ok tradeoff, do you have any major concerns?

rezrah: > when the update is not running then they won't have access to the captions at that moment.

Do you know what the exact impact of this is, specifically how long the delay for the captions to appear, catch up and synchronise is? I'm a little curious if and why the delay would even be noticeable, but if it is not within a second (for example), **happy to leave as-is**. A major concern with any kind of side effect in React is susceptibility to memory leaks. Thinking defensively, i'd prefer that we don't mutate dom elements directly unless we absolutely have to, but happy to leave as-is if we can't achieve synchronisation as Eric alluded to.

packages/react/src/VideoPlayer/VideoPlayer.tsx
```diff
@@ -0,0 +1,213 @@
...
+}: VideoPlayerProps) => {
+  const videoPlayerRef = useRef<HTMLVideoElement>(null)
+  const videoWrapperRef = useRef<HTMLDivElement>(null)
+  const videoRef = ref ? ref : videoPlayerRef
+  const [playing, setPlaying] = useState(false)
+  const [playedTime, setPlayedTime] = useState(0)
+  const [totalTime, setTotalTime] = useState(0)
+  const [trackInformation, setTrackInformation] = useState<TextTrackCueList | undefined>(undefined)
+
+  /* > Video Attr Functions                                            */
```
rezrah, line 47:
```suggestion
  /* Video Attr Functions  */
```

packages/react/src/VideoPlayer/VideoPlayer.tsx
rezrah: 👍 thanks for clarifying

can we change the class to correspond to the container it's applying against? 

the main class is `VideoPlayer__container` so can we call this one by its modifier `VideoPlayer__container--small`. This is to more concisely describe the characteristics without alluding to the underlying implementation detail.

packages/react/src/VideoPlayer/VideoPlayer.tsx
```diff
@@ -0,0 +1,213 @@
...
+              aria-hidden="true"
+            >
+              <path
+                d="M1.56489 23.8112C0.884817 24.2389 1.07491e-06 23.7501 1.0398e-06 22.9467L8.28679e-08 1.05473C4.73246e-08 0.241596 0.904067 -0.245404 1.58307 0.201969L18.5829 11.4026C19.2032 11.8113 19.1935 12.7244 18.5647 13.1198L1.56489 23.8112Z"
+                fill="currentColor"
+              />
+            </svg>
+          </span>
+        )}
+        <span className={styles.VideoPlayer__srOnly}>Play</span>
```
rezrah, line 188: we have a visually-hidden util class, can we use that instead here?

```suggestion
        <span className="visually-hidden">Play</span>
```

packages/react/src/VideoPlayer/VideoPlayer.tsx
```diff
@@ -0,0 +1,213 @@
...
+              viewBox="0 0 24 24"
+              fill="none"
+              style={{marginLeft: 5}}
+              aria-hidden="true"
+            >
+              <path
+                d="M1.56489 23.8112C0.884817 24.2389 1.07491e-06 23.7501 1.0398e-06 22.9467L8.28679e-08 1.05473C4.73246e-08 0.241596 0.904067 -0.245404 1.58307 0.201969L18.5829 11.4026C19.2032 11.8113 19.1935 12.7244 18.5647 13.1198L1.56489 23.8112Z"
+                fill="currentColor"
+              />
+            </svg>
```
rezrah, line 185: Could we make an Octicon request to have play icon included there?

rezrah: Can be done in a follow up PR. I know there's a play Octicon already, but this one is different.

packages/react/src/VideoPlayer/VideoPlayer.tsx
```diff
@@ -0,0 +1,213 @@
...
+      <button className={styles.VideoPlayer__playButton} onClick={handleVideoPlayback}>
+        {!playing && (
+          <span className={styles.VideoPlayer__playButtonInner}>
+            <svg
+              xmlns="http://www.w3.org/2000/svg"
+              width="24"
+              height="24"
+              viewBox="0 0 24 24"
+              fill="none"
+              style={{marginLeft: 5}}
```
rezrah, line 178: Can we remove this inline style please? Maybe put it in a class on the parent? We also prefer rem over px.

packages/react/src/VideoPlayer/VideoPlayer.module.css
```diff
@@ -0,0 +1,405 @@
...
+  justify-content: center;
+  pointer-events: none;
+}
+
+.VideoPlayer__captions__empty {
+  display: none;
+}
+
+.VideoPlayer__captions__text {
+  color: white;
```
rezrah, line 398: Can we use our tokens instead of named colors please?

```suggestion
  color: var(--base-color-scale-white-0);
```

packages/react/src/VideoPlayer/VideoPlayer.module.css
```diff
@@ -0,0 +1,405 @@
...
+  font-size: 16px;
```
rezrah, line 399: We have tokens that use rem values for text, can we either replace with those tokens please, or ideally not override the font-size in css and stick to our `Text` component.

packages/react/src/VideoPlayer/components/captions/Captions.tsx
```diff
@@ -0,0 +1,48 @@
...
+    }
+
+    if (videoRef.current) {
+      videoRef.current.addEventListener('timeupdate', compareAndSetCaption)
+    }
+    return () => window.removeEventListener('timeupdate', compareAndSetCaption)
+  }, [videoRef, trackInformation])
+
+  return (
+    <div className={clsx(styles.VideoPlayer__captions, (disabled || !caption) && styles.VideoPlayer__captions__empty)}>
```
rezrah, line 42: Rather than using display: none to hide the element, could we short-circuit the function altogether outside of this return statement, to render `null` if `disabled || !caption`. Or do you need the DOM elements to always exist?

rezrah: Can we also use BEM conventions here please for consistency..

```suggestion
    <div className={clsx(styles.VideoPlayer__captions, (disabled || !caption) && styles.VideoPlayer__captions---empty)}>
```

packages/react/src/VideoPlayer/components/captions/Captions.tsx
```diff
@@ -0,0 +1,48 @@
...
+      <Text as="p" className={styles.VideoPlayer__captions__text}>
```
rezrah, line 43:
```suggestion
      <Text as="p" className={styles.VideoPlayer__captionText}>
```

packages/react/src/VideoPlayer/components/controls/Controls.tsx
```diff
@@ -0,0 +1,304 @@
...
+          max={totalTime || 0}
+          step={0.0001}
+          onInput={e => {
+            if (videoRef.current) videoRef.current.currentTime = e.currentTarget.valueAsNumber
+          }}
+          value={currentTime}
+          className={styles.VideoPlayer__progressBar2}
+          tooltip
+          tooltipFormatter={value => getMinuteSecondTime(value as number)}
+          id="VideoPlayer__Seek"
```
rezrah, line 115: With multiple videos on the same page, this would cause duplicate id violations, can we make this unique or remove it? I can't see where it's being referenced, so not sure if being used.

packages/react/src/VideoPlayer/components/controls/Controls.tsx
```diff
@@ -0,0 +1,304 @@
...
+      <button
+        className={clsx(
+          styles.VideoPlayer__iconControl,
+          styles.VideoPlayer__closedCaption,
+          !closedCaption && styles.VideoPlayer__ccOff,
+        )}
+        onClick={() => setClosedCaption(!closedCaption)}
+      >
+        <Text className={styles.VideoPlayer__ccText}>CC</Text>
+        <Tooltip>{!closedCaption ? 'Enable Captions' : 'Disable Captions'}</Tooltip>
```
rezrah, line 140:
```suggestion
        <Tooltip>{!closedCaption ? 'Enable captions' : 'Disable captions'}</Tooltip>
```

packages/react/src/VideoPlayer/components/controls/Controls.tsx
```diff
@@ -0,0 +1,304 @@
...
+              fill="currentColor"
+            />
+            <path
+              d="M16.243 7.757C16.1733 7.68733 16.0906 7.63207 15.9996 7.59437C15.9086 7.55667 15.811 7.53726 15.7125 7.53726C15.614 7.53726 15.5164 7.55667 15.4254 7.59437C15.3344 7.63207 15.2517 7.68733 15.182 7.757C15.1123 7.82667 15.0571 7.90937 15.0194 8.0004C14.9817 8.09142 14.9623 8.18898 14.9623 8.2875C14.9623 8.38602 14.9817 8.48358 15.0194 8.5746C15.0571 8.66563 15.1123 8.74833 15.182 8.818C15.5999 9.23586 15.9313 9.73194 16.1575 10.2779C16.3836 10.8239 16.5 11.409 16.5 12C16.5 12.591 16.3836 13.1761 16.1575 13.7221C15.9313 14.2681 15.5999 14.7641 15.182 15.182C15.0495 15.3242 14.9774 15.5122 14.9808 15.7065C14.9843 15.9008 15.063 16.0862 15.2004 16.2236C15.3378 16.361 15.5232 16.4397 15.7175 16.4432C15.9118 16.4466 16.0998 16.3745 16.242 16.242C17.3671 15.1168 17.9992 13.5907 17.9992 11.9995C17.9992 10.4083 17.3671 8.88221 16.242 7.757H16.243Z"
+              fill="currentColor"
+            />
+          </svg>
```
rezrah, line 183: Can we replace this with an Octicon please? Looks like we have one [here](https://primer.style/design/foundations/icons/unmute-24).

packages/react/src/VideoPlayer/components/controls/Controls.tsx
```diff
@@ -0,0 +1,304 @@
...
+            <line
+              x1="16.0607"
+              y1="10.0604"
+              x2="20.6569"
+              y2="14.6566"
+              stroke="currentColor"
+              strokeWidth="1.5"
+              strokeLinecap="round"
+            />
+          </svg>
```
rezrah, line 210: Same for this one, you can use this octicon 

https://primer.style/design/foundations/icons/mute-16

packages/react/src/VideoPlayer/components/controls/Controls.tsx
```diff
@@ -0,0 +1,304 @@
...
+          setFullScreen(true)
+        } else {
+          document.exitFullscreen()
+          setFullScreen(false)
+        }
+      }
+    }
+
+    return (
+      <IconControl onClick={handleFullScreen} tooltip={!fullScreen ? 'Full Screen' : 'Exit Full Screen'}>
```
rezrah, line 255:
```suggestion
      <IconControl onClick={handleFullScreen} tooltip={fullScreen ? 'Exit full screen' : 'Full screen'}>
```

packages/react/src/VideoPlayer/components/controls/Controls.tsx
```diff
@@ -0,0 +1,304 @@
...
+              fill="currentColor"
+              stroke="#fff"
+              strokeWidth="1.5"
+              strokeLinecap="round"
+            />
+          </svg>
+        ) : (
+          <svg xmlns="http://www.w3.org/2000/svg" width="24" height="24" viewBox="0 0 24 24" fill="none">
```
rezrah, line 269: Same comment about using octicons for these, as we seem to have equivalents: https://primer.style/design/foundations/icons/screen-full-16

packages/react/src/VideoPlayer/components/icon-control/IconControl.tsx
```diff
@@ -0,0 +1,19 @@
...
+
+type IconControlProps = {
+  tooltip: string
+} & React.HTMLAttributes<HTMLButtonElement>
+
+export const IconControl = ({tooltip, children, className, ...rest}: IconControlProps) => {
+  return (
+    <button className={clsx(styles.VideoPlayer__iconControl, className)} {...rest} aria-label={tooltip}>
+      {children}
+      <span className={styles.VideoPlayer__srOnly}>{tooltip}</span>
```
rezrah, line 15: Can we use our `visually-hidden` class here instead of creating a new style for it?

packages/react/src/VideoPlayer/VideoPlayer.tsx
JoshBowdenConcepts: Definitely! Should we wait for that to be included or move forward with this and then update?

packages/react/src/VideoPlayer/components/icon-control/IconControl.tsx
JoshBowdenConcepts: This has been updated in this SHA: https://github.com/primer/brand/pull/373/commits/3476819281a2ff27b50715bf57261befc6f1d0f5

packages/react/src/VideoPlayer/components/controls/Controls.tsx
JoshBowdenConcepts: This has been updated in this SHA: https://github.com/primer/brand/pull/373/commits/3476819281a2ff27b50715bf57261befc6f1d0f5

packages/react/src/VideoPlayer/components/controls/Controls.tsx
JoshBowdenConcepts: This has been updated in this SHA: https://github.com/primer/brand/pull/373/commits/3476819281a2ff27b50715bf57261befc6f1d0f5

packages/react/src/VideoPlayer/components/controls/Controls.tsx
JoshBowdenConcepts: This has been updated in this SHA: https://github.com/primer/brand/pull/373/commits/3476819281a2ff27b50715bf57261befc6f1d0f5

packages/react/src/VideoPlayer/components/controls/Controls.tsx
JoshBowdenConcepts: This has been updated in this SHA: https://github.com/primer/brand/pull/373/commits/3476819281a2ff27b50715bf57261befc6f1d0f5

packages/react/src/VideoPlayer/components/controls/Controls.tsx
JoshBowdenConcepts: This has been updated in this SHA: https://github.com/primer/brand/pull/373/commits/3476819281a2ff27b50715bf57261befc6f1d0f5

packages/react/src/VideoPlayer/components/controls/Controls.tsx
JoshBowdenConcepts: This has been updated in this SHA: https://github.com/primer/brand/pull/373/commits/3476819281a2ff27b50715bf57261befc6f1d0f5

packages/react/src/VideoPlayer/components/controls/Controls.tsx
```diff
@@ -0,0 +1,304 @@
...
+        <Range
+          type="range"
+          min="0"
+          max={totalTime || 0}
+          step={0.0001}
+          onInput={e => {
+            if (videoRef.current) videoRef.current.currentTime = e.currentTarget.valueAsNumber
+          }}
+          value={currentTime}
+          className={styles.VideoPlayer__progressBar2}
```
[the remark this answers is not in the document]
JoshBowdenConcepts: This has been updated in this SHA: https://github.com/primer/brand/pull/373/commits/3476819281a2ff27b50715bf57261befc6f1d0f5

packages/react/src/VideoPlayer/components/controls/Controls.tsx
```diff
@@ -0,0 +1,304 @@
...
+  }, [playingProp])
+
+  const renderPlayPauseButton = useMemo(() => {
+    return (
+      <IconControl
+        onClick={() => {
+          setPlaying(!playing)
+        }}
+        className={styles.VideoPlayer__shiftTooltipRight}
+        tooltip={!playing ? 'Play Video' : 'Pause Video'}
```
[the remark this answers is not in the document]
JoshBowdenConcepts: This has been updated in this SHA: https://github.com/primer/brand/pull/373/commits/3476819281a2ff27b50715bf57261befc6f1d0f5

packages/react/src/VideoPlayer/components/captions/Captions.tsx
JoshBowdenConcepts: This has been updated in this SHA: https://github.com/primer/brand/pull/373/commits/3476819281a2ff27b50715bf57261befc6f1d0f5

packages/react/src/VideoPlayer/components/captions/Captions.tsx
JoshBowdenConcepts: This has been updated in this SHA: https://github.com/primer/brand/pull/373/commits/3476819281a2ff27b50715bf57261befc6f1d0f5

packages/react/src/VideoPlayer/VideoPlayer.module.css
JoshBowdenConcepts: This has been updated in this SHA: https://github.com/primer/brand/pull/373/commits/3476819281a2ff27b50715bf57261befc6f1d0f5

packages/react/src/VideoPlayer/VideoPlayer.module.css
JoshBowdenConcepts: Agreed we will just use the default size for this. This has been updated in this SHA: https://github.com/primer/brand/pull/373/commits/3476819281a2ff27b50715bf57261befc6f1d0f5

packages/react/src/VideoPlayer/VideoPlayer.tsx
JoshBowdenConcepts: This has been updated in this SHA: https://github.com/primer/brand/pull/373/commits/3476819281a2ff27b50715bf57261befc6f1d0f5

packages/react/src/VideoPlayer/VideoPlayer.tsx
JoshBowdenConcepts: This has been updated in this SHA: https://github.com/primer/brand/pull/373/commits/3476819281a2ff27b50715bf57261befc6f1d0f5

packages/react/src/VideoPlayer/VideoPlayer.tsx
JoshBowdenConcepts: Are we avoiding stylistic separators? Happy to update them all but just curious.

packages/react/src/VideoPlayer/VideoPlayer.tsx
JoshBowdenConcepts: Updated in the following SHA: https://github.com/primer/brand/pull/373/commits/421806a546e1941ec949334a3422e4f70e28faf0

packages/react/src/VideoPlayer/components/controls/Controls.tsx
```diff
@@ -0,0 +1,224 @@
...
+    return (
+      <>
+        <Range
+          type="range"
+          min="0"
+          max={totalTime || 0}
+          step={0.0001}
+          onInput={e => {
+            if (videoRef.current) videoRef.current.currentTime = e.currentTarget.valueAsNumber
+          }}
```
rezrah, line 111: Could this inline function also be hoisted out of the render function?

packages/react/src/VideoPlayer/VideoPlayer.test.tsx
rezrah: Which state are you looking to update?

packages/react/src/VideoPlayer/VideoPlayer.module.css
```diff
@@ -0,0 +1,398 @@
...
+/* ---------------------------------------------------------- */
+
+.VideoPlayer__tooltip {
+  position: absolute;
+  bottom: 100%;
+  /* left: 50%; */
+  z-index: 2;
+  width: 10px;
+  height: 10px;
+  margin-bottom: 16px;
```
rezrah, line 332:
```suggestion
  margin-bottom: var(--base-size-16);
```

packages/react/src/VideoPlayer/VideoPlayer.module.css
```diff
@@ -0,0 +1,398 @@
...
+  pointer-events: none;
+  opacity: 0;
+  transform: translate3d(0%, 50%, 0);
+  transition: opacity 0.5s cubic-bezier(0.16, 1, 0.3, 1), transform 0.5s cubic-bezier(0.16, 1, 0.3, 1);
```
rezrah, line 336: Can we use our global timings here?

```suggestion
  transition: opacity var(--brand-animation-duration-default) var(--brand-animation-easing-default), transform var(--brand-animation-duration-default) var(--brand-animation-easing-default);
```

packages/react/src/VideoPlayer/VideoPlayer.module.css
```diff
@@ -0,0 +1,398 @@
...
+/* 3. Center Play Button                                      */
+/* ---------------------------------------------------------- */
+
+.VideoPlayer__playButton {
+  position: absolute;
+  top: 0;
+  left: 0;
+  width: 100%;
+  height: 100%;
+  background-color: rgba(0, 0, 0, 0);
```
rezrah, line 45: Should this just be `transparent`? Or not set at all?

packages/react/src/VideoPlayer/VideoPlayer.test.tsx
JoshBowdenConcepts: I would love to pair on it. I specifically can't get the seek state to update in a way that I am able to test. I am free tomorrow morning if you have some time I will make an invite.

packages/react/src/VideoPlayer/VideoPlayer.module.css
JoshBowdenConcepts: This has been updated in this SHA https://github.com/primer/brand/pull/373/commits/ae69cf85dd5659e66feca9b7586ee79ae7c33dee

packages/design-tokens/src/tokens/functional/components/video-player/base.json
```diff
@@ -0,0 +1,46 @@
...
+        "borderRadius": {
+          "value": "3px"
+        },
+        "minWidth": {
+          "value": "75px"
+        }
+      },
+      "transition": {
+        "value": "all 0.4s ease-out"
+      },
```
rezrah, line 38: Could we use our global animation timing tokens here instead please? I think I left a related comment somewhere in this PR, so perhaps this is no longer required?

packages/react/src/VideoPlayer/VideoPlayer.module.css
```diff
@@ -0,0 +1,399 @@
...
+}
+
+/* ---------------------------------------------------------- */
+/* 7. Tooltip                                                 */
+/* ---------------------------------------------------------- */
+
+.VideoPlayer__tooltip {
+  position: absolute;
+  bottom: 100%;
+  /* left: 50%; */
```
rezrah, line 328: can you remove this if it's not needed?

packages/react/src/VideoPlayer/VideoPlayer.module.css
```diff
@@ -0,0 +1,399 @@
...
+
+.VideoPlayer__tooltip::before {
+  position: absolute;
+  bottom: -4px;
+  left: 0;
+  width: 10px;
+  height: 10px;
+  content: '';
+  background: var(--base-color-scale-white-0);
+  border-radius: 2px;
```
rezrah, line 348: Our smallest token radius is 3px, which feels like a negligible change. We should prefer tokens over hard-coded values where possible.

```suggestion
  border-radius: var(--brand-borderRadius-small);
```

JoshBowdenConcepts: This has been updated in this SHA: https://github.com/primer/brand/pull/373/commits/53e6be6f783cb211109e7502de6eb28f3b6699e4

packages/react/src/VideoPlayer/VideoPlayer.module.css
JoshBowdenConcepts: This has been updated in this SHA: https://github.com/primer/brand/pull/373/commits/53e6be6f783cb211109e7502de6eb28f3b6699e4

packages/design-tokens/src/tokens/functional/components/video-player/base.json
JoshBowdenConcepts: This has been updated in this SHA: https://github.com/primer/brand/pull/373/commits/53e6be6f783cb211109e7502de6eb28f3b6699e4

rezrah (approved):
It's looking good 👍 Thanks for going through the feedback @JoshBowdenConcepts. Let's get this out for the next release 🚀

PR #375, opened 2023-08-04
Bump Doctocat 4.6.3

josepmartins:
Bump `Doctocat` theme to version 4.6.3 in apps/docs. Fixes the selected state in the top navigation.


rezrah (approved):
Checked the changelog as we're going up multiple versions, but couldn't see anything else that could cause a regression.. 👍 :shipit:

PR #380, opened 2023-08-07
Add Timeline component

danielguillan:
## Summary

New Timeline component to display a list of related items as a sequence of events.

- [React documentation](https://primer-16881e08d0-26139705.drafts.github.io/components/Timeline)
- [Storybook examples](https://primer-16881e08d0-26139705.drafts.github.io/storybook/?path=/story/components-timeline--default)

## What should reviewers focus on?

- Component name (open to suggestions)
- Code
- Storybook examples
- Documentation

## Supporting resources (related issues, external links, etc):

- https://github.com/github/primer/issues/2508

## Contributor checklist:

- [x] All new and existing CI checks pass
- [x] Tests prove that the feature works and covers both happy and unhappy paths
- [x] Any drop in coverage, breaking changes or regressions have been documented above
- [x] New visual snapshots have been generated / updated for any UI changes
- [x] All developer debugging and non-functional logging has been removed
- [x] Related issues have been referenced in the PR description

## Reviewer checklist:

- [ ] Check that pull request and proposed changes adhere to our [contribution guidelines](../../CONTRIBUTING.md) and [code of conduct](../../CODE_OF_CONDUCT.md)
- [ ] Check that tests prove the feature works and covers both happy and unhappy paths
- [ ] Check that there aren't other open Pull Requests for the same update/change

## Screenshots:

A screenshot of a Timeline component example showing 3 items

PR #381, opened 2023-08-07
Improve trailingComponent feature of River.Content

rezrah:
## Summary

Part of https://github.com/github/primer/issues/2507

Reuses the existing `trailingComponent` prop (forgot this existed 🤦) to achieve requirement around custom post content in the River. This PR focusses now on adding the required spacing and optional divider.

ℹ️ **Adding Hubot sans support is outside the  scope of this PR**

🔗 [Try it in Storybook](https://primer-bdfebab8ac-26139705.drafts.github.io/storybook/?path=/story/components-river-fixtures--custom-trailing-content)

## List of notable changes:

- **added** default styling for the existing `trailingComponent` prop
- **added** optional divider after the `trailingComponent` prop
- **added** some example stories on how to use this prop, as it was previously undocumented in Storybook

- Check out Storybook
- Anything I've missed

- https://github.com/github/primer/issues/2507

## Contributor checklist:

## Reviewer checklist:

### Default (no divider)
Screenshot 2023-08-07 at 14 00 22

### With divider
Screenshot 2023-08-07 at 14 00 13

### Alternative custom content
Screenshot 2023-08-07 at 14 00 17


packages/react/src/River/River.fixtures.stories.tsx
```diff
@@ -120,6 +120,86 @@ export const AlternativeHeadingSize: StoryFn<typeof River> = () => (
   </Container>
 )
 
+const ExampleTrailingComponentOne = () => (
```
mperrotti, line 123: This doesn't use a `River` component. Is that intentional?

packages/react/src/River/River.fixtures.stories.tsx
```diff
@@ -120,6 +120,86 @@ export const AlternativeHeadingSize: StoryFn<typeof River> = () => (
...
+  </Stack>
+)
+
+export const CustomTrailingContent: StoryFn<typeof River> = () => (
+  <Container>
+    <River>
+      <River.Visual>
+        <PlaceholderImage />
+      </River.Visual>
+      <River.Content trailingComponent={ExampleTrailingComponentOne}>
```
mperrotti, line 152: Should we take this opportunity to convert to prop to a child component? I'm fine with either approach, but this seemed like an opportunity to consider.

mperrotti (approved):

packages/react/src/River/River.fixtures.stories.tsx
rezrah: Yeah, this is a local story fixture that we're using for the River's `trailingComponent`. It's used in the adjacent stories, and not a story itself (not exported).

packages/react/src/River/River.fixtures.stories.tsx
rezrah: 🤔 yeah possibly. 

Reasons for keeping this way:

- we already have the trailingComponent prop, and can therefore avoid a breaking change.
- a dedicated child may give open floodgates to misusing the current API. Right now it behaves more as an escape hatch, via a prop.. making it a dedicated child could lead to something like this:

```jsx
<River>
  <River.TrailingComponent>
    <Heading .../ >
    <Text />
    <Text />
  </River.TrailingComponent>
</River>
```

Granted, you _could_ do this with the prop version, but it _feels_ much more hacky, which is a good thing.

JoshBowdenConcepts (approved):
LGTM 👍🏽

ajashams:
@rezrah Will the `default trialing component` eventually become the `river with the timeline component` in it? (see below). The reason I ask is because the `default` is an _**experimental**_ component that we're not using anywhere and haven't really pressure tested yet. I think this may have been pulled from the Refactor file, but we'd ideally be going off of the [Universe, Enterprise and Preview templates](https://www.figma.com/file/m4U5DInCRrbXCwOfWJUfQf/Contentful-Templates?type=design&node-id=0%3A1&mode=design&t=MA6GXAimo8T2CSOx-1) and their variants.   

If that's not accurate, please feel free to clarify! 

| trailing component (default) | river + timeline |
| --- | --- |
| Screenshot 2023-08-08 at 5 07 22 PM | Screenshot 2023-08-08 at 5 09 33 PM

rezrah:
@ajashams - thanks for checking in on that.

The timeline component hasn't shipped yet, but ICYMI [there is an open PR by @danielguillan here](https://github.com/primer/brand/pull/380) if you want to review that also.

The plan is:

- **If Timeline merges first,** I update this PR to replace the trailing component examples with Timeline
- **If this PR merges first,** we open a follow-up PR to replace this version with Timeline. 

Does that work for you?

jesussandreas:
Hey @rezrah minor comments here.

Could we address how the spacing looks by default?

As in spacing between Headline and paragraph should be 16px. (currently 8px).

and the spacing when the border is not shown should be less. my recommendation is 32px.

josepmartins:
> Hey @rezrah minor comments here.
> 
> Could we address how the spacing looks by default?
> 
> As in spacing between Headline and paragraph should be 16px. (currently 8px).
> 
> and the spacing when the border is not shown should be less. my recommendation is 32px.


hey @jesussandreas, sounds good. As this is already merged, do you mind creating a [new primer pattern issue](https://github.com/github/primer/issues/new/choose) with those changes 🙏 ? similar to the latest [Pillar visual updates](https://github.com/github/primer/issues/2505). We need to keep track of these changes as they need to be applied in both figma and react libraries.

PR #390, opened 2023-08-10
Bento Component

JoshBowdenConcepts:
## Summary

Creation of the Bento component which will act as a presentational component. Most notably it will be used soon on the Universe page.

## List of notable changes:

- **added** Bento component

- The design and functionality works as expected based on the Figma component.

## Steps to test:

1. Open the Bento component in CI-deployed preview environment
2. Go to Bento and Bento.Item stories in Storybook
3. Verify that they behave as described in the following issue.

- [API Exploration](https://www.figma.com/file/gwcGUHN5MRHWFH5I7l2Z75/Primer-Brand-Bento-API-exploration?type=whiteboard&node-id=0-1&t=mOh4HxcxQowQF17Y-0)
- [Expected Variants](https://www.figma.com/file/BsVDcDotz7pyQPR6L0HUIb/Bento?type=design&node-id=79-468&mode=design)
- [Figma Component Loom](https://www.loom.com/share/cf7c2fb3f1e54f89a9e7f7ebe88a9cf5)

## Contributor checklist:

## Reviewer checklist:

image


packages/react/src/Bento/Bento.module.css
```diff
@@ -0,0 +1,285 @@
+.Bento {
+  display: grid;
+  width: 100%;
+  grid-template-columns: repeat(12, minmax(0, 1fr));
+  grid-template-rows: minmax(40px, auto);
```
josepmartins, line 5: Based on Enterprise designs, where some items are ~560px or ~640px tall I'd go with a default value of 5rem. As it's flexible enough to get those values by just spanning the item rows.

- 5rem x 7 rows = 560px
- 5rem x 8 rows = 640px

+ I think we could just use `grid-auto-rows`

grid
```suggestion
  grid-auto-rows: var(--base-size-80);
```

josepmartins (commented):
I'm aware its stilla a draft PR @JoshBowdenConcepts! 🙏 

just leaving early comments around the CSS that might be helpful

packages/react/src/Bento/Bento.module.css
JoshBowdenConcepts: I tried `--base-size-80` which looked a bit to large but `--base-size-40` was just right. Is this update alright to make or should I leave 80. I think creating a demo with content will help in the docs greatly.

packages/design-tokens/src/tokens/functional/components/bento/colors.json
```diff
@@ -0,0 +1,12 @@
+{
+  "brand": {
+    "Bento": {
+      "background": {
```
langermank, line 4:
```suggestion
      "bgColor": {
```

Just getting ahead of token name changes coming up, might as well start using it for new tokens

JoshBowdenConcepts: Ok awesome!

packages/design-tokens/src/tokens/functional/components/bento/colors.js
```diff
@@ -0,0 +1,10 @@
+module.exports = {
+  brand: {
+    Bento: {
+      background: {
```
langermank, line 4:
```suggestion
      bgColor: {
```

😄

packages/react/src/Bento/Bento.module.css
```diff
@@ -0,0 +1,333 @@
...
+}
+
+.Bento__Item {
+  background-color: var(--brand-Bento-background);
+  border-radius: var(--brand-Bento-Item-borderRadius-small);
+  display: grid;
+  grid-template-rows: 1fr 1fr;
+  grid-template-columns: 1fr;
+  overflow: hidden;
+  border: 1px solid var(--base-color-scale-gray-3);
```
langermank, line 17: Do you know if there's a reason this needs a special color, and not one of the default border colors?

JoshBowdenConcepts: This is all still draft so I was just doing mock-ups :) But it will use a default border color and if not I will ask design about it.

packages/design-tokens/src/tokens/functional/components/bento/colors.js
```diff
@@ -0,0 +1,10 @@
...
+        value: 'var(--base-color-scale-white-0)',
+        dark: 'var(--base-color-scale-gray-9)',
```
josepmartins, line 6: This should be the same as we use [in Card](https://github.com/primer/brand/blob/main/packages/design-tokens/src/tokens/functional/components/card/colors.json#L5): 

```suggestion
        value: 'var(--brand-color-canvas-default)',
        dark: 'var(--brand-color-canvas-subtle)',
```

packages/design-tokens/src/tokens/functional/components/bento/base.json
```diff
@@ -0,0 +1,30 @@
...
+    "Bento": {
+      "gap": {
+        "large": {
+          "value": "var(--base-size-32)"
+        },
+        "medium": {
+          "value": "var(--base-size-24)"
+        },
+        "small": {
+          "value": "var(--base-size-16)"
```
josepmartins, line 12: Small gaps should use 20px:
```suggestion
          "value": "var(--base-size-20)"
```

packages/react/src/Bento/Bento.module.css
```diff
@@ -0,0 +1,373 @@
...
+}
+
+/* Bento Content */
+.Bento__Item--visual-as-background .Bento__Content {
+  order: 1;
+  grid-row: 1 / span 2;
+  grid-column: 1 / span 2;
+}
+
+.Bento__Content {
```
josepmartins, line 267: The content area needs an inner padding of `56px`

packages/react/src/Bento/Bento.module.css
```diff
@@ -0,0 +1,373 @@
...
+  display: grid;
+  height: 100%;
+  align-content: space-between; /* Needs to be space-around to handle without img */
```
josepmartins, line 270: This might not be necessary, I've made some explorations using `margin-top: auto` on the call to action.

[Codepen example](https://codepen.io/josepmartins/pen/YzRBjYV)

packages/react/src/Bento/Bento.module.css
```diff
@@ -0,0 +1,373 @@
...
+.Bento__Visual {
+  height: 100%;
+  width: 100%;
+  overflow: hidden;
+}
+
+.Bento__Item--visual-as-background .Bento__Visual {
+  order: -1;
+  grid-row: 1 /-1;
+  grid-column: 1 /-1;
```
josepmartins, line 284: Nice ❤️

josepmartins (commented):
More minor comments 👍

packages/react/src/Bento/Bento.module.css
JoshBowdenConcepts: A static 56px rather than a configurable padding value like for Visual. The 56px would look like this on mobile:
image

josepmartins: I think we'll need something configurable, like the border radius or the gap between items in bento

JoshBowdenConcepts: So I guess my question is if we should provide a map as I have provided to specific values or should we not allow updating or some mix of both? Let's chat Monday about this. I would love to get your input.

packages/design-tokens/src/tokens/functional/components/bento/colors.js
JoshBowdenConcepts: Updated in this [SHA](https://github.com/primer/brand/pull/390/commits/25fcf6f8e2feb02da8119da3a847cbdf161e7807)

packages/react/src/Bento/Bento.module.css
JoshBowdenConcepts: I was able to remove this value by making sure the link was appended to the bottom content. I believe this should work. Let me know if you have any issues with this implementation. :) https://github.com/primer/brand/pull/390/commits/25fcf6f8e2feb02da8119da3a847cbdf161e7807

JoshBowdenConcepts:
Things to complete:
- [ ] vertical and horizontal align need to work for content. This might require adding a text alignment prop as well.
- [ ] Docs need to be checked for accuracy and that they provide enough information.

packages/react/src/Bento/Bento.tsx
```diff
@@ -0,0 +1,181 @@
...
+import styles from './Bento.module.css'
+
+export type Size = 'xsmall' | 'small' | 'medium' | 'large' | 'xlarge' | 'xxlarge'
+export type ColumnIndex = 1 | 2 | 3 | 4 | 5 | 6 | 7 | 8 | 9 | 10 | 11 | 12
+
+type ResponsiveColumnIndex = Record<Size, ColumnIndex> | ColumnIndex
+type ResponsiveRowIndex = Record<Size, number> | number
+
+type Flow = 'row' | 'column'
+type ResponsiveFlow = Record<Size, Flow> | Flow
```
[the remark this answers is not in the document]
JoshBowdenConcepts: I have updated this with documentation updates as well: https://github.com/primer/brand/pull/390/commits/dff3c0d6fe1ab0685952fc70c4a1bc70e31d81da

packages/design-tokens/src/tokens/functional/components/bento/base.json
```diff
@@ -0,0 +1,43 @@
...
+          "value": "var(--base-size-32)"
+        },
+        "medium": {
+          "value": "var(--base-size-24)"
+        },
+        "small": {
+          "value": "var(--base-size-20)"
+        }
+      },
+      "Item": {
```
rezrah, line 15: Can we lowercase the Item and Content please? It's only the pattern name that should be pascal cased.

packages/design-tokens/src/tokens/functional/components/bento/base.json
```diff
@@ -0,0 +1,43 @@
...
+        "borderRadius": {
+          "small": {
+            "value": "16px"
+          },
+          "medium": {
+            "value": "20px"
+          },
+          "large": {
+            "value": "24px"
+          }
```
rezrah, line 25: Do these need to be px instead of using our rem / design token values?

```suggestion
          "small": {
            "value": "var(--base-size-16)"
          },
          "medium": {
            "value": "var(--base-size-20)"
          },
          "large": {
            "value": "var(--base-size-24)"
          }
```

packages/react/src/Bento/Bento.features.stories.tsx
```diff
@@ -0,0 +1,163 @@
...
+      </Bento.Visual>
+    </Bento.Item>
+  </Bento>
+)
+
+export const Example: StoryFn<typeof Bento> = () => (
+  <>
+    <style>
+      {`.purple-background {
+                background: linear-gradient(90deg, #6C5DD3 0%, #613DB4 100%);
```
rezrah, line 118: Could we use our base color scale values here instead of the hard coded values? Also the formatting seems off, could we move this into a separate stylesheet perhaps? E.g. Bento.features.stories.module.css

packages/react/src/Bento/Bento.module.css
```diff
@@ -0,0 +1,1793 @@
...
+.Bento__Content {
+  display: flex;
+  flex-direction: column;
+  align-items: start;
+  height: 100%;
+  color: var(--brand-color-text-default);
+}
+
+/* Column Span */
+.Bento__Item--column-span-1 {
```
rezrah, line 54: This is quite a lot of duplicate css from Grid we're shipping for column spanning. Curious if we explored using the grid internally for this component? [Here's the same styles from that component](https://github.com/primer/brand/blob/04790926c578c535cd39b410441ca8252180b2d8/packages/react/src/Grid/Grid.module.css#L37).

packages/react/src/Bento/Bento.module.css
```diff
@@ -0,0 +1,1793 @@
...
+  height: 100%;
+  width: 100%;
+  object-fit: cover;
+}
+
+.Bento__Visual-no-fill > img {
+  height: 100%;
+  width: 100%;
+  object-fit: contain;
+  border-radius: 0px;
```
rezrah, line 298:
```suggestion
  border-radius: 0;
```

packages/react/src/Bento/Bento.module.css
```diff
@@ -0,0 +1,1793 @@
...
+  }
+  .Bento-padding--normal > img {
+    border-radius: var(--brand-Bento-Item-borderRadius-large);
+  }
+  .Bento-padding--spacious > img {
+    border-radius: var(--brand-Bento-Item-borderRadius-large);
+  }
+}
+
+@media screen and (min-width: 48rem) {
```
rezrah, line 334: Just for ease of reading, could you put the lower breakpoint queries above the higher ones? Especially because we're using min-width here.

packages/react/src/Bento/Bento.module.css
```diff
@@ -0,0 +1,1793 @@
...
+    align-items: center;
+  }
+
+  .Bento__Item--xsmall-horizontalAlign-end {
+    align-items: end;
+  }
+}
+
+/* SMALL breakpoint */
+@media screen and (min-width: 0) {
```
rezrah, line 606: This value should be `34rem` for the small breakpoint, currently the same as xsmall.

```suggestion
@media screen and (min-width: 34rem) {
```

packages/react/src/Bento/Bento.module.css
```diff
@@ -0,0 +1,1793 @@
...
+    align-items: center;
+  }
+
+  .Bento__Item--xlarge-horizontalAlign-end {
+    align-items: end;
+  }
+}
+
+/* XXLARGE breakpoint */
+@media screen and (min-width: 80rem) {
```
rezrah, line 1558: XXLARGE breakpoint value is `90rem`. 

```suggestion
@media screen and (min-width: 90rem) {
```

packages/react/src/Bento/Bento.features.stories.tsx
```diff
@@ -0,0 +1,163 @@
...
+    	}
+		.block-icon {
+			display: block !important;
+		}
+		`}
+    </style>
+    <Bento>
+      <Bento.Item columnSpan={7} rowSpan={4} colorMode="dark" visualAsBackground className="purple-background">
+        <Bento.Content padding="normal" fixedBottomLink leadingVisual={<CopilotIcon />}>
```
rezrah, line 127: There's no spacing between the leading visual and heading, which looks quite odd. Is that per the design?

Screenshot 2023-08-24 at 09 27 41

packages/react/src/Bento/Bento.tsx
```diff
@@ -0,0 +1,181 @@
...
+          size: HeadingChild.props.size || '3',
+        })}
+      {React.isValidElement(TextChild) &&
+        React.cloneElement(TextChild as React.ReactElement<TextProps>, {
+          variant: 'muted',
+          as: 'p',
+        })}
+      {React.isValidElement(LinkChild) &&
+        React.cloneElement(LinkChild as React.ReactElement<LinkProps>, {
+          className: clsx(
```
rezrah, line 138: From what I'm seeing in Figma, the default color for links is `accent`, can we apply that here? Currently using default.

packages/react/src/Bento/Bento.tsx
```diff
@@ -0,0 +1,181 @@
...
+  const TextChild = React.Children.toArray(children).find(child => React.isValidElement(child) && child.type === Text)
+  const LinkChild = React.Children.toArray(children).find(child => React.isValidElement(child) && child.type === Link)
+  return (
+    <div className={clsx(styles[`Bento-padding--${padding}`], ...bentoContentClassArray, className)} {...rest}>
+      {React.isValidElement(LeadingVisual) &&
+        React.cloneElement(LeadingVisual as React.ReactElement<IconProps>, {
+          size: LeadingVisual['size'] || 32,
+        })}
+      {React.isValidElement(HeadingChild) &&
+        React.cloneElement(HeadingChild as React.ReactElement<HeadingProps>, {
```
rezrah, line 127: Are we supporting duo tone heading text? Can we add an example in Storybook for how to do this?

Figma:

packages/react/src/Bento/Bento.tsx
```diff
@@ -0,0 +1,181 @@
...
+  const HeadingChild = React.Children.toArray(children).find(
+    child => React.isValidElement(child) && child.type === Heading,
+  )
+  const TextChild = React.Children.toArray(children).find(child => React.isValidElement(child) && child.type === Text)
+  const LinkChild = React.Children.toArray(children).find(child => React.isValidElement(child) && child.type === Link)
+  return (
+    <div className={clsx(styles[`Bento-padding--${padding}`], ...bentoContentClassArray, className)} {...rest}>
+      {React.isValidElement(LeadingVisual) &&
+        React.cloneElement(LeadingVisual as React.ReactElement<IconProps>, {
+          size: LeadingVisual['size'] || 32,
```
rezrah, line 124: I'm seeing `44px` icon size in the design 🤔 , where are you seeing `32`?

packages/react/src/Bento/Bento.tsx
```diff
@@ -0,0 +1,181 @@
...
+
+const Root = ({className, ...rest}: BentoProps) => {
+  return <section className={clsx(styles.Bento, className)} {...rest}></section>
+}
+
+type BentoItemProps = {
+  columnStart?: ResponsiveColumnIndex
+  columnSpan?: ResponsiveColumnIndex
+  rowStart?: ResponsiveRowIndex
+  rowSpan?: ResponsiveRowIndex
```
rezrah, line 33: Intellisense is a little misleading here IMO, as the value isn't just a responsive value as the name suggests but a union of responsive _and_ non-responsive values (`Partial<Record<Size, Align>> | Align`), which this value corresponds to. Can we reference the union here instead? We [do the same in Grid](https://github.com/primer/brand/blob/04790926c578c535cd39b410441ca8252180b2d8/packages/react/src/Grid/Grid.tsx#L91), for example.

packages/react/src/Bento/Bento.tsx
```diff
@@ -0,0 +1,181 @@
...
+  flow?: ResponsiveFlow
+  colorMode?: 'light' | 'dark'
```
rezrah, line 35: We have an [enum](https://github.com/primer/brand/blob/04790926c578c535cd39b410441ca8252180b2d8/packages/react/src/ThemeProvider/ThemeProvider.tsx#L3) for these values, could we use those instead please? 

```suggestion
  colorMode?: ColorModesEnum.LIGHT | ColorModesEnum.DARK
```

packages/react/src/Bento/Bento.tsx
```diff
@@ -0,0 +1,181 @@
...
+    ...returnClassBasedOnResponsiveMap('column-span', columnSpan),
+    ...returnClassBasedOnResponsiveMap('row-span', rowSpan),
+    ...returnClassBasedOnResponsiveMap('column-start', columnStart),
+    ...returnClassBasedOnResponsiveMap('row-start', rowStart),
+    ...returnClassBasedOnResponsiveMap('flow', flow),
+  )
+
+  return (
+    <div
+      data-color-mode={colorMode}
```
rezrah, line 78: Was there a reason for not using ThemeProvider here to set the data-color-mode attribute? Replicating that functionality here makes our color modes more brittle to refactoring, so would be great to use ThemeProvider directly instead.

packages/react/src/Bento/Bento.tsx
```diff
@@ -0,0 +1,181 @@
...
+          size: LeadingVisual['size'] || 32,
+        })}
+      {React.isValidElement(HeadingChild) &&
+        React.cloneElement(HeadingChild as React.ReactElement<HeadingProps>, {
+          as: HeadingChild.props.as || 'h3',
+          size: HeadingChild.props.size || '3',
+        })}
+      {React.isValidElement(TextChild) &&
+        React.cloneElement(TextChild as React.ReactElement<TextProps>, {
+          variant: 'muted',
```
rezrah, line 133: Can we forward the override too? I suspect this will come up down the line, and you're doing it for the other props too. Fine to set a default value though.

```suggestion
          variant: TextChild.props.variant || 'muted',
```

packages/react/src/Bento/Bento.tsx
```diff
@@ -0,0 +1,181 @@
...
+          as: 'p',
+        })}
+      {React.isValidElement(LinkChild) &&
+        React.cloneElement(LinkChild as React.ReactElement<LinkProps>, {
+          className: clsx(
+            LinkChild.props.className,
+            styles['Bento__call-to-action'],
+            fixedBottomLink && styles['Bento__call-to-action--fixed'],
```
rezrah, line 141: Do you need to also forward the `className` prop here too? I think it'll get overridden by this value right now.

packages/react/src/Bento/BentoItem.features.stories.tsx
```diff
@@ -0,0 +1,455 @@
...
+    <Bento.Content>
+      <Heading as="h3">Heading</Heading>
+      <Text>
+        Lorem ipsum dolor sit amet, consectetur adipiscing elit. In sapien sit ullamcorper id. Aliquam luctus sed turpis
+        felis nam pulvinar risus elementum.
+      </Text>
+      <Link href="#">Call to action</Link>
+    </Bento.Content>
+    <Bento.Visual>
+      <img className="test" alt="A random avatar" src={universeImage} />
```
rezrah, line 251: Appreciate this is just an example, but can we source a better image that doesn't interfere with text legibility here?

Screenshot 2023-08-24 at 09 53 36

Our Storybook is usually referenced for A11Y reviews, and this is something that'll get immediately flagged but is easily avoidable.

packages/react/src/Bento/Bento.stories.tsx
```diff
@@ -0,0 +1,32 @@
...
+    <Bento.Item columnSpan={6} rowSpan={8} />
+    <Bento.Item columnSpan={3} rowSpan={4} rowStart={9} />
+    <Bento.Item columnSpan={3} rowSpan={8} />
+    <Bento.Item columnSpan={6} rowSpan={4} />
+  </Bento>
+)
+
+export const Default = Template.bind({})
+
+export const Playground = Template.bind({})
```
rezrah, line 32: The playground is the same as default. Could we add some controls to this story?

Screenshot 2023-08-24 at 09 58 21

packages/react/src/Bento/Bento.module.css
JoshBowdenConcepts: Grid has this reversed as well. I will update both to use this format instead.

JoshBowdenConcepts: Or we can include that in another PR but it should be updated in the same way.

packages/react/src/Bento/Bento.tsx
```diff
@@ -0,0 +1,181 @@
...
+          ),
+        })}
```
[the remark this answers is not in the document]
JoshBowdenConcepts: Should this be the default? I would imagine this should be configured by the user to some degree. If we have this as default then users will need to manage how this will appear on different backgrounds and how they want to manage color modes.

packages/react/src/Bento/Bento.tsx
JoshBowdenConcepts: It was in the wrong spot at line 139 when it should be like 141 if that is what you are referring to. And I will update that in the next push. I will comment on the SHA when completed.

JoshBowdenConcepts:
@danielguillan just to be sure and make this note. The designs show a 12px border radius which would align with our `--brand-borderRadius-large` value. This design will be different with a value of 16px which we don't currently have a borderRadius brand token for.

image

danielguillan:
@JoshBowdenConcepts Interesting. We have `16px` on the [Figma library implementation](https://www.figma.com/file/BJ95AjraesmRCWsKA013GS/Primer-Brand?type=design&node-id=9628-43929&mode=design&t=HGADaDgFRvgF7LlT-11) of Bento, but sticking to existing tokens makes sense to me. Let's use `--brand-borderRadius-large` for now and make sure we update either the tokens or designs with the final decision when @josepmartins is back.

danielguillan:
@JoshBowdenConcepts Interesting. We have `16px` on the [Figma library implementation](https://www.figma.com/file/BJ95AjraesmRCWsKA013GS/Primer-Brand?type=design&node-id=9628-43929&mode=design&t=HGADaDgFRvgF7LlT-11) of Bento, but sticking to existing tokens makes sense to me. Let's use `--brand-borderRadius-large` for now and make sure we update either the tokens or designs with the final decision when @josepmartins is back.

packages/react/src/Bento/BentoItem.features.stories.tsx
JoshBowdenConcepts: This has been updated in this [SHA](https://github.com/primer/brand/pull/390/commits/015e64c3e342f101c896f8259daf2c49feb49558) to use the placeholder image which will have a better contrast ratio.

packages/react/src/Bento/Bento.tsx
JoshBowdenConcepts: This has been updated in the following SHA: https://github.com/primer/brand/pull/390/commits/015e64c3e342f101c896f8259daf2c49feb49558

packages/react/src/Bento/Bento.tsx
JoshBowdenConcepts: This has been updated in the following SHA: https://github.com/primer/brand/pull/390/commits/015e64c3e342f101c896f8259daf2c49feb49558. Now the functionality should behave as I mentioned above.

packages/react/src/Bento/Bento.tsx
JoshBowdenConcepts: Color modes via the ENUM have been added in the following SHA: https://github.com/primer/brand/pull/390/commits/015e64c3e342f101c896f8259daf2c49feb49558

packages/react/src/Bento/Bento.tsx
JoshBowdenConcepts: This has been updated in the following SHA: https://github.com/primer/brand/pull/390/commits/015e64c3e342f101c896f8259daf2c49feb49558

packages/react/src/Bento/Bento.module.css
JoshBowdenConcepts: This has been updated in the following SHA: https://github.com/primer/brand/pull/390/commits/015e64c3e342f101c896f8259daf2c49feb49558

packages/react/src/Bento/Bento.module.css
JoshBowdenConcepts: I found that this broke the working state for both of these components, so I didn't include this update.

packages/react/src/Bento/Bento.tsx
JoshBowdenConcepts: This has been updated in the following [SHA](https://github.com/primer/brand/pull/390/commits/015e64c3e342f101c896f8259daf2c49feb49558)

packages/react/src/Bento/Bento.tsx
JoshBowdenConcepts: This has been added in the following [SHA](https://github.com/primer/brand/pull/390/commits/679e181b3c465e1266854cdea0245a0c2a9378be)

packages/react/src/Bento/Bento.tsx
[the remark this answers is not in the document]
JoshBowdenConcepts: This has been updated in the following SHAs:
https://github.com/primer/brand/pull/390/commits/0a88f2328c740141414e178c219f6821d67de906
https://github.com/primer/brand/pull/390/commits/fedfe4234b082cc90d999ab00972e39402df8db3

packages/react/src/Bento/Bento.features.stories.tsx
JoshBowdenConcepts: This has been updated in this SHA: https://github.com/primer/brand/pull/390/commits/53fb31450e93b3c0e04a852af0a50db660c2c78a

packages/react/src/Bento/Bento.module.css
JoshBowdenConcepts: This has been updated in this SHA: https://github.com/primer/brand/pull/390/commits/015e64c3e342f101c896f8259daf2c49feb49558

packages/react/src/Bento/Bento.tsx
JoshBowdenConcepts: This was updated here: https://github.com/primer/brand/pull/390/commits/679e181b3c465e1266854cdea0245a0c2a9378be. This allows for updating via the Link prop or via CSS if needed. I think this would be the general user expectation.

packages/react/src/Bento/Bento.tsx
JoshBowdenConcepts: This has been updated in this SHA: https://github.com/primer/brand/pull/390/commits/1f0c4ef28849a35edff17f5c182554060fbc5167

packages/react/src/Bento/BentoItem.features.stories.tsx
```diff
@@ -0,0 +1,474 @@
...
+  </Bento.Item>
+)
+
+FlowRowVisualFirst.parameters = {
+  viewport: {
+    defaultViewport: 'iphonexr',
+  },
+}
+
+export const HeadingPaddingCondensed: StoryFn<typeof Bento> = () => (
```
danielguillan, line 127: Should these stories be called `ContentPadding` instead of `HeadingPadding`?

packages/react/src/Bento/Bento.tsx
```diff
@@ -0,0 +1,248 @@
...
+      {children}
+    </div>
+  )
+}
+
+type BentoContentProps = {
+  leadingVisual?: ReactElement | Icon
+  padding?: Padding | ResponsivePadding
+  verticalAlign?: Align | ResponsiveAlign
+  horizontalAlign?: Align | ResponsiveAlign
```
danielguillan, line 116: I don't think Bento should support a `end` horizontal alignment. I can't find any designs that use it or mentions of it in the API exploration.

danielguillan:
The API currently has a limitation that prevents users from defining responsive positioning of the visual. This is because the position is determined by the order of the elements in the JSX.

Take this example: 

| xsmall | >= medium |
|--------|--------|
| screenshot-H6XIAuDg-000661@2x | image | 

The visual here should be at the top of the Bento item when the content is displayed as a column on narrow viewports and at the end when displayed as row on larger viewports. Currently, this isn't possible because the Visual is defined after the content in JSX (i.e., the image appears at the bottom on small viewports).

One potential solution could be to add `column-reverse` and `row-reverse` options to the `flow` prop. Or introduce a dedicated prop to handle this.

packages/react/src/Bento/Bento.features.stories.tsx
```diff
@@ -0,0 +1,171 @@
...
+      <Bento.Content padding={{xsmall: 'normal', small: 'spacious'}} fixedBottomLink>
+        <Bento.Heading as="h4" size="5">
+          Mercado Libre frees developers minds to focus on their missions with GitHub.
+        </Bento.Heading>
+        <Link href="#">Read customer story</Link>
+      </Bento.Content>
+      <Bento.Visual padding="condensed">
+        <img src={mercardo} alt="A white man with grey hair and a beard, running a coffee stand, waving." />
+      </Bento.Visual>
+    </Bento.Item>
```
danielguillan, line 169:
```suggestion
    <Bento.Item
      columnSpan={{xsmall: 12, medium: 7}}
      rowSpan={{xsmall: 3, small: 5}}
      flow={{xsmall: 'row', small: 'row'}}
      colorMode={ColorModesEnum.DARK}
      visualAsBackground
      className={styles['purple-background']}
    >
      <Bento.Content padding={{xsmall: 'normal', small: 'spacious'}} fixedBottomLink leadingVisual={<CopilotIcon />}>
        <Bento.Heading as="h4" size="5">
          Unlocking innovation at scale with AI-driven software development.
        </Bento.Heading>
        <Link href="#">Learn more about Copilot</Link>
      </Bento.Content>
    </Bento.Item>
    <Bento.Item columnSpan={{xsmall: 12, medium: 5}} rowSpan={{xsmall: 3, small: 5}} colorMode={ColorModesEnum.DARK}>
      <Bento.Content padding={{xsmall: 'normal', small: 'spacious'}} horizontalAlign={'center'}>
        <Bento.Heading as="h4" size="1">
          88%
        </Bento.Heading>
        <Text align="center">of developers experience increased productivity.</Text>
      </Bento.Content>
      <Bento.Visual padding="normal" fillMedia={false}>
        <img src={platformAI} alt="Platform Artificial Intelligence Logo" />
      </Bento.Visual>
    </Bento.Item>
    <Bento.Item columnSpan={12} rowSpan={{xsmall: 3, small: 5}} visualAsBackground>
      <Bento.Visual>
        <img
          src={analyzeAndDebug}
          alt="A green and purple gradient screen with a input that has the placeholder, ask a question or type '/' for topics"
        />
      </Bento.Visual>
    </Bento.Item>
    <Bento.Item columnSpan={12} rowSpan={6} flow={{xsmall: 'row', medium: 'column'}} colorMode={ColorModesEnum.DARK}>
      <Bento.Content padding={{xsmall: 'normal', small: 'spacious'}} fixedBottomLink>
        <Bento.Heading as="h4" size="5">
          Mercado Libre frees developers minds to focus on their missions with GitHub.
        </Bento.Heading>
        <Link href="#">Read customer story</Link>
      </Bento.Content>
      <Bento.Visual padding="condensed">
        <img src={mercardo} alt="A white man with grey hair and a beard, running a coffee stand, waving." />
      </Bento.Visual>
    </Bento.Item>
```

Suggesting some updates to better align the examples with the designs (taller Bento items) and improving their responsive layout (stacking items under `medium` breakpoints)

packages/react/src/Bento/Bento.tsx
josepmartins: + 1 on this, no need of `end` here

josepmartins:
> @JoshBowdenConcepts Interesting. We have 16px on the [Figma library implementation](https://www.figma.com/file/BJ95AjraesmRCWsKA013GS/Primer-Brand?type=design&node-id=9628-43929&mode=design&t=HGADaDgFRvgF7LlT-11) of Bento, but sticking to existing tokens makes sense to me. Let's use --brand-borderRadius-large for now and make sure we update either the tokens or designs with the final decision when @josepmartins is back.


Let's use `--brand-borderRadius-large` here, I'll update the Figma designs

> One potential solution could be to add column-reverse and row-reverse options to the flow prop. Or introduce a dedicated prop to handle this.

@rezrah and I initially discarded the `column/row-reverse` options in pro of JSX order, but I agree that's a good usage example.

packages/react/src/Bento/Bento.tsx
```diff
@@ -0,0 +1,248 @@
...
+          size: LeadingVisual['size'] || 44,
+        })}
+      {React.isValidElement(HeadingChild) && React.cloneElement(HeadingChild as React.ReactElement<BentoHeadingProps>)}
+      {React.isValidElement(TextChild) &&
+        React.cloneElement(TextChild as React.ReactElement<TextProps>, {
+          variant: TextChild.props.variant || 'muted',
+          as: TextChild.props.as || 'p',
+          size: TextChild.props.size || '300',
+          className: clsx(styles['Bento_Content-text'], TextChild.props.className),
+        })}
```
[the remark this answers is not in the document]
JoshBowdenConcepts: Hey @danielguillan I am not seeing this. Are you referring to the Item with the 88% heading? If not happy to hop on a quick call at some point to look into it!

packages/react/src/Bento/BentoItem.features.stories.tsx
JoshBowdenConcepts: This has been updated in this [SHA](https://github.com/primer/brand/pull/390/commits/05326e33f19d98a937c0c91d7af92b843b0d1c0c)

packages/react/src/Bento/Bento.module.css
JoshBowdenConcepts: I got this working as the design intended with the --base-size-80! I just had to work with it a bit!

packages/react/src/Bento/Bento.tsx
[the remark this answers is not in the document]
danielguillan: @JoshBowdenConcepts the problem is with the text under the heading. It should use the default variant, not the muted one if there are no `em` elements.

| Current | Expected |
|--------|--------|
| screenshot-Fua18qIz-000666@2x | screenshot-0NH8iLPJ-000668@2x |

ericwbailey:
>@josepmartins I am okay with implementing flow reverse again but let's check with the accessibility team just to make sure we won't have to remove it down the line. It could always be handled by the user if they needed a workaround, instead of us including the logic.

Thanks for thinking of this, and for asking! 

If we reverse content we'll want to write the updated structure to the DOM and not rely on CSS to accomplish this. Essentially, we'll want to ensure focus order matches language reading order, and CSS features like `order` and row-reverse` can break that.

Here's an illustration that hopefully helps to explain:

Wireframe illustration showing an abstract bento box component design with three squares of content. Each box within the bento component has a focusable element in it. Arrows show how focus moves left to right, then top to bottom. The bento box then has its order reconfigured. Focus order is updated to show that left to right, then top to bottom order is preserved.

ericwbailey:
>Does that mean that using the grid order property wouldn't be viable in this situation? 

It might be, depending on the implementation. If y'all have a prototype I could give it a test!

packages/react/src/Bento/Bento.tsx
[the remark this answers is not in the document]
josepmartins: I don't think that the description needs to handle `em` elements. Using the `muted` color as default + providing a token to modify its color should be good enough

danielguillan:
@JoshBowdenConcepts visuals with `spacious` padding do not appear to have a border radius applied, at least in [this Storybook example](https://primer-5574827877-26139705.drafts.github.io/storybook/?path=/story/components-bento-item--visual-padding-spacious).

screenshot-Xh6DPaYF-000682@2x

danielguillan (approved):

apps/docs/content/components/Bento.mdx
```diff
@@ -0,0 +1,227 @@
...
+
+### Bento.Visual
+
+| Name        | Type                                    |   Default   | Required | Description                                                                                              |
+| :---------- | :-------------------------------------- | :---------: | :------- | -------------------------------------------------------------------------------------------------------- |
+| `className` | `string`                                |             | `false`  | Sets a custom class on the root element                                                                  |
+| `fillMedia` | `boolean`                               |   `true`    | `false`  | Determines how the visual reacts within its container. Either filling the container or contained within. |
+| `position`  | `string`                                | `"50% 50%"` | `false`  | The positioning of the visual.                                                                           |
+| `padding`   | `"condensed"`, `"normal"`, `"spacious"` |             | `false`  | Sets a custom class on the root element                                                                  |
```
josepmartins, line 225: The `padding` properties in Visual and Content are responsive right? if that's the case we might want to add the `ResponsivePaddingIndex` option here in both tables

packages/design-tokens/src/tokens/functional/components/bento/base.json
```diff
@@ -0,0 +1,43 @@
+{
+  "brand": {
+    "Bento": {
+      "gap": {
+        "large": {
+          "value": "var(--base-size-32)"
+        },
+        "medium": {
+          "value": "var(--base-size-24)"
```
josepmartins, line 9: Medium should be 32px + not sure we need a large treatment here as they have the same values

```suggestion
          "value": "var(--base-size-32)"
```

packages/design-tokens/src/tokens/functional/components/bento/base.json
```diff
@@ -0,0 +1,43 @@
...
+        }
+      },
+      "item": {
+        "borderRadius": {
+          "small": {
+            "value": "var(--base-size-16)"
+          },
+          "medium": {
+            "value": "var(--base-size-20)"
+          },
```
josepmartins, line 22: Same comment as above, medium should be 24px

```suggestion
          "medium": {
            "value": "var(--base-size-24)"
          },
```

packages/react/src/Bento/Bento.tsx
```diff
@@ -0,0 +1,298 @@
...
+      {React.isValidElement(LinkChild) &&
+        React.cloneElement(LinkChild as React.ReactElement<LinkProps>, {
+          variant: LinkChild.props.variant || 'accent',
+          className: clsx(
+            styles['Bento__call-to-action'],
+            fixedBottomLink && styles['Bento__call-to-action--fixed'],
```
josepmartins, line 213: Why is this prop needed? 

Ideally, the Text description should always have a `margin-bottom: 24` and the Link a `margin-top: auto` and that should work for all cases. I made a codepen to describe this: https://codepen.io/josepmartins/full/YzRBjYV

Without that margin bottom in the description, even if you set `fixedBottomLink` to `true`, you'll get cases where the text + link gets stuck

Screenshot 2023-08-31 at 09 42 54

apps/docs/content/components/Bento.mdx
```diff
@@ -0,0 +1,227 @@
...
+  <Bento.Item columnSpan={3} rowSpan={4} />
+  <Bento.Item columnSpan={3} rowSpan={2} />
+  <Bento.Item columnSpan={6} rowSpan={4} />
+  <Bento.Item columnSpan={3} rowSpan={2} rowStart={5} />
+  <Bento.Item columnSpan={3} rowSpan={4} />
+  <Bento.Item columnSpan={6} rowSpan={2} />
+</Bento>
+```
+
+### Example Layout 1
```
rezrah, line 28:
```suggestion
### Example layout 1
```

apps/docs/content/components/Bento.mdx
```diff
@@ -0,0 +1,227 @@
...
+```jsx live
+<Bento>
+  <Bento.Item columnSpan={12} rowSpan={2} />
+  <Bento.Item columnSpan={7} rowSpan={2} />
+  <Bento.Item columnSpan={5} rowSpan={2} />
+  <Bento.Item columnSpan={12} rowSpan={2} />
+</Bento>
+```
+
+### Example Layout 2
```
rezrah, line 39:
```suggestion
### Example layout 2
```

apps/docs/content/components/Bento.mdx
```diff
@@ -0,0 +1,227 @@
...
+```jsx live
+<Bento>
+  <Bento.Item columnSpan={12} rowSpan={2} />
+  <Bento.Item columnSpan={5} rowSpan={2} />
+  <Bento.Item columnSpan={7} rowSpan={2} />
+  <Bento.Item columnSpan={12} rowSpan={2} />
+</Bento>
+```
+
+### Example Layout 3
```
rezrah, line 50:
```suggestion
### Example layout 3
```

apps/docs/content/components/Bento.mdx
```diff
@@ -0,0 +1,227 @@
...
+```jsx live
+<Bento>
+  <Bento.Item columnSpan={7} rowSpan={2} />
+  <Bento.Item columnSpan={5} rowSpan={2} />
+  <Bento.Item columnSpan={12} rowSpan={2} />
+  <Bento.Item columnSpan={12} rowSpan={2} />
+</Bento>
+```
+
+### Dark Mode Items
```
rezrah, line 61:
```suggestion
### Dark mode items
```

apps/docs/content/components/Bento.mdx
```diff
@@ -0,0 +1,227 @@
...
+
+```jsx live
+<Bento>
+  <Bento.Item columnSpan={3} rowSpan={4} />
+  <Bento.Item columnSpan={3} rowSpan={2} />
+  <Bento.Item columnSpan={6} rowSpan={4} />
+  <Bento.Item columnSpan={3} rowSpan={2} rowStart={5} />
+  <Bento.Item columnSpan={3} rowSpan={4} />
+  <Bento.Item columnSpan={6} rowSpan={2} />
+</Bento>
```
rezrah, line 25: Should the default example just be a single bento item? Or is this particular layout the standard one that we want folks to apply everywhere? 

Screenshot 2023-08-31 at 09 17 32

Related, should Bento.Item span the entire width of the parent container by default, like we do with Grid.Column? Currently doesn't do this, but I think it should 🤔👇 


Screenshot 2023-08-31 at 09 21 50

packages/react/src/Bento/Bento.tsx
```diff
@@ -0,0 +1,298 @@
...
+        className,
+      )}
+      {...rest}
+    >
+      {childrenToRender}
+    </div>
+  )
+}
+
+export const Bento = Object.assign(Root, {Item, Visual, Content, Heading: _Heading})
```
rezrah, line 298: Noticed that we have a namespaced child for headings, but. not the other content. Do we need to namespace it? I can see you're filtering the children for a heading anyway, so could we apply any default props internally to Bento.Content instead? 

Screenshot 2023-08-31 at 09 27 36

packages/react/src/Bento/Bento.tsx
```diff
@@ -0,0 +1,298 @@
...
+        })
+      }
+      return child
+    }
+  })
+  return (
+    <div
+      className={clsx(
+        styles.Bento__Visual,
+        !fillMedia && styles['Bento__Visual-no-fill'],
```
rezrah, line 287: Could we use the standard double dash for modifiers here? `--`

apps/docs/content/components/Bento.mdx
```diff
@@ -0,0 +1,227 @@
...
+  <Bento.Content padding="normal">
+    <Bento.Heading as="h3">Heading</Bento.Heading>
+    <Text>
+      Lorem ipsum dolor sit amet, consectetur adipiscing elit. In sapien sit
+      ullamcorper id. Aliquam luctus sed turpis felis nam pulvinar risus
+      elementum.
+    </Text>
+    <Link href="#">Call to action</Link>
+  </Bento.Content>
+  <Bento.Visual position="50% 100%">
```
rezrah, line 156: Nit.. the position seems redundant here.. recommend removing to keep the example focussed.

```suggestion
  <Bento.Visual>
```

apps/docs/content/components/Bento.mdx
```diff
@@ -0,0 +1,227 @@
...
+    <img
+      className="test"
```
rezrah, line 158: ☝️ can we remove this?

packages/react/src/Bento/Bento.tsx
```diff
@@ -0,0 +1,298 @@
...
+type ResponsiveRowIndex = Partial<Record<Size, number>>
+
+type Flow = 'row' | 'column'
+type ResponsiveFlow = Partial<Record<Size, Flow>>
+
+type Align = 'start' | 'center' | 'end'
+type ResponsiveAlign = Partial<Record<Size, Align>>
+
+type ReducedAlign = 'start' | 'center'
+type ResponsiveReducedAlign = Partial<Record<Size, Align>>
```
rezrah, line 25: The value is the original align, should it be the subset instead?

```suggestion
type ResponsiveReducedAlign = Partial<Record<Size, ReducedAlign>>
```

packages/react/src/Bento/Bento.tsx
```diff
@@ -0,0 +1,298 @@
...
+          className: styles['Bento__Content-icon'],
+          size: LeadingVisual['size'] || 44,
+        })}
+      {React.isValidElement(HeadingChild) && React.cloneElement(HeadingChild as React.ReactElement<BentoHeadingProps>)}
+      {React.isValidElement(TextChild) &&
+        React.cloneElement(TextChild as React.ReactElement<TextProps>, {
+          variant: TextChild.props.variant || 'muted',
+          as: TextChild.props.as || 'p',
+          size: TextChild.props.size || '300',
+          className: clsx(styles['Bento_Content-text'], TextChild.props.className),
```
rezrah, line 206: For consistency, can we use the same naming convention please? Will need to be updated in a few different places.

```diff
- Bento_Content-text
+ Bento__Content-text
```

packages/react/src/Bento/BentoItem.features.stories.tsx
```diff
@@ -0,0 +1,480 @@
...
+    <Bento.Content>
+      <Bento.Heading as="h3">Heading</Bento.Heading>
+      <Text>
+        Lorem ipsum dolor sit amet, consectetur adipiscing elit. In sapien sit ullamcorper id. Aliquam luctus sed turpis
+        felis nam pulvinar risus elementum.
+      </Text>
+      <Link href="#">Call to action</Link>
+    </Bento.Content>
+    <Bento.Visual position="50% 100%">
+      <img className="test" alt="placeholder, blank area with an gray background color" src={placeholderImage} />
```
rezrah, line 50: Seeing `className="test"` in quite a few stories and the docs, do we need those? If not, can we remove them please?

```suggestion
      <img alt="placeholder, blank area with an gray background color" src={placeholderImage} />
```

packages/react/src/Bento/BentoItem.features.stories.tsx
```diff
@@ -0,0 +1,480 @@
...
+VisualPositionBottom.parameters = {
+  viewport: {
+    defaultViewport: 'iphonexr',
+  },
+}
+
+export const FlowColumn: StoryFn<typeof Bento> = () => (
+  <Bento.Item flow="column">
+    <Bento.Content>
+      <Bento.Heading as="h3">Heading</Bento.Heading>
```
rezrah, line 64: Similar comment to the test class, we're setting `as="h3"` manually in a lot of the stories and docs examples, can we remove those please and rely on the default value? It may give the impression that users need to set this each time.

```suggestion
      <Bento.Heading>Heading</Bento.Heading>
```

apps/docs/content/components/Bento.mdx
```diff
@@ -0,0 +1,227 @@
...
+| `flow`               | `"row"`, `"column"`, `ResponsiveFlow`                                                  |           | `false`  | The flow of the `Bento.Content` and `Bento.Visual` within the `Bento.Item`                                                      |
+| `colorMode`          | `"light"`, `"dark"`                                                                    | `"light"` | `false`  | The color mode specified for this `Bento.Item`                                                                                  |
+| `visualAsBackground` | `boolean`                                                                              |  `false`  | `false`  | Determines if the `Bento.Visual` child element is set as the background for the `Bento.Item` with the `Bento.Content` overlayed |
+| `styleOrder`         | `"default"`, `"reversed"`, `ResponsiveStyleOrder`                                      | `default` | `false`  | Determines the visual order for the `Bento.Item`. This changes the semantic ordering of elements for accessibility.             |
```
rezrah, line 189: Should this prop be called order instead of styleOrder, given it isn't just only a visual change as you mention in the description?

packages/react/src/Bento/Bento.features.stories.tsx
```diff
@@ -0,0 +1,175 @@
...
+      flow={{xsmall: 'row', small: 'row'}}
+      colorMode={ColorModesEnum.DARK}
+      visualAsBackground
+      className={styles['purple-background']}
+    >
+      <Bento.Content padding={{xsmall: 'normal', small: 'spacious'}} fixedBottomLink leadingVisual={<CopilotIcon />}>
+        <Bento.Heading as="h4" size="5">
+          Unlocking innovation at scale with AI-driven software development.
+        </Bento.Heading>
+        <Link href="#">Learn more about Copilot</Link>
```
rezrah, line 132: For this example, can we override the variant to achieve higher contrast please?

Screenshot 2023-08-31 at 10 16 20

packages/react/src/Bento/Bento.module.css
```diff
@@ -0,0 +1,1947 @@
...
+
+.Bento__Item {
+  background-color: var(--brand-Bento-bgColor);
+  border-radius: var(--brand-Bento-item-borderRadius-small);
+  display: grid;
+  grid-gap: var(--base-size-32);
+  grid-template-rows: auto 1fr;
+  grid-template-columns: 1fr;
+  overflow: hidden;
+  border: 1px solid var(--brand-color-border-default);
```
rezrah, line 18:
```suggestion
  border: var(--brand-borderWidth-thin) solid var(--brand-color-border-default);
```

packages/react/src/Bento/Bento.module.css
rezrah: I think we need to revisit this soon and find a way to consolidate some of the duplicate styles...

Having a quick look at the css bundle size increase:

Current: 411kb 
In this PR: 467kb 

Result: 13.6% increase being introduced in this PR.

We can internally refactor this after shipping it, but if there's a way to reuse the grid styles and bring this down that would be great 🙏

packages/react/src/Bento/Bento.tsx
```diff
@@ -0,0 +1,298 @@
...
+    ...returnClassBasedOnResponsiveMap('Bento__Item', 'row-start', rowStart),
+    ...returnClassBasedOnResponsiveMap('Bento__Item', 'flow', flow),
+  )
+
+  if (!visualAsBackground) {
+    flow === 'column' && bentoItemClassArray.push(styles['Bento-column-padding-override'])
+    flow === 'row' && bentoItemClassArray.push(styles['Bento-row-padding-override'])
+  }
+
+  const colorModeProp = colorMode ? {'data-color-mode': colorMode} : {}
```
rezrah, line 131: I asked this before, but the conversation thread seems to have gone 🤔 so asking again 😅 ... why do we need to set data-color-mode manually instead of using ThemeProvider, which [supports nesting](https://primer-f4d7c33f83-26139705.drafts.github.io/storybook/?path=/story/components-themeprovider--nested) and can apply the same attribute in a centralised location?

JoshBowdenConcepts:
> I asked this before, but the conversation thread seems to have gone 🤔 so asking again 😅 ... why do we need to set data-color-mode manually instead of using ThemeProvider, which [supports nesting](https://primer-f4d7c33f83-26139705.drafts.github.io/storybook/?path=/story/components-themeprovider--nested) and can apply the same attribute in a centralized location?

Hey Reza, 👋🏽  I am creating this as a comment so that it can persist a bit better hopefully. In regards to your question as you can see below this ThemeProvider creates a div element that is not a bento item.

image

This creates a layout issue that you can see here where the styles are not applied to the right children:

image

I wasn't able to find a way around this directly without creating a property.

packages/react/src/Bento/Bento.tsx
JoshBowdenConcepts: Hey Reza, I answered this here: https://github.com/primer/brand/pull/390#issuecomment-1701107528

packages/react/src/Bento/Bento.tsx
JoshBowdenConcepts: Please let me know if I am missing some piece of information here.

packages/react/src/Bento/Bento.tsx
JoshBowdenConcepts: When we spoke last and were talking about our rules for namespacing. We spoke about using namespacing when there was a significant change to the underlying element that was outside what a prop might accomplish. For instance, a className could be added by looping whereas something more complex like this would be namespaced. 

I am happy to update it if you think that will be easier to maintain, however.

packages/react/src/Bento/Bento.tsx
josepmartins: I believe that can be fixed using `align-self: end/start/center` instead, it overrides the `margin-top: auto` of the link.

I just updated the Codepen with it 👍 : https://codepen.io/josepmartins/full/YzRBjYV

apps/docs/content/components/Bento.mdx
JoshBowdenConcepts: This has been updated in the following [SHA](https://github.com/primer/brand/pull/390/commits/5036e104cc735ec7a6348df403eafc8beaa6416d)

apps/docs/content/components/Bento.mdx
JoshBowdenConcepts: This has been updated in the following [SHA](https://github.com/primer/brand/pull/390/commits/c338ccd5c18e379757a65b3862d659a87bc9ba92)

packages/react/src/Bento/BentoItem.features.stories.tsx
JoshBowdenConcepts: This has been updated in the following [SHA](https://github.com/primer/brand/pull/390/commits/12573536c648bb891828e7aa42347f12b44fb46a)

packages/react/src/Bento/BentoItem.features.stories.tsx
JoshBowdenConcepts: This has been updated in the following [SHA](https://github.com/primer/brand/pull/390/commits/12573536c648bb891828e7aa42347f12b44fb46a)

apps/docs/content/components/Bento.mdx
JoshBowdenConcepts: This has been updated in the following [SHA](https://github.com/primer/brand/pull/390/commits/12573536c648bb891828e7aa42347f12b44fb46a)

packages/react/src/Bento/Bento.tsx
JoshBowdenConcepts: This has been fixed in the following [SHA](https://github.com/primer/brand/pull/390/commits/6b8a1979c1bd4c4c3f8a4bfb1db8b529a2755c2a)

packages/react/src/Bento/Bento.tsx
JoshBowdenConcepts: This has been updated in the following [SHA](https://github.com/primer/brand/pull/390/commits/1f2a7cbc6aab5b74777ef577ee3bb68bc2b7cd3e#diff-9d379f39a3608e961347e076e7bd6000ff5c4ffbe3cbac8186b6565f8831c4f0)

jesussandreas:
i know the bentos are working as intended because i've seen examples of it but i'm not sure we are reflecting these examples on the documentation.

not sure what the solution here is but i think we are not representing bento examples at their best and would love to work with you all on this.

on this [example](https://primer-5574827877-26139705.drafts.github.io/storybook/iframe.html?args=&id=components-bento-features--example&viewMode=story) the container is stretching full width. 

not sure if it's because of the container rule hasn't being applied but the max width should be 1248px (as per our grid default).

borders should be deactivated by default, so perhaps on the documentation. we lead with fill color examples instead or borders? that could make it easier to understand.

the secondary body paragraph, i know this is something that was built into the API, but i would like to keep this as a "hidden feature" if we decide to keep it as is not something we do on the design side. would prefer to lead with headlines instead.

---

**considerations that i would like to see included**

what these might look like when we have an icon or logo?

Screenshot 2023-08-31 at 10 43 18 AM

what if we decide to include a testimonial? although, this last one might be just a nice to have.

Screenshot 2023-08-31 at 10 43 52 AM

jesussandreas:
adding these from a slack thread:

`Fixing the bento content to adapt to different viewport sizes
Adding additional examples of the bento in different viewports (like we do with mobile, but perhaps for tablet?)`

josepmartins:
>
> on this [example](https://primer-5574827877-26139705.drafts.github.io/storybook/iframe.html?args=&id=components-bento-features--example&viewMode=story) the container is stretching full width.
> 
> not sure if it's because of the container rule hasn't being applied but the max width should be 1248px (as per our grid default).
>

It's intended! Bento takes full width of it's container as we expect to be used inside the Grid component. We could wrap Bento examples into a Grid component if that helps  @rezrah @JoshBowdenConcepts 
 
> borders should be deactivated by default, so perhaps on the documentation. we lead with fill color examples instead or borders? that could make it easier to understand.
> 

👍  we could remove the default border style, that aligns to the latest Card changes

> the secondary body paragraph, i know this is something that was built into the API, but i would like to keep this as a "hidden feature" if we decide to keep it as is not something we do on the design side. would prefer to lead with headlines instead.
> 

I'm not sure I'm following here, how do we deal with these examples where there's a headline and a description?

Screenshot 2023-09-01 at 09 14 57

> **considerations that i would like to see included**
> 
> what these might look like when we have an icon or logo?
> 
> Screenshot 2023-08-31 at 10 43 18 AM

Can you provide us an example? Logo or icons can be used in the Visual area as images/svgs btw

> what if we decide to include a testimonial? although, this last one might be just a nice to have.
> 
> Screenshot 2023-08-31 at 10 43 52 AM

If that's needed, I believe we could include the Testimonial component as an accepted child of the Bento item

packages/react/src/Bento/Bento.tsx
JoshBowdenConcepts: I see how you accomplished this using flex rather than a grid. This will change the logic needed to create a background image the way mine currently layers. Let me see if I can accomplish this another way before refactoring.

JoshBowdenConcepts: I was able to solve this while keeping grid for the overlay content. https://github.com/primer/brand/pull/390/commits/c9a828d8cce5d75405cdcc1df581478f7bca0665

JoshBowdenConcepts:
@josepmartins @jesussandreas I think we can handle the max-width property internally rather than wrapping via a grid because that would add extra logic internally.

As for the icon and the testimonial. We can currently include an Icon but not with the spacing you have shown and we currently limit the content within based on Josep's API exploration. I think we could add these features down the line for sure! I think we want to move this forward to get it used by @enstyled ASAP. As for the documentation, I would definitely update it to better represent what can be done. @jesussandreas would you like to pair today to show me what you were thinking I will update the docs to provide better examples. 

As for the text under the heading. I think the best we can do to hide this for now would be to not show examples of us including this. Would that work for you?

josepmartins:
> Just to clarify another decision here: We are removing the default border in favor of including a class for background color only in the docs. when implemented the background color will default to the default background color which is white in light mode and a dark color in dark mode.

Yep, I'll open a new branch in Figma to remove the border styles as well

ajashams:
### Body copy in bentos
This is using `Heading 6` in `medium`, so just confirming here that we're not supporting any `body` styles.
| image |
| --- | 

### Removing border from bentos
Yes! Thank you for the change. 

### Logos + Icons used in image box
@josepmartins you asked for an example. Below are two versions where we want to float the icon and logo in a bento's image container. It's my assumption that the circle around the icons are `optional`. In the case of enterprise, we want to be able to animate this by drawing it in. It's almost like a progress bar or percentage. CC @enstyled here as I know he's already shared similar feedback. 

| Screenshot 2023-09-05 at 4 23 06 PM | Screenshot 2023-09-05 at 4 23 13 PM | Screenshot 2023-09-05 at 4 23 20 PM
| --- | --- | --- |

### Testimonial bento
Agree with @JoshBowdenConcepts on this one. Let's hold on using the testimonial within the component until we've stress tested that use case. That'll unblock this work for @enstyled to leverage. 

### Mobile
Can we take another pass at the mobile layouts? I'm noticing some condensed boxes and cut off text. Probably best to do another round of that before we provide feedback. 🙏 

| Screenshot 2023-09-05 at 4 32 43 PM | Screenshot 2023-09-05 at 4 35 07 PM

PR #401, opened 2023-08-21
Fixes type scale on narrow viewports

danielguillan:
Fixes duplicate type scale sizes on mobile viewports, as pointed out by @jesussandreas on the Figma implementation.

### DIFF
--- a/.changeset/chilly-badgers-melt.md
+++ b/.changeset/chilly-badgers-melt.md
@@ -1,0 +1,5 @@
+---
+'@primer/brand-primitives': patch
+---
+
+Fixes duplicate typographic scale sizes on small viewports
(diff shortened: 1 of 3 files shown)


.changeset/chilly-badgers-melt.md
```diff
@@ -0,0 +1,5 @@
+---
+'@primer/brand-primitives': patch
+---
+
+Fixes duplicate typographic scale sizes on small viewports
```
rezrah, line 5: Can we remove the changeset on this occasion please. The global typography change hasn't been released yet, so this fix can piggyback off that existing one in `main`.

packages/react/src/Avatar/Avatar.visual.spec.ts-snapshots/Visual-Comparison-Avatar-Avatar-Default-1-linux.png
rezrah, line 1: Could this one be deleted please? Looks like the avatar didn't load in time.

jesussandreas (commented):
thank you all!

PR #442, opened 2023-10-03
Remove `hsl` and use Primitive v8 colors

langermank:
## Summary

As an initial step to upgrade `primer/primitives` to v8 tokens, this PR moves the existing color scales from `hsl` to `hex` for an easier transition.

## List of notable changes:

- Change base color scales to `hex` instead of `hsl`
- Remove the `hsl` logic that handles opacity
- Use a raw hex8 value where alpha was used temporarily until we have the new build process in place from Primitives
- Adds new tokens for `videoPlayer` to help with this transition
- Use two files for light/dark vs. one in the base tokens

Did we break anything? 😄 

Closes https://github.com/github/primer/issues/2756

## Contributor checklist:

## Reviewer checklist:


packages/react/src/animation/Animation.stories.module.css
```diff
@@ -16,11 +16,7 @@
   inset: var(--base-size-16);
   filter: blur(var(--base-size-48));
   transform: translateZ(0);
-  background: linear-gradient(
-    to right,
-    hsl(var(--base-color-scale-blue-2-hsl)),
-    hsl(var(--base-color-scale-purple-2-hsl))
-  );
+  background: linear-gradient(to right, hsl(208 100% 74%), hsl(269 100% 83%));
```
rezrah, line 19: Did this one still need to use HSL to work, or can we switch to using the hex base tokens?

packages/react/src/component-helpers.tsx
```diff
@@ -26,7 +26,7 @@ export function RedlineBackground({height, hasBorder = true, ...rest}: PropsWith
         overflow: 'hidden',
         border: hasBorder ? '1px solid var(--base-color-scale-red-2)' : undefined,
         backgroundImage:
-          'linear-gradient(45deg, var(--base-color-scale-red-0) 12.5%, hsla(var(--base-color-scale-red-2-hsl) / 50%) 12.5%, hsla(var(--base-color-scale-red-2-hsl) / 50%) 50%, var(--base-color-scale-red-0) 50%, var(--base-color-scale-red-0) 62.5%, hsla(var(--base-color-scale-red-2-hsl) / 50%) 62.5%, hsla(var(--base-color-scale-red-2-hsl) / 50%) 100%)',
+          'linear-gradient(45deg, var(--base-color-scale-red-0) 12.5%, #ffb5b380 12.5%, #ffb5b380 50%, var(--base-color-scale-red-0) 50%, var(--base-color-scale-red-0) 62.5%, #ffb5b380 62.5%, #ffb5b380 100%)',
```
rezrah, line 29: These hex values are a fair bit harder to read IMO. I think we're losing readability and ease of correlation with base color scale tokens. 

Some ideas 💡
 
- As this is a JS file, how about we create a helper function that generates the final output, by using a named token as a parameter?
- Create new and locally scoped tokens with associated comments on what these hex values are and how they were created for future reference

packages/react/src/CTABanner/CTABanner.visual.spec.ts-snapshots/Visual-Comparison-CTABanner-CTABanner-Playground-1-linux.png
rezrah, line 1: Wonder why the border radius is changing here 🤔

packages/design-tokens/src/tokens/functional/components/subdomain-nav-bar/colors.js
```diff
@@ -3,8 +3,8 @@ module.exports = {
     SubdomainNavBar: {
       canvas: {
         default: {
-          value: 'hsl(var(--base-color-scale-white-0-hsl) / 50%)',
-          dark: 'hsl(var(--base-color-scale-gray-9-hsl) / 50%)',
+          value: '#ffffff80',
+          dark: '#0d111780',
```
rezrah, line 7: Could we add a TODO here to handle this via a StyleDictionary transformer? If the gray-9 value changes, it won't get reflected here, so would be better to handle this via interpolation in the long term. An annotation here would help remind us to do this (and also clarify this is gray-9 at 50%)

langermank: Remember this is temporary, so the new build process already handles this. It looks something like this:

```
hover: {
          $value: '{base.color.gray.2}',
          $type: 'color'
          },
          alpha: 0.2,
        },
```

Good idea to annotate 👍 will do that to make it easier to upgrade later

langermank, line 7:
```suggestion
          value: '#ffffff80', // base.color.white 0.5 alpha
          dark: '#0d111780', // base.color.gray.9 0.5 alpha
```

packages/react/src/component-helpers.tsx
langermank: Since this is just for presentation and not product UI, I wonder if we should hold off for now and add some documentation specific tokens when we have access to the new build with alpha logic later? Otherwise we can just add them now and reference the hex, but I don't think that should be a blocker for this PR IMO.

packages/react/src/CTABanner/CTABanner.visual.spec.ts-snapshots/Visual-Comparison-CTABanner-CTABanner-Playground-1-linux.png
langermank: No idea.. but I just compared it to production, and the updated screenshot looks correct 🤷

packages/react/src/AnchorNav/AnchorNav.module.css
```diff
@@ -331,7 +331,8 @@
     width: 100%;
     height: 100vh; /* Keep as fallback */
     height: 100dvh;
-    background: hsla(var(--base-color-scale-black-0-hsl) / 0.5);
+    /* replace with overlay-backdrop-bgColor */
+    background: #8c959f33;
```
rezrah, line 335: This is the only one i'm seeing a regression on 👇 

Screenshot 2023-10-30 at 14 23 59

Think we can reduce opacity for time being. I'll push the change...

langermank: Yeah, I sort of on a whim aligned this closer to Primer product design tokens but we can totally revert this change

rezrah: 👍 Yeah if that's okay, let's go back to old one for now to minimise visual regression as much as possible. I've reverted this to black at 50% now (still as a hex value)... We can put in a separate PR down the line to change the visuals.

rezrah (approved):
Looks great.. visual regressions appear to be minimal.

Nice work @langermank ✨ 👏 . Ship it!

packages/react/src/component-helpers.tsx
rezrah: doc tokens sound good, yeah let's leave as is for now 👍

PR #500, opened 2023-11-16
Update index.mdx

nsolerieu:
## Summary

Suggesting to fix the image for the river stack "Don't" section to match actual use case since we do use long stack of river on the same side but we do NOT want to create asymmetry in river Units.

## List of notable changes:

Caption of the image and would need help to update the image too (I don't know how to upload it to update the URL int eh markedown)

river-stack-dont

## Steps to test:

None -simple text and image update

### DIFF
--- a/apps/docs/content/components/River/index.mdx
+++ b/apps/docs/content/components/River/index.mdx
@@ -10,7 +10,7 @@ export default ComponentLayout
 
 ## Usage
 
-Use rivers to showcase features and introduce topics. Rivers are composed of images or videos paired with text content. River's text content includes a heading, body text, and an optional link.
+Use rivers to showcase features and introduce topics. Rivers are composed of images or videos paired alongside text content like headings, paragraphs, and links.
 
 A river’s content should be short and concise, no longer than 3 or 4 sentences and in a single paragraph when possible.
 
--- a/apps/docs/content/components/River/index.mdx
+++ b/apps/docs/content/components/River/index.mdx
@@ -49,7 +49,7 @@ Two or more rivers can be stacked to guide the user through a set of feautres. W
       src="https://github.com/primer/brand/assets/6951037/169d5c19-67f4-4f15-8e3b-80c877b34af1"
     />
     <Caption>
-      River component stacks create a unit with coherent flow. Here the last item breaks the flow.
+      Don't break the flow in a stack of rivers that would otherwise stay aligned.
     </Caption>
   </Dont>
 </DoDontContainer>
--- a/apps/docs/content/components/River/index.mdx
+++ b/apps/docs/content/components/River/index.mdx
@@ -110,7 +110,7 @@ A river breakout component has an image that fills the width of the container wh
 
 This is useful when you want to create a highlight which can be used to create more emphasis or simply break a monotonous vertical flow of left/right rivers. It can also be helpful on smaller viewports to avoid stacking two images or two chunks of text from adjacent river components. 
 
-As it's name suggest, a river breakout should not be the first in the stack, it should be in-between two left/right rivers.
+As its name suggest, a river breakout should not be the first in the stack, it should be positioned in-between two standard rivers.
 
 Some examples of when to use a river breakout:
 
--- a/apps/docs/content/components/River/index.mdx
+++ b/apps/docs/content/components/River/index.mdx
@@ -49,8 +49,7 @@ Two or more rivers can be stacked to guide the user through a set of feautres. W
       src="https://github.com/primer/brand/assets/912236/ff254125-73ae-4838-9ee9-32aac6bcf19b"
     />
     <Caption>
-      Don't stack numerous rivers with 40:60 ratio without alternating left and
-      right alignments.
+      River component stacks create a unit with coherent flow. Here the last item breaks the flow.
     </Caption>
   </Dont>
 </DoDontContainer>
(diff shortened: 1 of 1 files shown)


nsolerieu:
Once I know how to update image I'll update this (text overlapping with the right side visual)

Screenshot 2023-11-15 at 4 38 55 PM

apps/docs/content/components/River/index.mdx
```diff
@@ -46,11 +46,10 @@ Two or more rivers can be stacked to guide the user through a set of feautres. W
   <Dont>
     <img
       alt=""
-      src="https://github.com/primer/brand/assets/912236/ff254125-73ae-4838-9ee9-32aac6bcf19b"
+      src="https://github.com/primer/brand/assets/6951037/169d5c19-67f4-4f15-8e3b-80c877b34af1"
     />
     <Caption>
-      Don't stack numerous rivers with 40:60 ratio without alternating left and
-      right alignments.
+      River component stacks create a unit with coherent flow. Here the last item breaks the flow.
```
rezrah, line 52: Minor suggestion around active voice...

```suggestion
      Don't break the flow in a stack of rivers that would otherwise stay aligned.
```

rezrah (approved):
LGTM 👍, nice one @nsolerieu ✨

apps/docs/content/components/River/index.mdx
```diff
@@ -10,9 +10,9 @@ export default ComponentLayout
 
 ## Usage
 
-Use rivers to showcase features and introduce topics. Rivers compose images or videos and text together. River's text includes a heading, body text, and an optional link.
+Use rivers to showcase features and introduce topics. Rivers are composed of images or videos paired with text content. River's text content includes a heading, body text, and an optional link.
```
rezrah, line 13: Suggestion for brevity...

```suggestion
Use rivers to showcase features and introduce topics. Rivers are composed of images or videos paired alongside text content like headings, paragraphs, and links.
```

apps/docs/content/components/River/index.mdx
```diff
@@ -103,13 +104,15 @@ A single river component acts as a section and it should not include any child s
   </Dont>
 </DoDontContainer>
 
-### Breakout river
+### River Breakout
```
rezrah, line 107:
```suggestion
### River breakout
```

apps/docs/content/components/River/index.mdx
```diff
@@ -103,13 +104,15 @@ A single river component acts as a section and it should not include any child s
...
 
 A river breakout component has an image that fills the width of the container while providing more space for content such as text and lists.
 
-This is useful when you want to create a highlight which can be used to create more emphasis or simply break a monotonous vertical flow of left/right rivers. It can also be helpful on smaller viewports to avoid stacking two images or two chunks of text from adjacent river components.
+This is useful when you want to create a highlight which can be used to create more emphasis or simply break a monotonous vertical flow of left/right rivers. It can also be helpful on smaller viewports to avoid stacking two images or two chunks of text from adjacent river components. 
+
+As it's name suggest, a river breakout should not be the first in the stack, it should be in-between two left/right rivers.
```
rezrah, line 113: Such a good point, I only just realised this 😅 😆

rezrah, line 113:
```suggestion
As its name suggest, a river breakout should not be the first in the stack, it should be positioned in-between two standard rivers.
```

On narrow viewports, there is no left/right, so suggesting that we tweak wording to reference it as the default / standard one.

apps/docs/content/components/River/index.mdx
```diff
@@ -174,6 +177,12 @@ Always use high quality images and videos. Avoid using images with text or singl
 
 The image is automatically styled to fit the width of the parent container. If you have bespoke styling requirements, you can disable the provided default styles and style the image manually.
 
+### Additional guidance
+
+- We recommend using a river left as the first river of the flow to start with a visual and avoid stacking text with above section on mobile
```
rezrah, line 182: Nit. In RTL reading directions, this would be right-aligned, so could we use directionally-agnostic terms here if possible. E.g. start-aligned if you're referring to 'left'. This is going to become more important as over time, as we expand on i18n.

Could this guidance go into the stacking section too? Feels kind of relevant to that, unless I've misinterpreted it.

apps/docs/content/components/River/index.mdx
```diff
@@ -174,6 +177,12 @@ Always use high quality images and videos. Avoid using images with text or singl
...
+- River components stacks create a unit, the set should be considered to be the main parent section. The parent section should be spaced as a regular section with the largest spacing while internal elements should use smaller spacing
+- If the image is critical for the user to understand the message, default to right aligned. If the text is critical for the user to understand the image, choose left aligned.
```
rezrah, line 184: Fine like this for now, but I do think these would make for some **great** visual do/don't examples that can sit under the stacking section above.

apps/docs/content/components/River/index.mdx
nsolerieu: I see what you mean by directionally agnostic. I was intending to only talk about the layout and not the text. I don't see any issue with this advice being applied in a RTL language context. Perhaps a different wording could solve this (will make a suggestion)

apps/docs/content/components/River/index.mdx
nsolerieu: Ya, that last section `Additional guidance` feels a bit too abstract and could help some definition via a few visuals and be meshed with the other section. Will do another PR to address this.

auareyou (approved):
It looks good @nsolerieu!

Issue #404, opened 2023-08-24
[Accessibility] Add a "Skip To Content" button to `SubdomainNavBar`


simurai:
**Note**: This is not really urgent, but still thought it might be a "nice to have" at some point, since I assume we will run into it for every site/page we'll build.

---

This came up in https://github.com/github/accessibility-audits/issues/5220, where we got asked to add a "Skip To Content" button that is shown when a user uses the `tab` key to navigate and wants to skip having to go through all the options of the nav (`SubdomainNavBar`).

![Button in the top left corner with a Skip To Content label](https://github.com/github/githubuniverse.com/assets/378023/3303e661-3547-46a0-9edf-3f920398e041)

Here is the [implementation for Universe](https://github.com/github/githubuniverse.com/pull/1805). The button could be the first item in the `SubdomainNavBar`, but not sure about adding an `id="start-of-content"` that the button links to. For Universe we added it on the following [`<main>` element](https://github.com/github/githubuniverse.com/pull/1805/files#diff-273e11bd53a1ddfe40b66fd62d9b9b70337ab54bd39ab1718f43e8999198aee1R32). But ideally that should happen automatically without the need to remember it. Hmm..  Maybe since it's essentially just a "skip the nav", we might can also add an empty `<div id="start-of-content"` at the very end of `SubdomainNavBar`? 🤔

stamat:
@simurai I was gonna say that we should have a main element too, maybe we should introduce it to dotcom marketing pages. Otherwise we could query the first paragraph maybe? Or since we now use React maybe find an element that we can latch onto. Or maybe we detect the hero and then when you click we scroll down under the hero. I'll give it a thought, and I'll report back!
Maybe there is some noninvasive way to do this. But the logic should be run on click event.

stamat:
@simurai after looking at it... And realizing that skiping to content actually means skipping the header menu lol, I was about to suggest the same:

> Maybe since it's essentially just a "skip the nav", we might can also add an empty <div id="start-of-content" at the very end of SubdomainNavBar?

And the code already does this if the `main` with the ID is not available! Man I gotta update myself on React lol...

simurai:
> @stamat And realizing that skiping to content actually means skipping the header menu

Yeah, I think it's mostly for users that already know the nav and don't want to tab through it every time they land on a page. So instead of having to `tab tab tab tab...` they can just do `tab enter` and skip the nav.

> And the code already does this if the `main` with the ID is not available!

[...]

PR #503, opened 2023-11-22
add SkipToContent button to SubdomainNavBar

joseph-lozano:
## Summary

Adds a `SkipToContent` button to the `SubdomainNavBar`

## List of notable changes:

- Adds a `SkipToContent` button to the `SubdomainNavBar`

Does the button work? 
Does it properly hide itself when not in focus?
Does it jump to the content of the page.

## Steps to test:

You can test this component on any of the SubdomainNavBar stories.

Solves #404 

## Contributor checklist:

## Reviewer checklist:

## Screenshots:

<table>
<tr>
</tr>
<tr>
<td valign="top">

 </td>
<td valign="top">

</td>
</tr>
</table>


packages/react/src/SubdomainNavBar/SubdomainNavBar.module.css
```diff
@@ -649,6 +649,24 @@
   color: var(--base-color-scale-black-0);
 }
 
+.SubdomainNavBar-skip-to-content {
+  position: fixed;
+
+  /* Visually center button (header 75px - button 49px) / 2 = 13px */
+  top: 13px;
+  left: 13px;
```
rezrah, line 657: Does it need to be 13px or can you use `var(---base-size-12)` here? We typically avoid hard coding px values and prefer rem.

packages/react/src/SubdomainNavBar/SubdomainNavBar.module.css
```diff
@@ -649,6 +649,24 @@
...
+  z-index: 100;
+}
+
+.SubdomainNavBar-skip-to-content:not(:focus) {
+  width: 1px;
+  height: 1px;
+  padding: 0;
+  overflow: hidden;
+  clip: rect(1px, 1px, 1px, 1px);
+  border: 0;
```
rezrah, line 667: This is just visually hiding it, right? If so, could this be done programmatically by adding the global `visually-hidden` class?

packages/react/src/SubdomainNavBar/SubdomainNavBar.tsx
```diff
@@ -299,12 +268,61 @@ function Root({
...
+                </div>
+              )}
+            </div>
           </div>
-        </div>
-      </header>
-    </div>
+        </header>
+      </div>
+      <div id="start-of-content" />
```
rezrah, line 324: Do avoid any id clashes on end-user side, could we make this id unique please? Some precedents in other components via `useId` hooks.

packages/react/src/SubdomainNavBar/SubdomainNavBar.tsx
```diff
@@ -135,140 +135,109 @@ function Root({
...
-              </NavigationVisbilityObserver>
+    <>
+      <Button
+        as="a"
+        href="#start-of-content"
+        variant="primary"
+        className={clsx(styles['SubdomainNavBar-skip-to-content'])}
+      >
+        Skip to content
+      </Button>
```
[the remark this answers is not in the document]
rezrah: cc. @nsolerieu @natalie-iv

nsolerieu:
@joseph-lozano I'd suggest use banner stretching under the header to avoid having the button awkwardly overlap the header UI like this:

Screenshot 2023-12-04 at 11 22 49 AM

Since this would be a custom additions we could consider using our button component and stretch it full width (like Linear does). Since our CTA component is very chunky it would look like this - ideally we'd want to remove the border radius for it too look more like the banner above.

Screenshot 2023-12-04 at 11 26 47 AM

Happy to see/review whatever is feasible here

packages/react/src/SubdomainNavBar/SubdomainNavBar.tsx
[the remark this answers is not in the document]
rezrah: Discussed at maintainer sync. We'll move forward with this design as an interim-step, with @auareyou exploring more integrated presentation down the line.

packages/react/src/SubdomainNavBar/SubdomainNavBar.tsx
```diff
@@ -16,6 +16,7 @@ import '@primer/brand-primitives/lib/design-tokens/css/tokens/functional/compone
 
 /** * Main Stylesheet (as a CSS Module) */
 import styles from './SubdomainNavBar.module.css'
+import {useId} from '@reach/auto-id'
```
rezrah, line 19: Nit: We have our own internal `useId` hook btw, which might be worth using here. Right now takes a dependency on `'@reach/auto-id'`, but helps centralise usage of it so we can find and replace them more easily in future.

packages/react/src/SubdomainNavBar/SubdomainNavBar.tsx
```diff
@@ -135,140 +138,111 @@ function Root({
...
-                {menuItems}
-              </NavigationVisbilityObserver>
+    <>
+      <Button
+        as="a"
+        href={`#${startOfContentID}`}
+        variant="primary"
+        className={clsx(styles['SubdomainNavBar-skip-to-content'], !startOfContentButtonFocused && 'visually-hidden')}
+        onFocus={() => setStartOfContentButtonFocused(true)}
+        onBlur={() => setStartOfContentButtonFocused(false)}
```
rezrah, line 148: Possible to extract these functions outside of the return statement please? Potentially wrapping in a `useCallback`? Micro-perf optimisation to avoid creating a new function each time. I missed this last time, sorry.

rezrah (approved):
LGTM. Added some more minor feedback, otherwise good to go :shipit:

Nice work @joseph-lozano ✨

joseph-lozano:
After chatting at the Accessiblity Eng+ Office Hours, it sounds like we should first look for a `<main>` element before falling back to creating an empty div for the start-of-content.

Additionally, if we create the `start-of-content` empty div, we should make sure to give it `tabindex="-1"`.

https://github.com/github/accessibility/issues/5240

packages/react/src/SubdomainNavBar/SubdomainNavBar.tsx
joseph-lozano: Everywhere else we are using `useId` it imports from `@react/auto-id`

packages/react/src/SubdomainNavBar/SubdomainNavBar.tsx
```diff
@@ -81,11 +82,21 @@ function Root({
...
   const handleMobileMenuClick = () => setMenuHidden(!menuHidden)
   const handleSearchVisibility = () => setSearchVisible(!searchVisible)
   const focusTrapRef = useRef<HTMLDivElement | null>(null)
 
+  useEffect(() => {
+    const main = document.getElementsByTagName('main')[0] as HTMLElement | undefined
+    if (!main) return
+    if (!main.id && startOfContentID) main.id = startOfContentID
+    setMainTag(main)
+  }, [setMainTag, startOfContentID])
```
joseph-lozano, line 98: @rezrah this seems... somehow wrong, but unsure how else I might satisfy the a11y team's recommendation. Do you have any thoughts?

rezrah: I wouldn't recommend storing the element in state, if that's what you're asking?That's really heavy and `refs` are better for this IMO.

This is just pseudocode so you'll need to try it, but can you do something like:

```jsx
const mainElRef = useRef<HTMLElement | null>(null);

useEffect(() => {
    const main = document.getElementsByTagName('main')[0] as HTMLElement | undefined
    
  if(el and otherStuff) {
    mainElRef.current = el;
  }

}, []);
```

stamat: @rezrah my man, I think this is GTG ☺️ , except these [two testing stories](https://github.com/primer/brand/blob/ddd5b28e4aee6835c25b4b2681768bda29987214/packages/react/src/SubdomainNavBar/SubdomainNavBar.stories.tsx#L643-L690) related to this update are giving AXE errors in CI! Would you know why it is failing? My AXE tooling is showing 0 issues even on strict. And running `test:axe` doesn't work on Codespaces, so I'll have to try this on my local.

stamat:
@rezrah thanks for helping me sort out running the `test:axe`. It turned out that it was missing a directory so `mkdir apps/docs/public/static` which was breaking the run of `npm run build:prod --workspace=apps/storybook`.  So adding the empty dir like you said made it run.

But then this wasn't what I needed, what I needed was to get that `storybook-static` directory. So you helped me by mentioning `npx storybook build` so I did this:
```bash
cd apps/storybook
npm i # which I think doesn't do anything cause node_modules are in the root directory
npx storybook build
```
This gave me the `storybook-static` directory with the `stories.json` that is needed for the `packages/e2e/scripts/playwright/axe-clean.spec.ts` to run when called by `npm run test:axe --workspace=packages/e2e` since this was the CI that was failing and since we don't list all of the violations, I couldn't know this before I got this working locally. **I offer to try to write the code for listing the violations, shouldn't be that hard ☺️, so we can see directly from the CI what AXE is telling us. 🪓 **

So I curated the `stories.json` to run the two stories in question and found out what it was complaining about, which is this 👇 

Screenshot 2024-03-05 at 16 03 51

These stories include `<main>` tags which automatically have the `role="main"` and AXE doesn't allow us to have any other role for the `<main>` tag. But also we have a default Storybook `div role="main"` which was killing our buzz. 👇 

Screenshot 2024-03-05 at 16 16 27

**I wonder if we can load the desired story independently of the story chooser. I think we should be able to do so. I'll try to update the code.**

But as the instant remedy I added these two stories in question to the ignore list. Since they are just for testing the ability to scroll to `<main>`.

P.S. The reason I needed to curate the `stories.json` is because I was getting an AXE violation for this 👇 😆 It's a banner telling us to update storybook. I think loading the stories independently of the chooser will ignore this violation.

Screenshot 2024-03-05 at 15 56 39

PR #507, opened 2023-11-29
Rename `Image` token + refactor CSS

langermank:
Part of https://github.com/github/primer/issues/2097

Tokens renamed:
--brand-Image-ratio-custom → --brand-image-aspectRatio

In order for this name to make sense/fit our convention, I modified the CSS to reference the token and then reset the value for each prop.

### DIFF
--- a/.changeset/perfect-boxes-fetch.md
+++ b/.changeset/perfect-boxes-fetch.md
@@ -1,6 +1,6 @@
 ---
-"@primer/brand-primitives": patch
-"@primer/react-brand": patch
+"@primer/brand-primitives": minor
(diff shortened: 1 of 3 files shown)


.changeset/perfect-boxes-fetch.md
```diff
@@ -0,0 +1,6 @@
+---
+"@primer/brand-primitives": patch
+"@primer/react-brand": patch
```
rezrah, line 3:
```suggestion
"@primer/brand-primitives": minor
"@primer/react-brand": minor
```

PR #515, opened 2023-12-18
Remove unused tokens

langermank:
## Summary

Part of https://github.com/github/primer/issues/2097

I've got a lot of changes coming soon, so just removing anything that's not actually used to keep things cleaner. We have a better strategy in place for these kinds of generic tokens, but I haven't run into a use case to add them here.

## List of notable changes:

Removed unused tokens:
```
--brand-animation-variant-scaleInTop-distance
--brand-color-success-emphasis
--brand-color-success-muted
--brand-color-success-subtle
--brand-color-error-emphasis
--brand-color-error-muted
--brand-color-error-subtle
--brand-color-neutral-emphasisPlus
--brand-color-neutral-emphasis
--brand-color-neutral-muted
--brand-color-accent-secondary
```

## Contributor checklist:

## Reviewer checklist:

### DIFF
--- a/packages/design-tokens/src/tokens/functional/colors/global.json
+++ b/packages/design-tokens/src/tokens/functional/colors/global.json
@@ -5,46 +5,18 @@
         "fg": {
           "value": "var(--base-color-scale-green-5)",
           "dark": "var(--base-color-scale-green-3)"
-        },
-        "emphasis": {
-          "value": "var(--base-color-scale-green-4)",
-          "dark": "var(--base-color-scale-green-5)"
-        },
-        "muted": {
-          "value": "#4bc36b99",
-          "dark": "#2e9e4299"
-        },
-        "subtle": {
-          "value": "var(--base-color-scale-green-0)",
-          "dark": "#2e9e42d9"
         }
       },
       "error": {
         "fg": {
           "value": "var(--base-color-scale-red-5)",
           "dark": "var(--base-color-scale-red-4)"
-        },
-        "emphasis": {
-          "value": "var(--base-color-scale-red-5)",
-          "dark": "var(--base-color-scale-red-5)"
-        },
-        "muted": {
-          "value": "#ff878599",
-          "dark": "#f8524999"
-        },
-        "subtle": {
-          "value": "var(--base-color-scale-red-0)",
-          "dark": "#f85249d9"
         }
       },
       "accent": {
         "primary": {
           "value": "var(--base-color-scale-pink-5)",
           "dark": "var(--base-color-scale-pink-3)"
-        },
-        "secondary": {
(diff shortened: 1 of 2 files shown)


danielguillan (approved):
Looks good to me. However, I'm unsure if the `--brand-animation-variant-scaleInTop-distance` was intentionally left unused or not. Either way, we can bring it back if needed later.

rezrah:
While these tokens aren't used internally, I'd say there's a high likelihood that they're referenced by the end-user somewhere. E.g. I know that we reference `--brand-color-success-*` in a few marketing pages, and in the upcoming brand toolkit. 

Removing these tokens technically constitutes a breaking change, so wondering:

a) do we need to do this right now? Can it be deferred and batched with the other upcoming breaking changes?
b) can we provide alternatives for folks that are currently relying on those tokens?

langermank:
@rezrah what is the plan for upgrading packages that consume Primer Brand tokens when we complete the mass renaming? I think the last time we talked about it, you weren't concerned about custom CSS usage. 

Since I'm adding a stylelint config/plugin to handle find and replace for all the other tokens, I can do the same for these. But it would be helpful if I can see where they are used to best recommend a replacement token.

rezrah:
> @rezrah what is the plan for upgrading packages that consume Primer Brand tokens when we complete the mass renaming?

For tokens, a combination of stylelint codemods and/or written migration guides in the changelog should be enough, I think 👍 

[Here's a very recent example](https://github.com/github/brand.github.com/blob/main/pages/foundations/mascots.mdx#L63) of a functional token being used consumer-side.

> Since I'm adding a stylelint config/plugin to handle find and replace for all the other tokens, I can do the same for these.

🙏 That sounds great. If the configs aren't ready / available.. could also just expand the changelog entry to include a migration `diff`? Should be examples from previous deprecations in there already. If we don't have alternative tokens, we could also ask users to create local replacements for now. Just a one-liner would be fine. WDYT.

Leaving approval in meantime, as I'll be out until the new year.

rezrah (approved):

packages/design-tokens/src/tokens/functional/colors/global.json
```diff
@@ -5,46 +5,18 @@
...
         }
       },
       "accent": {
         "primary": {
           "value": "var(--base-color-scale-pink-5)",
           "dark": "var(--base-color-scale-pink-3)"
-        },
-        "secondary": {
-          "value": "var(--base-color-scale-purple-5)",
-          "dark": "var(--base-color-scale-purple-3)"
```
rezrah, line 47: [Looks like we may need additional accent color variables in future](https://turbo-adventure-73kwy3g.pages.github.io/foundations/color) to support our brand pillars, so worth keeping this around?

PR #536, opened 2024-02-19
Add interface guidelines for section intro content length

danielguillan:
Adds content length guidance and do's and don'ts for section intro.

[Docs preview](https://primer-72c54b3967-26139705.drafts.github.io/components/SectionIntro)

### DIFF
--- a/apps/docs/content/components/SectionIntro/index.mdx
+++ b/apps/docs/content/components/SectionIntro/index.mdx
@@ -45,7 +45,7 @@ Ensure the description is concise, keeping it to one paragraph. Use a link to di
 <DoDontContainer>
   <Do>
     <img src="https://github.com/primer/design/assets/175638/f8bd5ed9-d716-40c2-aef6-33979a87784d" />
-    <Caption>Use a short description to introduce the Selection.</Caption>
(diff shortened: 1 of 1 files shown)


auareyou (approved):
👍

rezrah (approved):
:shipit:

rezrah:
@danielguillan Minor comment on something unrelated to this change, but in the SectionIntro guidelines is the text in the images, which seem to be incorrectly justified. Maybe a Figma thing as it doesn't happen in React.

Screenshot 2024-02-20 at 09 33 16

Screenshot 2024-02-20 at 09 34 58

nsolerieu (approved):
🟢 Well worded @danielguillan 

`Don't use section intro for long form content.` is the biggie here.

----

🙋‍♂️

Perhaps we could also add the fact that one should not stretch the width of the component to flex across the full width of the container (like in the screenshot below?)

Screenshot 2024-02-20 at 3 29 45 PM

The line length could (in a future version of the component) be set depending on the type of heading? or left to the designer to flex according the the grid and specific case? Maybe we should discuss this during PB office hours. Not a blocker

PR #541, opened 2024-02-27
Add hover states to social icon links in MinimalFooter

danielguillan:
Solves https://github.com/github/primer/issues/3063

[Storybook examples](https://primer-3ecc927f1b-26139705.drafts.github.io/storybook/?path=/story/components-minimalfooter-features--dark-theme)

_Please note: this temporary solution uses CSS filters to work around a limitation in the current implementation of MinimalFooter. The social icons are loaded in an `img` tag, preventing direct fill color changes in CSS. In a future PR, we will inline SVG icons to set the `hover` state color directly._

### DIFF
--- a/packages/react/src/MinimalFooter/MinimalFooter.module.css
+++ b/packages/react/src/MinimalFooter/MinimalFooter.module.css
@@ -59,7 +59,8 @@
   padding-bottom: var(--base-size-48);
 }
 
-.Footer__social-link:hover .Footer__social-icon {
+.Footer__social-link:hover .Footer__social-icon,
+.Footer__social-link:focus-visible .Footer__social-icon {
   filter: var(--brand-footer-socialIcon-hoverFilter);
 }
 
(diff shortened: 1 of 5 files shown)


packages/react/src/MinimalFooter/MinimalFooter.module.css
```diff
@@ -59,6 +59,11 @@
   padding-bottom: var(--base-size-48);
 }
 
+.Footer__social-link:hover .Footer__social-icon,
+.Footer__social-link:focus-visible .Footer__social-icon {
+  filter: var(--brand-footer-socialIcon-hoverFilter);
```
rezrah, line 64: Should there be any transition between the idle and active states @danielguillan? Feels a bit sudden currently, but non-blocking feedback / question.

rezrah (approved):
LGTM  👍  :shipit:

packages/react/src/MinimalFooter/MinimalFooter.module.css
danielguillan: Definitely! We can add those in a follow-up PR.

PR #544, opened 2024-03-02
Remove icon lazy loading from EyebrowBanner stories

stamat:
## Summary

I think we should remove lazy loading of the icon from storybook examples of our `EyebrowBanner`, since I accidentally without thinking copied it, and this component is intended to be used always above the fold. So lazy loading just slows down the first viewport render.

## List of notable changes:

- **removed** Remove icon lazy loading from EyebrowBanner stories since it's intended to be used always above the fold

-  On the removed part. It's just presentation anyways.

## Steps to test:

1. Right click inspect. There is no `loading="lazy" decoding="async"`

## Contributor checklist:

## Reviewer checklist:


rezrah (commented):
👍 Yeah, that's fine @stamat. Happy to remove it from Storybook. 

By 'slowing down first viewport render', are you referring to the perceived delay experienced by the user, where you'd prefer this to be potentially render-blocking? We added lazy loading as a recommendation to avoid that situation, and this was copied verbatim from the GitHub homepage during Universe last year.

stamat:
@rezrah Yeah###  I'm referring to a quick blink the user will see instead of it instantly loading with the hero. I made the original Rails EyebrowBanner component, which uses Tobi's Rails ImageComponent with a `lazyload` that defaults to `true`. Either I didn't set the `lazyload` to `false` or the `lazyload` changed the default state last year. But I believe I was to blame lol... And that I wasn't thinking about it. But it really shouldn't have the lazyloading.

Do see the convo here ☺️ https://github.com/github/github/pull/315003#pullrequestreview-1910816415

rezrah (approved):
Thanks for context. This one's good to go 👍

PR #546, opened 2024-03-05
`axe:test` target individual stories

stamat:
## Summary

I suggest the modification of `e2e/scripts/playwright/axe-clean.spec.ts` which used to load links like this 👇 

`http://localhost:6006/?path=/story/components-accordion--default`

This means that it was loading the whole Storybook interface with the story sidebar. This was causing some of the stories to fail AXE tests, [here is an example](https://github.com/primer/brand/pull/503#issuecomment-1979076072).

The fix for it should be to load the stories via links like this 👇 

`http://localhost:6006/iframe.html?id=components-accordion--default&viewMode=story`

But then we will need to add an ignore list for certain AXE violations, because not all stories contain certain page elements, like all necessary landmarks and headings, cause we are testing components, not the whole pages.

@rezrah do tell me what you think 🙏  Keep in mind that this is just a quick sketch! ☺️ 

## Contributor checklist:

## Reviewer checklist:


packages/e2e/scripts/playwright/axe-clean.spec.ts
```diff
@@ -65,6 +64,11 @@ const testsToSkip = [
 const testsWithCustomDelay = {
   'components-subdomainnavbar--mobile-menu-open': 5000, // takes a while for the menu to open
 }
+const ignoreViolations = [
+  'landmark-one-main', // on most of the sotries we don't have a main landmark
+  'page-has-heading-one', // on some stories we dont have a heading,
```
rezrah, line 69: In some components, we do need to test for this, like Hero 🤔... can we create a safe list for those stories?

rezrah (commented):
Great suggestion @stamat 👍... left one item of feedback for you.

packages/e2e/scripts/playwright/axe-clean.spec.ts
stamat: @rezrah you are absolutely right! We can expand on this 100% and then have a list on which stories we need the check "page-has-heading-one".

We can have it as an object, since the filter would be much faster then. Something like this:
```javascript
const ignoreViolations = {
  'landmark-one-main':  {except: []}, // or {} or {except: null | false | []} 
  'page-has-heading-one': {except: [...]}
  'region': {except: []}
}
```

stamat:
@rezrah developing the ignore exclude list (you can also exclude via a part of the ID, meaning the whole groups) in order to test this a bit complex excluding feature I needed to test it, so I also updated the script to print violations so we can now know what failed directly from the CI 👇 

Screenshot 2024-03-06 at 15 16 42

Hope that's ok? I'm also going to speed up the test a bit. Seems like the collector of all violations is within the `for` loop, and launch the browser outside of the loop, and close it after the loop if possible. So we don't launch it every time.

packages/e2e/scripts/playwright/axe-clean.spec.ts
```diff
@@ -47,7 +47,7 @@ let page: Page
 
 let allViolations: Result[] = []
```
stamat, line 48: @rezrah I wanted to improve this playwright test so bad, but I realized few things:
1. This variable never aggregates data, it get's reset for every test - tests don't share states in playwright. (I wonder if the browser is shared though, could try to share the state through browsers localstorage) The only way to share the state is to write it to a file - but that poses another issue 👇 
2. `beforeAll` and `afterAll` get triggered on every test fail - maybe it's a bug, maybe I'm setting things in a bad way, but what I'm reading it's how it works? `beforeAll` and `afterAll` are currently in the `for loop`, and I don't think that's good and performant. This makes these hooks run for every story...

I tried sharing state through a file, but then because `beforeAll` works unexpectedly I will need to have an additional two scripts that get executed before and after running of playwright to set and clean the stage for the state file.

I think this is better suited for another PR. I think I made a big progress.

packages/e2e/scripts/playwright/axe-clean.spec.ts
```diff
@@ -56,12 +56,59 @@ const testsToSkip = [
...
   'components-videoplayer--playground', // video makes this too flakey
-  'components-videoplayer--playground', // video makes this too flakey
   'recipes-feature-previews-level-1--level-one-side-by-side-enterprise', // video makes this too flakey
   'recipes-feature-previews-level-1--level-one-side-by-side', // custom, unrelated background image
   'components-eyebrowbanner-features--on-custom-background-dark', // custom, unrelated background image
   'components-eyebrowbanner-features--on-custom-background-light', // custom, unrelated background image
 ]
+
+const ignoreViolations = {
+  'landmark-one-main': {except: []}, // on most of the sotries we don't have a main landmark
```
rezrah, line 66: qq about this one @stamat... do we need this rule? So far most tests pass without needing to ignore it, so it feels like we should instead opt-in certain stories instead. wdyt?

packages/e2e/scripts/playwright/axe-clean.spec.ts
```diff
@@ -56,12 +56,59 @@ const testsToSkip = [
...
+      color = '\x1b[35m' // magenta
+      break
+    case 'critical':
+      color = '\x1b[31m' // red
+      break
+  }
+  return `${color}${impact}\x1b[0m`
+}
+
+function printViolations(violations: Result[]) {
```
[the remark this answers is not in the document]
rezrah: I've just seen the screenshots to see how this looks 👍.

rezrah (approved):
This is very helpful @stamat, thank you. Let's get this shipped.

PR #547, opened 2024-03-06
Add Tooltip component

mperrotti:
Closes https://github.com/github/primer/issues/3060

## Summary

Adds a Tooltip component for showing contextual information when focusing or hovering interactive elements. I made the decision to remove the caret shown in the screenshot in the issue's description. Other Primer designers were happy with this, and it's the same style we use in `@primer/react`'s new Tooltip.

## List of notable changes:

**added** Tooltip component

-
-

Confirm that the Tooltip looks and behaves as expected.

- 
-

## Steps to test:

1. Open the Tooltip component in the CI-deployed preview environment
2. Check out all of the features demo'd in the "Features" directory or use Storybook's controls to play with the "Playground" story
3. Verify that the Tooltip behaves as described in the following issue.

## Contributor checklist:

## Reviewer checklist:

## Screenshots:

<table>
<tr>
</tr>
<tr>
<td valign="top">

N/A

 </td>
<td valign="top">

https://github.com/primer/brand/assets/2313998/bfe95579-059e-48bb-b1a3-a15c95f55f43

</td>
</tr>
</table>


nsolerieu:
@mperrotti a few questions for ya:

- Do we need the `North East`, `North West`, `Southeast`, `Southwest` - is there a reason why we'd not have only top/bottom/left/right?

- Should we consider a min-width to match the target element width for top/bottom variants?
Screenshot 2024-03-07 at 11 43 11 AM

- Have you considered adding a triangle/cursor?
Screenshot 2024-03-07 at 11 39 03 AM

mperrotti:
> Do we need the `North East`, `North West`, `Southeast`, `Southwest` - is there a reason why we'd not have only top/bottom/left/right?

We don't strictly _need_ those additional options, but I think it's nice to have them as options. If you and other designers want to limit the options, we can remove them.

---

> Should we consider a min-width to match the target element width for top/bottom variants?

This is possible, but I'd rather not do this. I don't see what we'd gain by doing this, and I like that the tooltip is sized to "shrink-wrap" its content. I'm also concerned about having a noticeable performance hit on pages with many tooltips. Whenever you measure a DOM node, the browser needs to perform a reflow.

---

> Have you considered adding a triangle/cursor?

Yes. We used to have that triangle in `@primer/react`'s old Tooltip component, but we chose to remove it for aesthetic reasons. I personally prefer the version without the caret.

nsolerieu:
@mperrotti Thank you for your answers, that all make sense. I like the minimalism and flexibility of this new version.
LGTM 🟢

nsolerieu:
@mperrotti following up here this as I brought this the the broader team for review.

We have review and decided that the proposed component is too small for marketing contexts.
We understand that this is a product pattern trickling down to brand and would like to consider this difference of context in the definition of the attributes of the component.

**We'd like the see a revised version including:**
- A caret for visual anchor
- Larger font size and lateral padding
- Limit the positions to `top`, `left`, `right`, `bottom`

nsolerieu (approved):
Thank you for this new version @mperrotti 

Screenshot 2024-03-14 at 11 13 27 AM

LGTM 🟢

packages/react/package.json
```diff
@@ -43,6 +43,7 @@
   },
   "dependencies": {
     "@primer/behaviors": "^1.3.4",
+    "@oddbird/popover-polyfill": "^0.3.1",
```
rezrah, line 46: Looks like the latest version is `0.4.0`. Could we use that version for this initial ship please?

packages/react/src/Tooltip/Tooltip.features.stories.tsx
```diff
@@ -0,0 +1,77 @@
...
+
+export default {
+  title: 'Components/Tooltip/Features',
+  component: Tooltip,
+} as Meta<typeof Tooltip>
+
+export const AnchorHasMargin = () => (
+  <Box padding="spacious">
+    <Tooltip text="Tooltip is still centered">
+      <Button sx={{marginLeft: 3}}>Button has 16px margin Left</Button>
```
rezrah, line 15: Brand components don't have `sx` prop support, so I'm guessing that this example isn't working as expected?

To fix this, you can add a new stylesheet called `Tooltip.stories.module.css` and apply the override style using CSS modules.

packages/react/src/Tooltip/Tooltip.features.stories.tsx
```diff
@@ -0,0 +1,77 @@
...
+
+export const LabelType = () => (
+  <Box padding="spacious">
+    <Tooltip text="Contribution Documentation for 'Primer React'" type="label">
+      <Link href="https://github.com/primer/react/contributor-docs/CONTRIBUTING.md">
+        <BookIcon />
+      </Link>
+    </Tooltip>
+  </Box>
+)
```
rezrah, line 28: FYI, this story doesn't seem to be working as expected (compared to the PRC version).


https://github.com/primer/brand/assets/13340707/06e9e5aa-f4ec-4834-94fb-940becbed832

rezrah: I'm also not sure we'd want to encourage using tooltip this way., using an Octicon inside a Link component. cc. @nsolerieu to keep me honest though.

packages/react/src/Tooltip/Tooltip.features.stories.tsx
```diff
@@ -0,0 +1,77 @@
...
+
+// As a supplementary description for a button
+export const DescriptionType = () => (
+  <Box padding="spacious">
+    <Tooltip text="Supplementary text" direction="n">
+      <Button>Save</Button>
+    </Tooltip>
+  </Box>
+)
```
rezrah, line 37: This example looks the same as the default example, but with an alternative direction. Is that right? Wondering if we can skip this one and dedupe? WDYT @mperrotti, i'm probably missing something crucial here though.

packages/react/src/Tooltip/Tooltip.features.stories.tsx
```diff
@@ -0,0 +1,77 @@
...
+  <Stack gap="spacious" padding="spacious">
+    <Tooltip direction="w" text="But appears in the east direction due to not having enough space in the west">
+      <Button>West</Button>
+    </Tooltip>
+
+    <Tooltip text="The direction here is north by default but there is not enough space in the north therefore the tooltip appears in the south">
+      <Button>North</Button>
+    </Tooltip>
+  </Stack>
+)
```
rezrah, line 77: Cool example! I think the west instance isn't working correctly though, because the buttons have displayed at full width.

Screenshot 2024-03-25 at 16 31 24




I _think_ the fix here could be to use `direction="horizontal"` in Stack, so that the buttons appear horizontally.

packages/react/src/Tooltip/Tooltip.module.css
```diff
@@ -0,0 +1,156 @@
+.Tooltip {
+  display: none;
+}
+
+@media (forced-colors: active) {
+  .Tooltip {
+    outline: 1px solid transparent;
```
rezrah, line 7:
```suggestion
    outline: var(--brand-borderWidth-thin) solid transparent;
```

packages/react/src/Tooltip/Tooltip.module.css
```diff
@@ -0,0 +1,156 @@
...
+  animation-delay: 0s;
+}
+
+/* This is needed to keep the tooltip open when the user leaves the trigger element to hover tooltip */
+.Tooltip::after {
+  position: absolute;
+  display: block;
+  right: 0;
+  left: 0;
+  height: 8px; /* This is a "magic number". Just gives some room to make sure the tooltip stay open when mouse leaves the trigger to hover tooltip */
```
rezrah, line 58:
```suggestion
  height: var(--base-size-8); /* This is a "magic number". Just gives some room to make sure the tooltip stay open when mouse leaves the trigger to hover tooltip */
```

packages/react/src/Tooltip/Tooltip.module.css
```diff
@@ -0,0 +1,156 @@
...
+  /* for scrollbar */
+  overflow: visible;
+}
+
+.Tooltip[popover]:popover-open, /* selector in chrome is :popover-open */
+.Tooltip[popover].\:popover-open {
+  /* selector in firefox and safari is \:popover-open */
+  display: block;
+  animation-name: tooltip-appear;
+  animation-duration: 0.1s;
```
rezrah, line 46: The fastest animation timing in Brand so far is 0.2s... could that work here?

Screenshot 2024-03-25 at 16 34 49

packages/react/src/Tooltip/Tooltip.module.css
```diff
@@ -0,0 +1,156 @@
...
+
+.Tooltip[data-direction='s']::after {
+  bottom: 100%;
+}
+
+.Tooltip[data-direction='w']::after {
+  position: absolute;
+  display: block;
+  height: 100%;
+  width: 8px; /* This is a "magic number". Just gives some room to make sure the tooltip stay open when mouse leaves the trigger to hover tooltip */
```
rezrah, line 132:
```suggestion
  width: var(--base-size-8); /* This is a "magic number". Just gives some room to make sure the tooltip stay open when mouse leaves the trigger to hover tooltip */
```

packages/react/src/Tooltip/Tooltip.module.css
```diff
@@ -0,0 +1,156 @@
...
+
+.Tooltip[data-direction='e']::after {
+  position: absolute;
+  display: block;
+  height: 100%;
+  width: 8px;
+  content: '';
+  bottom: 0;
+  right: 100%;
+  margin-left: -8px; /* This is a "magic number". Just gives some room to make sure the tooltip stay open when mouse leaves the trigger to hover tooltip */
```
rezrah, line 146: As we're using this value in quite a few places, can it be turned into a locally scoped variable instead which maps to ` var(--base-size-8)`?

packages/react/src/Tooltip/Tooltip.features.stories.tsx
nsolerieu: What @rezrah recorded definitely looks odd

- Having octicon to a link is okay **WITH** a label
- Having **ONLY** an octicon in a link is not okay
- Using a tooltip on a link is okay as long as the link is setup properly (label + icon + arrow OR label + arrow)

stamat (commented):
Great work! ☺️ Glad you managed to swim out of the obstacles! ✨ 

Looking at this code now, and I can't help to think if this could have been done in a much simpler way... I'm not sure why `popover` was used at the first place?  It seems that it's used for modal dialogs, that overlay the screen?

> When open, popover elements will appear above all other elements in the [top layer](https://developer.mozilla.org/en-US/docs/Glossary/Top_layer), and won't be influenced by parent elements' [position](https://developer.mozilla.org/en-US/docs/Web/CSS/position) or [overflow](https://developer.mozilla.org/en-US/docs/Web/CSS/overflow) styling.

Secondly, maybe we could have the Tooltip just plug in `aria-label` to the child element if the child elemen't doesn't have one, or that the tooltip uses the existing label of the child if the tooldtip doesn't hace one. Basically the Tooltip would be `aria-hidden` itself, since it's just using the `aria-label` text within it, and the screen readers shouldn't doesn't care if it reads the label from `arial-labeledby` or the actual `aria-label` (or maybe I got it wrong). We can then modify it's appearance with the viewport/container in mind when toggling it's visibility on programmatic hover/focus like you already do. Unless of course we want these tooltips to potentially contain some UI controls in the future?

I mean I feel you, I always go for a more complex solution! 😆 But I'm trying not to... Because 80% of the time it turns out that the simplest solution was the way to go.

packages/react/src/Tooltip/Tooltip.module.css
```diff
@@ -0,0 +1,156 @@
...
+  content: '';
+  bottom: 0;
+  left: 100%;
+}
+
+.Tooltip[data-direction='e']::after {
+  position: absolute;
+  display: block;
+  height: 100%;
+  width: 8px;
```
rezrah, line 142:
```suggestion
  width: var(--base-size-8);
```

packages/react/src/Tooltip/Tooltip.module.css.d.ts
```diff
@@ -0,0 +1,7 @@
+declare const styles: {
+  readonly "Tooltip": string;
+  readonly ":popover-open": string;
+  readonly "tooltip-appear": string;
+};
+export = styles;
+
```
rezrah, line 7: Think this file will need to be regenerated 🤔 

`npm run build:lib` might do it

packages/react/src/Tooltip/Tooltip.stories.tsx
```diff
@@ -0,0 +1,49 @@
...
+export const Default = () => (
+  <Box padding="spacious">
+    <Tooltip text="Hello, Tooltip!">
+      <Button>Hover me</Button>
+    </Tooltip>
+  </Box>
+)
+
+export const Playground = args => {
+  // this is a hack to remove the `type` prop from the args because for this example type label is not a valid choice and violates accessibility
```
rezrah, line 22: 👍 this gets picked up in Storybook controls too...

Screenshot 2024-03-26 at 15 08 37

Should be able to remove it with:

```js
type: {
  table: {
    disable: true,
  },
},
```

packages/react/src/Tooltip/Tooltip.tsx
```diff
@@ -0,0 +1,211 @@
...
+/** * Main Stylesheet (as a CSS Module) */
+import styles from './Tooltip.module.css'
+
+type TooltipDirection = 'n' | 'e' | 's' | 'w'
+export type TooltipProps = React.PropsWithChildren<{
+  direction?: TooltipDirection
+  id?: string
+  text: string
+  type?: 'label' | 'description'
+  children?: React.ReactNode
```
rezrah, line 20: As you're already using `PropsWithChildren`, I don't think this manual declaration is necessary?

packages/react/src/Tooltip/Tooltip.tsx
```diff
@@ -0,0 +1,211 @@
...
+      const hasInteractiveChild = Array.from(triggerChildren).some(triggerChild => {
+        return triggerChild instanceof HTMLElement && isInteractive(triggerChild)
+      })
+
+      if (
+        !(isTriggerInteractive || hasInteractiveChild) &&
+        (process.env.NODE_ENV === 'development' || process.env.NODE_ENV === 'test')
+      ) {
+        throw new Error(
+          'The `Tooltip` component expects a single React element that contains interactive content. Consider using a `<button>` or equivalent interactive element instead.',
```
rezrah, line 106: This feels more like a warning than an error. Curious if we want to use a console.warn here, wrapped in a guard to only appear in non-prod environments.

danielguillan:
Tooltip in dark color theme

Are these the intended colors for tooltips in the dark color theme? They don't seem to meet the minimum required contrast ratio. Could we perhaps consider using a dark tooltip in the dark theme instead of having "inverted" colors?

packages/react/src/Tooltip/Tooltip.features.stories.tsx
mperrotti: @nsolerieu @rezrah 

Aren't we planning to use tooltips on the social media icon links in the footer? Wouldn't that be an example of a visible label being replaced by a Tooltip?

mperrotti: I agree this example looks strange with with the arrow that automatically gets added by `Link`. How do you recommend I move forward?
- Update this story to use the correct component for rendering an icon-only link or button
- Remove this story and the ability to use a Tooltip as a label
- Something else

packages/react/src/Tooltip/Tooltip.features.stories.tsx
mperrotti: You're not missing anything critical. Since "description" tooltips are the default, this is essentially the same as the default. Happy to remove.

packages/react/src/Tooltip/Tooltip.tsx
mperrotti: Strangely, I'm getting TypeScript errors when I remove the `children` property from the type. I'm just going to remove `React.PropsWithChildren`.

packages/react/src/Tooltip/Tooltip.features.stories.tsx
mperrotti: I just used the `style` prop since this is just used to demonstrate how Tooltip alignment doesn't break.

mperrotti:
Good catch @danielguillan - these colors do not pass for smaller font sizes. **The contrast ratio is 4.47** (minimum is 4.5). The background color is set to `--brand-color-neutral-emphasisPlus` (`#6e7681`), and the foreground color is set to `--brand-color-text-onEmphasis` (`#010409`).

Maybe we should rethink the dark mode value for `--brand-color-neutral-emphasisPlus`. If we change it to `--base-color-scale-gray-3` (`#8b949e`), the contrast ratio would be 6.68.

Here's what that looks like:
Screenshot 2024-04-11 at 2 30 48 PM

Let me know how you think we should move forward.

danielguillan (approved):

packages/react/src/Tooltip/Tooltip.features.stories.tsx
rezrah: I'd recommend using a Button instead @mperrotti... 

Example:

Screenshot 2024-04-16 at 09 35 27

packages/react/src/Tooltip/Tooltip.module.css
```diff
@@ -0,0 +1,160 @@
...
+  .Tooltip {
+    outline: var(--brand-borderWidth-thin) solid transparent;
+  }
+}
+
+/* Overriding the default popover styles */
+.Tooltip[popover] {
+  --offset: var(
+    --base-size-8,
+    8px
```
rezrah, line 15:
```suggestion
    --base-size-8
```

packages/react/src/Tooltip/Tooltip.module.css.d.ts
rezrah: weird, could you please try `build:css-types` script in `packages/react`?

packages/react/src/Tooltip/Tooltip.tsx
```diff
@@ -0,0 +1,211 @@
...
+    const openTooltip = () => {
+      if (tooltipElRef.current && triggerRef.current && !tooltipElRef.current.matches(':popover-open')) {
+        tooltipElRef.current.showPopover()
+      }
+    }
+    const closeTooltip = () => {
+      if (tooltipElRef.current && triggerRef.current && tooltipElRef.current.matches(':popover-open')) {
+        tooltipElRef.current.hidePopover()
+      }
+    }
```
rezrah, line 86: Could we memoize these functions in a useCallback?

packages/react/src/Tooltip/Tooltip.module.css
```diff
@@ -0,0 +1,160 @@
...
+
+  position: absolute;
+  padding: 0.5em 0.75em;
+  width: max-content;
+  margin: auto;
+  clip: auto;
+  white-space: normal;
+  font: normal normal var(--brand-text-size-100) / var(--brand-text-lineHeight-100) var(--brand-fontStack-sansSerif);
+  -webkit-font-smoothing: subpixel-antialiased;
+  color: var(--brand-color-text-onEmphasis);
```
rezrah, line 32: In dark mode this feels quite difficult to read, despite contrast ratio meeting requirements. Is it an aesthetic choice to use two dark values here? Or could we switch to using the default text token (white in dark mode) here?

Current:
Screenshot 2024-04-16 at 09 51 11



With white text:

packages/react/src/Tooltip/Tooltip.features.stories.tsx
```diff
@@ -0,0 +1,68 @@
...
+
+export default {
+  title: 'Components/Tooltip/Features',
+  component: Tooltip,
+} as Meta<typeof Tooltip>
+
+export const AnchorHasMargin = () => (
+  <Box padding="spacious">
+    <Tooltip text="Tooltip is still centered">
+      <Button style={{marginLeft: '16px'}}>Button has 16px margin Left</Button>
```
rezrah, line 15: Stray capitalization

```suggestion
<Button style={{marginLeft: '16px'}}>Button has 16px margin left</Button>
```

rezrah (commented):
This is looking great @mperrotti 🙌 

Left some more minor feedback, and a few questions.

A few other things that don't fit as inline feedback..

- Changeset will need to be added (examples for how we call our new components [here](https://github.com/primer/brand/blob/main/packages/react/CHANGELOG.md))
- Visual snapshots can be regen'd using `update snapshots` label
- Can't see the a documentation page. Planning to add this now or after?

packages/react/src/Tooltip/Tooltip.module.css
mperrotti: @rezrah - I'm comfortable with the white text in dark mode. Let me just confirm it meets the contrast requirement.

mperrotti:
Thanks for the updated review @rezrah !

> Can't see the a documentation page. Planning to add this now or after?

I think this should be a separate PR. Do we need to create an issue?

rezrah:
> Thanks for the updated review @rezrah !
> 
> > Can't see the a documentation page. Planning to add this now or after?
> 
> I think this should be a separate PR. Do we need to create an issue?

We usually add it to the same PR, but if you need to add it after, maybe capture it on a tasklist on the tracking issue?

Btw, I'm referring to the React component docs, not the interface guidelines.

mperrotti:
Ahh the React docs. Yes I can add those here.

mperrotti:
@rezrah - I added React docs, but I'm getting this error:
```
ReferenceError: Tooltip is not defined
```
Can you confirm that I'm doing my exports from `packages/react` correctly?

Also, I'm having trouble running Gatsby locally to test my docs locally, so it's hard to debug.

apps/docs/content/components/Tooltip.mdx
```diff
@@ -0,0 +1,57 @@
...
+    <Button hasArrow={false}><ScreenFullIcon /></Button>
+</Tooltip>
+```
+
+## Component props
+
+| name         | type                                                              | default         | required | description                             |
+| ------------ | ----------------------------------------------------------------- | --------------- | -------- | --------------------------------------- |
+| `direction`  | <PropTableValues values={['n', 'e', 's', 'w']} />                 | `'s'`           | `false`  | The text to be displayed in the tooltip |
+| `text`       | `string`                                                          | `undefined`     | `true`   | Determines whether the ActionMenu is open by default |
```
rezrah, line 55: Think these are the wrong way around, and one is referencing a different component.

```suggestion
| `direction`  | <PropTableValues values={['n', 'e', 's', 'w']} />                 | `'s'`           | `false`  | The side the tooltip caret appears on |
| `text`       | `string`                                                          | `undefined`     | `true`   |  The text to be displayed in the tooltip |
```

apps/docs/content/components/Tooltip.mdx
```diff
@@ -0,0 +1,57 @@
...
+
+If a ubiquitously understood icon is used in place of a text label, you can use the `type` to provide a text label for assistive technologies such as screen readers.
+
+Use this with caution. It is generally better to use a visible label for better usability and accessibility.
+
+```jsx live
+<Tooltip text="Go fullscreen" type="label">
+    <Button hasArrow={false}><ScreenFullIcon /></Button>
+</Tooltip>
+```
```
rezrah, line 48: i think we should go without the live previews here, as they don't work in Gatsby

Screenshot 2024-05-02 at 16 14 03

rezrah (approved):
🙌 Nice one @mperrotti, this looks great. 

Just some minor, additional feedback around docs but otherwise LGTM 👍 :shipit:

PR #563, opened 2024-04-04
Fix documentation examples and note for Pricing options

danielguillan:
Fixes a note formatting issue in the pricing options documentation and changes three-item examples to two-item examples for a more accurate representation of the live component in the docs.

| Before | After |
|--------|--------|
| screenshot-nuaMo1aA-000948@2x | screenshot-EWemIrcK-000947@2x |
| screenshot-djVhfqbD-000949@2x | screenshot-ExECHbky-000950@2x |
| screenshot-Y2Ykna73-000951@2x | screenshot-reOhkvFB-000952@2x |

### DIFF
--- a/apps/docs/content/components/PricingOptions.mdx
+++ b/apps/docs/content/components/PricingOptions.mdx
@@ -255,15 +255,17 @@ This is the two-item layout for the pricing options component.
 
 This is the three-option layout for the pricing options component.
 
-```jsx live
 <Note>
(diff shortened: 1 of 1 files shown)


rezrah (approved):
Thanks for fixing this @danielguillan 🙏 

One other thing, if you're able to add it quickly is an updated example showing the `originalPrice` text. This is quite an important update, and couldn't see it anywhere besides the prop table.

PR #593, opened 2024-05-17
SEO

joshfarrant:
## Summary

[Link to Storybook](https://primer-9b9e948839-26139705.drafts.github.io/storybook/?path=/story/recipes-seo-article-page--playground)
[Link to Slack for PR discussion](https://github.slack.com/archives/C06UE379Y3U/p1715973027064179)


packages/react/src/recipes/seo/helpers.ts
```diff
@@ -0,0 +1,70 @@
...
+    },
+    dark: {
+      accent1: 'var(--base-color-scale-purple-5)',
+      accent2: 'var(--base-color-scale-lemon-1)',
+    },
+  },
+  security: {
+    light: {
+      accent1: 'var(--base-color-scale-teal-2)',
+      accent2: '#A9E500',
```
nsolerieu, line 53: @joshfarrant all the color have `--base-scale` equivalents - sorry I should have mentioned this.
I'll not them in here to avoid using custom HEX values and provide some guidance:

Here security.accent2 should be `var(--base-color-scale-lime-1)`

packages/react/src/recipes/seo/helpers.ts
```diff
@@ -0,0 +1,70 @@
...
+    },
+    dark: {
+      accent1: 'var(--base-color-scale-teal-2)',
+      accent2: '#A9E500',
+    },
+  },
+  productivity: {
+    light: {
+      accent1: '#7C72FF',
```
nsolerieu, line 62: @joshfarrant `--base-scale` color matching:

`productivity.accent1` => `var(--base-color-scale-indigo-4)`

packages/react/src/recipes/seo/helpers.ts
```diff
@@ -0,0 +1,70 @@
...
+    },
+    dark: {
+      accent1: '#096BDE',
+      accent2: '#00FF46',
+    },
+  },
+  collaboration: {
+    light: {
+      accent1: 'var(--base-color-scale-purple-5)',
+      accent2: '#FF5934',
```
nsolerieu, line 33: @joshfarrant `--base-scale` color matching:

`collaboration.accent2` => `var(--base-color-scale-coral-4)`

packages/react/src/recipes/seo/helpers.ts
```diff
@@ -0,0 +1,70 @@
...
+
+export type ThemeDetailsMap = {
+  [theme in Themes]: {
+    [ColorModesEnum.LIGHT]: Palette
+    [ColorModesEnum.DARK]: Palette
+  }
+}
+
+export const themeDetailsMap: ThemeDetailsMap = {
+  ai: {
```
nsolerieu, line 20: @joshfarrant `--base-scale` color matching:

`ai.accent1` => `var(--base-color-scale-blue-5)`
`ai.accent2` => `var(--base-color-scale-lime-2)`

nsolerieu (changes requested):
I added a few comments regarding matching the brand color to the `--base-scale` - this is something @site-design-sg is aware and has been pushing for. The (customized) brand color in the [Brand Platform library file](https://www.figma.com/design/Opg3W9M1Qb9rfs55M1LrMF/Brand-Asset-Library-2023?m=auto&node-id=17%3A1972&t=1C4mXdUWoRW2ggP2-1) are not web optimized and the context of this implementation is fitting for this to happen.

The refactor is started the foundation for this web transition [here](https://www.figma.com/design/9TVSd8V5R9FAJmu4ub4jvA/Refactor---Working-File-2024?node-id=84%3A2050&t=JLn6E8s42Ak7BWJL-1) - This issue has already been brought up to Primer Brand so we might need to discuss this implementation again

@ajashams @jesussandreas Refactor 👀

packages/react/src/recipes/seo/Article/Article.stories.tsx
```diff
@@ -24,59 +26,158 @@ export default {
...
-    args: {
-      heroTitle: 'Headline',
-      lede: "There's one word that perfectly describes successful DevOps: flow. As individuals, we experience a state of flow when everything in our work comes together naturally and at the right time. DevOps enables that kind of flow at the organizational level through a combination of tooling, culture, and process.",
+  },
+  args: {
+    heroTitle: 'Should we think of DevOps as a methodology?',
+    lede: "There's one word that perfectly describes successful DevOps: flow. As individuals, we experience a state of flow when everything in our work comes together naturally and at the right time. DevOps enables that kind of flow at the organizational level through a combination of tooling, culture, and process.",
+    content: 'real-world',
+    accentColor: 'ai',
+    heroHeight: 70,
```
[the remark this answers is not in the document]
joshfarrant: Removed in 4932c3f31c98e5e39807d0d46e788fb0ea07f352

packages/react/src/recipes/seo/Article/Article.tsx
```diff
@@ -23,10 +23,66 @@ import {
...
 
 import {ColorModesEnum, ThemeProvider} from '../../../ThemeProvider'
-import heroImage from '../../../fixtures/images/background-lozenge-ai.png'
+import heroImage from '../../../fixtures/images/background-full-bleed.png'
 import placeholderImage from '../../../fixtures/images/background-poster-ai.png'
 
 import styles from './Article.module.css'
+import {Themes, themeDetailsMap} from '../helpers'
+
+const useWrapLines = (text?: string) => {
```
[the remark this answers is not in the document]
joshfarrant: dfe8ee7066cf8277e143b891fc82521e345f7529

packages/react/src/recipes/seo/Article/Article.module.css
```diff
@@ -28,13 +20,144 @@
   padding: 0;
 }
 
-.heroImageArea {
-  overflow: hidden;
+.subdomainNavBarContainer {
```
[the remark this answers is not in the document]
joshfarrant: 2e206312f5c6f6cefae4e17f136afe1b885560e0

packages/react/src/recipes/seo/Article/Article.module.css
```diff
@@ -28,13 +20,144 @@
...
+  min-width: 100%;
+  min-height: 100%;
+}
+
+.heroImageOverlay {
+  position: absolute;
+  top: 0;
+  width: 100%;
+  height: 100%;
+  background: linear-gradient(0deg, rgba(0, 0, 0, 0.8) 0%, rgba(0, 0, 0, 0) 50%, rgba(0, 0, 0, 0.6) 100%);
```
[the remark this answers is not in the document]
joshfarrant: The former, but I'm not sure we'd need separate values for light/dark colour modes as the hero styling stays the same across both light and dark; white text, over a semi-transparent black overlay, over an image.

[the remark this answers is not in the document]
rezrah: > as the hero styling stays the same across both light and dark; white text, over a semi-transparent black overlay, over an image.

👍 thanks Josh.

How are we planning to enforce usage of dark background images through the CMS? Is that something WebExp are onboard with and enforcing further down the pipeline? I think we need to account for light/dark variations of the hero at the systems level, like we do in F2 templates to account for future use-cases also. I can almost guarantee we'll see light backgrounds being used by content teams eventually, and content is impossible to guardrail unless we do it early.

Having only a dark hero also feels a little limiting and inconsistent with how other templates behave. Is that part of design spec? (sorry if I missed this).

If we're set on always having a dark mode, this is fine. Could we just leave a quick comment above that line to remind readers that this is 'based on assumption that hero will only be available in dark color mode' (or similar).

cc. @nsolerieu and @simmonsjenna - in case either of you know too 🙏

packages/react/src/fixtures/images/background-full-bleed.png
rezrah, line 1: More a design question, but can we use a different image here @simmonsjenna @nsolerieu? I think we should recommend one that isn't so busy at the bottom in our Storybook examples, to indicate best practice. With longer titles we see this issue 👇 

Screenshot 2024-06-12 at 09 36 14

Perhaps also worth enforcing a character limit in the Storybook control @joshfarrant?

packages/react/src/recipes/seo/Article/Article.module.css
```diff
@@ -28,13 +15,133 @@
...
+  background: black;
+  position: relative;
+  height: calc(70vh - var(--parallax-start-spacing));
+}
+
+.parallax {
+  position: absolute;
+  top: 0;
+  width: 100%;
+  height: 70vh;
```
rezrah, line 30: Nit.. 70vh is referenced a few times here. Worth making this a locally scoped token too like `parallax-start-spacing`?

packages/react/src/recipes/seo/Article/Article.tsx
```diff
@@ -405,223 +418,250 @@ export function Article({
...
-                                        ullamcorper id. Aliquam luctus sed turpis felis nam pulvinar risus elementum.
-                                      </p>
-                                    </FAQ.Answer>
-                                  </FAQ.Item>
-                                </FAQ>
-                              </FAQGroup>
-                            </section>
-                          </Stack>
+                                    {line}
+                                  </span>{' '}
```
rezrah, line 457:
```suggestion
                                  </span>
```

packages/react/src/recipes/seo/Article/Article.visual.spec.ts-snapshots/Visual-Comparison-Article-Article-Ai-Theme-1-linux.png
rezrah, line 1: @joshfarrant FYI these snapshots still seem like they're taken too early. Might need a longer delay.

packages/react/src/recipes/seo/Article/Article.module.css
joshfarrant: Fine by me — It was a CSS var a while ago (when it was exposed as a control in Storybook) and I removed it when I removed those controls.

Happy to split it out though, agree that it makes it a bit clearer 👍

packages/react/src/recipes/seo/Article/Article.tsx
joshfarrant: So I struggled with this as, annoyingly, that space is actually necessary. I previously had a comment there explaining that, which might still be worth doing.

TL;DR is that I split the words in the title by spaces, then when reconstructing the lines I add the spaces back in between words. I also need to make sure I add a space in between the lines too, otherwise we end up with the last span of one line and the first span of the next squished together without a space between them, meaning that two words get combined into one.

Eg, without the space we have

```html
<span>This is the first line</span><span>and this is the second</span>
```

Which will render as either

```
This is the first lineand
this is the second
```

or

```
This is the first
lineand this is the second
```

Annoying and messy, I know. Open to alternate approaches, and at the very least I can add a comment explaining this.

packages/react/src/recipes/seo/Article/Article.visual.spec.ts-snapshots/Visual-Comparison-Article-Article-Ai-Theme-1-linux.png
joshfarrant: I might have forgotten to regenerate after making that change. Let me do it now and see how we're looking

packages/react/src/fixtures/images/background-full-bleed.png
joshfarrant: Another image works for me. @simmonsjenna @nsolerieu lmk if you have a recommendation please 🙂

It doesn't look like there's a built-in way to enforce a character limit on text controls in Storybook unfortunately. ([Docs](https://storybook.js.org/docs/essentials/controls#annotation))

I can add a description to nudge the user, or I could modify the story and do a `heroTitle.substring(0, 90)` to truncate the amount of text that gets rendered in the preview, although that's not the nicest UX. lmk your thoughts @rezrah

image

packages/react/src/recipes/seo/Article/Article.module.css
joshfarrant: b790ac37db8404329ed24e6d738705138a3278af

packages/react/src/recipes/seo/Article/Article.module.css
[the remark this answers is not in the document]
joshfarrant: V good question. My assumption was that we'd always have that white text and dark overlay, regardless of the lightness of the image, but this could be a bad assumption on my part.

At the moment, a light image will look like this (left). The contrast is there, but the dark gradient overlay doesn't look quite right imo, maybe it needs some tweaking. Switching the colours would give us something like this (right) just for reference, which doesn't feel 100% right either, so input on this would be appreciated 🙂


| Current | Colours switched |
|--------|--------|
| image | image
 |

NB: If we're going to use different text/gradient styles on lighter images vs darker images then we'd also need to expose a control in Storybook to allow the user to say whether it is a light or a dark image. We can then look at the value of that control to decide what colours to use.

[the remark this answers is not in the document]
joshfarrant: A note on contrast actually. In the first image the contrast of the white text compared to the lightest part of the background is bang on 3:1 which, for large text, [is the minimum value for MAS compliance](https://liquid.microsoft.com/Web/Object/Read/ms.accessibility/Requirements/01.04.03#Zhome). We'd need to be careful not to reduce this further if we tweak the gradient at all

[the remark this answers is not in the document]
nsolerieu: @joshfarrant thank you for your suggestions, those are good callouts

I gave these a thought from a design perspective:
- Can we instead suggest to use the phase 1 `boxed` variant for light theme covers? The wallpaper variant is optimized for dark mode visuals
- Can we "force" dark mode in any way - this template can only be used in one color mode?
- The color switched version you suggested blends the hero with the content drawer, could we add a subtle box shadow to the content to create some separation?
Screenshot 2024-06-12 at 9 49 29 AM

packages/react/src/fixtures/images/background-full-bleed.png
nsolerieu: @joshfarrant this is a perennial issue - we definitely need some guardrails as our copywriter tends to use long wordy titles. We'll reach out to MPS as they have the ability to limit character count in contentfull and we'll also make sure it it clearly stated in-situ that we recommend a maximum of 3 lines (90 characters) with a sweet spot around 2 lines

packages/react/src/recipes/seo/helpers.ts
```diff
@@ -0,0 +1,70 @@
...
+    light: {
+      accent1: 'var(--base-color-scale-blue-5)',
+      accent2: 'var(--base-color-scale-lime-1)',
```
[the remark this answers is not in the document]
nsolerieu: @rezrah these are technically a stop-gap solution (see full --base-color / brand color matching [here](https://www.figma.com/design/n7dHtVVkFOvDLyxWaNfKou/CTA-banner-expressive-variant?node-id=1-13765&t=MljT3KnlwWpflZki-1) ). I'd recommend starting here and since Brand is working with Gretel (agency) to rework our color strategy, that may indeed change. That's why I'd recommend sticking to our color scale for the time being

packages/react/src/recipes/seo/Article/Article.module.css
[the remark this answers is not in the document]
joshfarrant: Thanks for the input @nsolerieu. If you're happy, I think option 3 would be the simplest approach to implement and maintain, whilst still offering flexibility.

Option 1 would require maintaining two separate templates going forward, so I'd be hesitant to go down that road.

Just to check my own understanding here — do articles have a single cover image (which could be light or dark), or would one article have two cover images; one for light mode and one for dark mode?

packages/react/src/fixtures/images/background-full-bleed.png
joshfarrant: Good suggestion to have the limit in Contentful, that makes perfect sense.

I'll add the note into the Storybook description (as in the above image) to act as a reminder of this too.

packages/react/src/recipes/seo/Article/Article.tsx
joshfarrant: 4f4e4971ac25106d2350acc7d6a9d6250bb34871

joshfarrant: 2e3b4b1751fb52a5fc01103452c89d2cffdfd8e1

packages/react/src/recipes/seo/Article/Article.module.css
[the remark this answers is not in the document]
joshfarrant: b14e45c277b4a08c31dc4f2861a3e3896093ea45

packages/react/src/recipes/seo/Article/Article.module.css
```diff
@@ -28,13 +15,166 @@
...
+  content: '';
+  display: inline;
+  position: absolute;
+  top: 0;
+  right: 0;
   width: 100%;
+  height: 100%;
+  background-color: var(--brand-color-accent-primary);
+  transform-origin: right;
+  animation: wipe 800ms forwards var(--brand-animation-easing-default);
```
rezrah, line 176: Non-blocking.. but we have an `--brand-animation-duration-extended` token at .7s that might be relevant for you here. It's 100ms faster though, but not sure how we landed at 800ms either.

packages/react/src/recipes/seo/Article/Article.stories.tsx
```diff
@@ -24,59 +28,127 @@ export default {
...
-      },
-      lede: {
-        control: 'text',
-        table: {
-          category: 'Section: Article body',
-        },
+    isLightHero: {
+      description: 'Should be true if hero image is light, false if dark',
+      control: 'boolean',
+    },
```
rezrah, line 52: Just a reminder about the light mode image being swapped in when this is true

rezrah (approved):
Nice one @joshfarrant... let's ship it 🚀 

- left some non-blocking feedback
- VRT issues are persisting. Recommend removing them if them from the suite if animations are still causing an issue
- Anything else can be mopped up later

packages/react/src/recipes/seo/Article/Article.module.css
joshfarrant: That works for me. I'll update it and ping @nsolerieu with a link to make sure he's happy

PR #595, opened 2024-05-21
[a11y] `AnchorNav` on small or zoomed in screens inaccessible elements fix

stamat:
## Summary

Hello there, again! 👋 

When there is a lot of items and the user is on a small 320x480 screen or zooms in the page, few last items can become inaccessible.

Adding `max-height: calc(100vh - var(--base-size-12) * 2); overflow-y: auto;` to `.AnchorNav-inner-container` fixes this issue. This adds the maximum height of the container to match the 100% of the viewport and adds an `overflow-y: auto` so the user can scroll and see all of the items.

I would love if we could 🚢  this ASAP. 🙏 

- Closes: https://github.com/github/accessibility-audits/issues/7852
- Closes: https://github.com/github/accessibility-audits/issues/8323

## List of notable changes:

- **updated** `.AnchorNav-inner-container` by adding `max-height: calc(100vh - var(--base-size-12) * 2); overflow-y: auto;`
- **removed** The custom `window.scrolTo()` was removed in favor of native anchoring and smooth scrolling. The issue was that the `event.preventDefault()` was preventing that the focus is shifted to a targeted element, and there is no substitute hack for this. We don't need the custom calculations because the browser already does them. The only downside of this is that the `#` hash anchor will appear in the browser... which is also the desired native behavior.

## Steps to test:

1. Open `/story/components-anchornav--default`
2. Reduce the screen height
3. <kbd>CMD</kbd> + <kbd>+</kbd> press several times untill the opened anchor nav is not fully visible

- https://github.com/github/accessibility-audits/issues/7852
- https://github.com/github/accessibility-audits/issues/8323

## Contributor checklist:

## Reviewer checklist:


joshfarrant (approved):

packages/react/src/AnchorNav/AnchorNav.stories.tsx
```diff
@@ -119,6 +119,7 @@ export const Playground = ({data, ...args}: {data: MockData; offset: number; sec
               <Text as="p">
                 AnchorNav is a component that allows users to navigate to different sections of a page.
               </Text>
+              <Button>Learn More</Button>
```
rezrah, line 122: @stamat Do you need the button here? Can we remove if not, or align to start of container?

stamat: @rezrah I needed it to test the focus and see if it works correctly. I think it should be there for that sole purpose. But I'll align it to the start of the container! ❤️

stamat:
@rezrah all done!

rezrah (approved):
Thanks @stamat ✨

PR #606, opened 2024-06-12
Add fullWidth prop to Card component

### LINKED ISSUE #603: [Feature Request] Allow Card to span the full width of its container.
**Problem**
Card contained in `Grid.Column` with span={12} is not stretching the entire width of the column. The use case would be viewing the Cards on a tablet where we have changed to span={12} at the xsmall breakpoint. Image of issue:

Image

The current workaround implemented is to center the Cards in this case, but this is a break from design that shows the cards either stretching the entire width of the `Grid` or having span={6}

**Design**
Image

**Workaround**

Image

**Proposal**
Allow users to set the `max-width` of the Card. Use type to control what values users can enter, but include "100%" so users can have the card fill the entire space of the container if need. If no `maxWidth` is set fallback to the `--brand-Card-maxWidth`

Image

rezrah:
## Summary

Closes #603 

Adds a `fullWidth` prop to the Card component to allow the card to take up the full width of the parent container. 

To date, we've expected end-users to override the value of `--brand-Card-maxWidth` to be 100%. We're seeing this happen too frequently, and this should now be supported in the API itself as a full width card is a different use case than a card with a custom width. 

## List of notable changes:

**added** `fullWidth` for Card component, along with tests and documentation

- Any visual design reasons for not enabling this in the API

## Steps to test:

1. Open the following story 
2. Increase/decrease window size
3. Verify that the Card stretches to the full width of the parent container, unlike the default card

- Original request #603 

## Contributor checklist:

## Reviewer checklist:

## Screenshots:

<table>
<tr>
</tr>
<tr>
<td valign="top">

Screenshot 2024-06-12 at 15 23 59

 </td>
<td valign="top">

Screenshot 2024-06-12 at 15 23 44

</td>
</tr>
</table>


packages/react/src/Card/Card.features.stories.tsx
```diff
@@ -44,6 +44,25 @@ export const CTAText: StoryFn<typeof Card> = () => {
   )
 }
 
+export const FullWidth: StoryFn<typeof Card> = () => {
```
joshfarrant, line 47: Is it worth adding the `fullWidth` prop to the Card playground story too?

joshfarrant (approved):
Looks great! One tiny question but not a blocker

packages/react/src/Card/Card.features.stories.tsx
rezrah: 💯 thanks for reminder @joshfarrant... totally missed that. Would be nice if auto args just worked though 🤷

nsolerieu (approved):
We discussed this during Site design jam with @rezrah - We agreed to move forward with enabling full width cards.

For the image, I'd recommend scaling the image proportionally (standard `display: block; width:100%; height: auto` behavior), instead of cropping

PR #617, opened 2024-06-19
Add schema

joshfarrant:
[Link to discussion on Slack](https://github.slack.com/archives/C06UE379Y3U/p1718802455538099)


rezrah (approved):
Looks great 👍

Out of interest @joshfarrant, were you given a minimum set of attributes to add here? Wondering if this is enough, or if we need someone in marketing to cross-check it. Either way, it can always be added in later during integration phase, so this is non-blocking.

joshfarrant:
@rezrah No, the attributes I added just came from my own research on [schema.org](https://schema.org/). If there's someone internally with more knowledge on these specific schemas then I'd gladly have them look over it

PR #623, opened 2024-06-24
Solutions + Story scroll

rezrah:
See [Slack thread](https://github.slack.com/archives/C04E44ZPNEA/p1720537193279009) for more info...


nsolerieu (approved):

packages/react/src/recipes/SolutionTemplates/CategoryPage/CategoryPage.tsx
```diff
@@ -0,0 +1,341 @@
...
+  useEffect(() => {
+    setIsLightMode(colorMode === ColorModesEnum.LIGHT)
+  }, [colorMode])
+
+  const handleMode = e => {
+    e.preventDefault()
+    setIsLightMode(!isLightMode)
+  }
+
+  const handleKeyPress = useCallback(
```
joshfarrant, line 57: Is this staying in, or just a dev/debug thing?

packages/react/src/recipes/SolutionTemplates/SolutionPage/SolutionPage.tsx
```diff
@@ -0,0 +1,865 @@
...
+        </SubdomainNavBar.SecondaryAction>
+      </SubdomainNavBar>
+      <AnimationProvider runOnce visibilityOptions={0.2}>
+        <section>
+          {(variant === 'industry' || variant === 'use-case') && (
+            <Box className={styles.hideHorizontalOverflow}>
+              <Grid enableOverlay={enableGridOverlay} className={clsx({[styles.heroVideoGlow]: args.heroVideo})}>
+                <Grid.Column>
+                  <Box marginBlockStart={20}>
+                    <Breadcrumbs>
```
joshfarrant, line 127: The breadcrumbs might need their spacing looking at, it looks like they're disappearing behind the nav bar. I ran into something similar with the parallax hero and applied a bit of margin-top to them to make sure they're visible iirc

rezrah: The breadcrumbs only appear in the industry and use-case variants, not size (which is the one that has parallax enabled). You can see the conditional for it on [line 122](https://github.com/primer/brand/pull/623/files/1e9c3c803987fb1ad7f45725195a237ba85146e4#diff-e4e4c13b20e60d9cd21c463988e14fd7c99bf606e5c2667c9890dabf8fb913fcR122). 

Should appear fine in the other templates though?
Screenshot 2024-07-10 at 14 33 19

joshfarrant: Uploading Untitled.mov…

joshfarrant: This is what I'm seeing (top-left)

https://github.com/user-attachments/assets/eda036fb-0152-4c6a-b0b6-a505b409931a

rezrah: Ah! you're right, thanks for spotting that 👍 . I've added the margin offset here, as it should appear here but not on the overview page.

packages/react/src/river/RiverStoryScroll/RiverStoryScroll.module.css
danielguillan, line 1: This CSS appears to rely heavily on magic numbers, which can make the code difficult to understand and track down what a specfic value represents. Can we consider assigning those values to local variables and adding comments to explain their purpose?

danielguillan (approved):
This looks great! I've added a few non-blocking comments for  RiverStoryScroll. Additionally, I believe the component may not feel quite right on mobile devices, but this is something we can discuss and revisit as a follow-up.

packages/react/src/river/RiverStoryScroll/RiverStoryScroll.tsx
```diff
@@ -0,0 +1,151 @@
...
+   */
+  align?: 'start' | 'end'
+  /**
+   * Apply an alternative image to text column ratio. The default is `50:50`.
+   */
+  imageTextRatio?: '60:40' | '50:50'
+  /**
+   * Disable the scroll effect.
+   */
+  disable?: boolean
```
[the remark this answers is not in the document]
rezrah: Yeah, i'll switch to disabled for consistency. Thanks for calling that out 👍

packages/react/src/river/RiverStoryScroll/RiverStoryScroll.module.css
rezrah: I know 😭... a large part of the mobile CSS is adapted from the Copilot one, which has already had a ton of QA remediation applied to it. I didn't have time to fundamentally change it too much in this PR.

joshfarrant (approved):

PR #624, opened 2024-06-24
Fix local links from docs to Storybook

### LINKED ISSUE #625: Links to Storybook from docs site don't work locally
Some of the component pages on the docs site link to non-existent Storybook stories (fixed in #621). As Storybook runs on a different server locally, all the links from the dev docs site are broken, which makes it much harder to tell when they're actually pointing to the wrong place.

This issue explores possible fixes to this problem to reduce the likelihood of future broken links and generally improve DX.

## Options investigated

- [x] Redirecting the `/brand/storybook` links on local
  - Gatsby have [wildcard redirects](https://www.gatsbyjs.com/docs/how-to/cloud/working-with-redirects-and-rewrites/#wildcard-path-redirects) but they don't support query params
  - [Query param redirects](https://www.gatsbyjs.com/docs/how-to/cloud/working-with-redirects-and-rewrites/#query-param-redirects) are only available on more recent version of Gatsby
  - You could do a check inside `gatsby-browser.js` but that's a level of custom mess this doesn't warrant
- [x] I thought maybe the page or Storybook link component itself could do a check for environment and manipulate the href, but the layout and [storybook link](https://github.com/primer/doctocat/blob/%40primer/gatsby-theme-doctocat%404.7.1/theme/src/components/storybook-link.js) are inside the doctocat package, so that's a no-go.
- [x] I'm considering whether I could write a lightweight test to check there are no broken links
  - There's nothing like this on the site so I'd have to add a test suite or do a dodgy script, which seems overkill for this issue
- [ ] Can we build a Gatsby page at `/brand/storybook` that redirects to localhost:6006, but only in dev?
  #624

pouretrebelle:
## Summary

Fixes #625, other options explored in that issue.

This adds a Gatsby page at `/brand/storybook` to redirect to a locally-running Storybook instance.

On production this is not accessible because all `primer.style/brand/storybook` requests are routed to the private Brand Storybook [via primer.style](https://github.com/primer/primer.style/blob/main/redirects.json#L887-L896).

As far as I know there aren't any non-prod dev environments that actually attach the docs site to the storybook so this won't break any existing functionality.

- Should I put this behind a node env dev flag? Currently on preview deploys of docs it would redirect to localhost instead of 404ing which may be confusing

## Steps to test:

1. Run docs and storybook locally
1. Clicking a Storybook link from a component docs page should redirect to your local Storybook instance

## Contributor checklist:

## Reviewer checklist:


apps/docs/src/pages/brand/storybook.tsx
```diff
@@ -0,0 +1,16 @@
+/**
+ * This page is only used during local development to redirect Storybook links to the Storybook dev server
+ * On production this directory is redirected to the privately hosted Storybook by primer.style
+ */
+
+import {useEffect} from 'react'
+
+const StorybookRedirect = ({location}) => {
+  useEffect(() => {
+    window.location.replace(`http://localhost:6006${location.search}`)
```
pouretrebelle, line 10: Hardcoded port because it's hardcoded in `package.json` and in a bunch of test files

apps/docs/src/pages/brand/storybook.tsx
```diff
@@ -0,0 +1,16 @@
+/**
+ * This page is only used during local development to redirect Storybook links to the Storybook dev server
+ * On production this directory is redirected to the privately hosted Storybook by primer.style
```
rezrah, line 3: @pouretrebelle Has this been verified somehow? ☝️. I can't see anything that would prevent this from applying to prod in this diff, but maybe missed it. It applies to previews too, which indicates it would apply to production too.

pouretrebelle: It would apply to prod, but `primer.style/brand/storybook` [redirects to the SB instance](https://github.com/primer/primer.style/blob/main/redirects.json#L887-L896) using a web.config, see https://github.com/primer/primer.style/pull/358. I could put a `if (process.NODE_ENV !== 'development') return null` in there for safety, so it'd show a blank page rather than redirect to localhost and 404 on previews.

rezrah: Gotcha, thanks. The guard seems like a good idea 👍, just so it doesn't inadvertendly affect any other environments. There's a plan to eventually get rid of that reverse proxy in the web.config, so we might get caught out by that eventually.

rezrah (approved):

PR #629, opened 2024-06-26
Fix FAQGroup autofocus with Alloy issue

stamat:
## Summary
I gated the focus logic via the initialRender ref to move focus on arrow navigation. However, the Alloy actually triggers the render of the FAQGroup component, which causes the initialRender ref to be set to false. This causes the autofocus to the first FAQGroup element and the page is scrolled to it.

This PR changes the logic to gate the focus logic via the hasInteracted state instead of the initialRender ref. This way, the focus logic is only triggered after the user has interacted with the FAQGroup component.

**This issue is only present in development environment** probably due the Alloy dev setup...

## List of notable changes:

- **updated** `FAQGroupProps` type to include `hasInteracted` state because the focus logic is gated via the `hasInteracted` state instead of the `initialRender` ref.

This can only be tested on the dotcom pages where the FAQGroup component is used, for instace `/features/copilot`. Arrow navigation should move the focus correctly and the page should not scroll to the first FAQGroup element on initial render.

- https://github.com/github/marketing-platform-services/issues/3167#issuecomment-2191895803

## Contributor checklist:

## Reviewer checklist:

### DIFF
--- a/.changeset/long-ties-burn.md
+++ b/.changeset/long-ties-burn.md
@@ -1,0 +1,5 @@
+---
+"@primer/react-brand": patch
+---
+
+FAQGroup use interaction gating instead of initialization gating due the Alloy prerender issue
--- a/.changeset/long-ties-burn.md
+++ b/.changeset/long-ties-burn.md
@@ -2,4 +2,4 @@
 "@primer/react-brand": patch
 ---
 
-FAQGroup use interaction gating instead of initialization gating due the Alloy prerender issue
+Updated `FAQGroup` autofocusing conditions
--- a/package-lock.json
+++ b/package-lock.json
@@ -46,7 +46,7 @@
     },
     "apps/storybook": {
       "name": "@primer/brand-storybook",
-      "version": "0.34.0",
+      "version": "0.34.1",
       "license": "MIT",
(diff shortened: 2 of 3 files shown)


.changeset/long-ties-burn.md
```diff
@@ -0,0 +1,5 @@
+---
+"@primer/react-brand": patch
+---
+
```
rezrah, line 4: Updated wording for public change log

```suggestion
Updated `FAQGroup` autofocusing conditions
```

.changeset/long-ties-burn.md
```diff
@@ -0,0 +1,5 @@
...
+FAQGroup use interaction gating instead of initialization gating due the Alloy prerender issue
```
rezrah, line 5: Updated wording for public change log

```suggestion
Updated `FAQGroup` autofocusing conditions
```

rezrah (approved):
👍 

> This issue is only present in development environment probably due the Alloy dev setup...

@stamat the change looks fairly context-agnostic to me, so I think it's fine. Just an FYI for future that we don't wanna put dotcom-specific fixes into Primer Brand, as the library is used outside the monolith too.

PR #634, opened 2024-06-28
VideoPlayer component improvements

joshfarrant:
## Summary

A refactor of the `VideoPlayer` component to make it more modular and customisable. The refactor introduces new hooks and controls to manage video state and provide a more flexible API for customising the video player.

## List of notable changes:

- allow video title to be hidden via the `visuallyHiddenTitle` prop
- allow a custom play overlay to be provided via the `renderPlayOverlay` prop
- refactored video controls into standalone components. Split some of the controls CSS into separate files, but styling would benefit from a more comprehensive refactor in the future
- allow video controls to be customised via the `renderControls` prop
- added new `useVideo` hook and context to manage video state
- added new `useIsElementFullScreen` hook to manage fullscreen state
- added new `useVideoKeypressHandler` hook to add keyboard shortcuts to control video (see updated documentation for specific shortcuts)
- added new `useVideoResizeObserver` hook to consolidate resize logic. When Safari 18 drops later this year this hook can be removed in favour of container queries
- updated custom `useRef` hook to accept `ForwardedRef` types as well
- updated Storybook with more examples
- updated documentation
- updated test suite

- Look and feel of the video player
- Test it and see if you run into any issues
- Check the documentation and Storybook for the new examples
- Make sure the API changes are clear and easy to understand

## Steps to test:

[View the Storybook here](https://primer-2c317f85db-26139705.drafts.github.io/storybook/?path=/story/components-videoplayer-features--with-poster)

## Contributor checklist:

## Reviewer checklist:

## Screenshots:

<table>
<tr>
</tr>
<tr>
<td valign="top">

 </td>
<td valign="top">

</td>
</tr>
</table>


packages/react/src/VideoPlayer/VideoPlayer.stories.tsx
```diff
@@ -29,7 +41,7 @@ export default {
 export const Playground: StoryFn<typeof VideoPlayer> = args => (
   <VideoPlayer {...args}>
     <VideoPlayer.Source src="/example.mp4" type="video/mp4" />
-    <VideoPlayer.Track src="/example.vtt" />
+    <VideoPlayer.Track src="/example.vtt" default />
```
rezrah, line 44: @joshfarrant general comment about storybook.. the examples don't seem to work for me. 

https://github.com/user-attachments/assets/6b986c52-495e-48c9-9cc9-8aa05e2de9e2

Few other little things I noticed there:


- Timestamp is wrapping on both lines, probably due to broken video but we might want to double check it won't do this on longer videos (1hr+)?
  Screenshot 2024-07-10 at 10 36 17


- Could we add the poster to all of the examples please? Looks a little odd when not included.

  Screenshot 2024-07-10 at 10 36 08

joshfarrant: Doh! Sorry, the timestamp wrapping is due to an issue in my previous commit. I reverted a previous commit from the command line but didn't test the change

I'll have a look at the examples as well, they work locally and (I think) they will work when deployed to the prod storybook, it looks like it's just these preview storybooks where they're not working. Leave that one with me.

Once I fix that the videos will look a bit less strange (you'll see the first frame instead of a white box) so we might not need posters there. 

If you pull the branch and run locally you'll be able to see that

joshfarrant: image

Timestamp when > 1 hour btw

apps/docs/content/components/VideoPlayer.mdx
```diff
@@ -42,7 +43,24 @@ TODO: Fix asset loading in environments that don't use `pathPrefix`. This is a t
...
+
+```jsx live
+<VideoPlayer
+  title="GitHub media player"
+  renderControls={() => (
+    <ControlsBar>
+      <PlayPauseButton />
+      <SeekControl />
+    </ControlsBar>
+  )}
```
rezrah, line 62: I'd usually expect these to be flags on the VideoPlayer parent like `showPlayPause` or `showControls`. Particularly as these components [don't seem to take any props](https://github.com/primer/brand/pull/634/files#diff-40d32b8a29dde73e255060785b9239f79dcef575383ee2290479a91ef83ec563R6), I don't think they need to be composable children personally 🤔, but if they did can we stick to name spacing these for consistency please? E.g. `<VideoPlayer.SeekControl />`. Could you elaborate on why you went with this approach vs the two I described above so I'm understanding this correctly.

Right now, the two issues I'm seeing with this approach are:

- We require passing an anonymous inline function inside the return render statement, so you can't take advantage of memoization. We also haven't used this pattern in PB before, so this is a precedent.
- We're surfacing sub-components in the root of the package, some of which could overlap with future component names. E.g. ControlsBar.

  Screenshot 2024-07-12 at 10 05 11

apps/docs/content/components/VideoPlayer.mdx
```diff
@@ -56,41 +74,86 @@ TODO: Fix asset loading in environments that don't use `pathPrefix`. This is a t
...
+<h2>Keyboard shortcuts</h2>
+
+The `VideoPlayer` component supports the following keyboard shortcuts for controlling video playback:
+
+- `Space` or `k`: Toggles between playing and pausing the video.
+- `m`: Toggles between muting and unmuting the video.
+- `f`: Toggles between entering and exiting fullscreen mode.
+- `c`: Toggles closed captions on and off.
+- `ArrowLeft`: Seeks the video backward by 5 seconds.
+- `ArrowRight: Seeks the video forward by 5 seconds.
```
rezrah, line 153:
```suggestion
- `ArrowRight`: Seeks the video forward by 5 seconds.
```

apps/docs/content/components/VideoPlayer.mdx
```diff
@@ -56,41 +74,86 @@ TODO: Fix asset loading in environments that don't use `pathPrefix`. This is a t
...
+- `j`: Seeks the video backward by 10 seconds.
+- `l`: Seeks the video forward by 10 seconds.
+- `ArrowUp` or `ArrowDown`: Increases or decreases the volume by 10%.
```
rezrah, line 156: qq about this one... is this okay from accessibility perspective? Some users may use up/down keys to scroll the page and this overwrites that default behaviour. Were a11y team okay with us adding these particular shortcuts?

packages/react/src/VideoPlayer/VideoPlayer.features.stories.tsx
```diff
@@ -11,14 +12,85 @@ export default {
...
+      <ControlsBar>
+        <PlayPauseButton />
+        <SeekControl />
+      </ControlsBar>
+    )}
+  >
+    <VideoPlayer.Source src="https://primer.github.io/brand/assets/example.mp4" type="video/mp4" />
+    <VideoPlayer.Track src="https://primer.github.io/brand/assets/example.vtt" />
+  </VideoPlayer>
+)
```
rezrah, line 47: It seems like play/pause and seek controls are enabled by default, so this story feels a little redundant. Could we consider flipping this to 'No controls' instead?

packages/react/src/VideoPlayer/VideoPlayer.features.stories.tsx
```diff
@@ -11,14 +12,85 @@ export default {
...
-    <VideoPlayer.Source src="/example.mp4" type="video/mp4" />
-    <VideoPlayer.Track src="/example.vtt" />
+  <VideoPlayer showBranding={false} title="GitHub media player">
+    <VideoPlayer.Source src="https://primer.github.io/brand/assets/example.mp4" type="video/mp4" />
+    <VideoPlayer.Track src="https://primer.github.io/brand/assets/example.vtt" />
+  </VideoPlayer>
+)
+
+export const WithoutTitle = () => (
+  <VideoPlayer showTitle={false} title="GitHub media player">
```
rezrah, line 28: these props `showTitle={false} title="GitHub media player"` appear to be a little at odds with one another. 

I'm inferring that we need to always set a title, but `showTitle` hides it visually? If that's the case, could we change the prop name of `showTitle` to `visuallyHiddenTitle`? We'd need to capture it in the changeset, but `visuallyHidden` is the naming we use for this type of behaviour in Primer.

packages/react/src/VideoPlayer/VideoPlayer.features.stories.tsx
```diff
@@ -11,14 +12,85 @@ export default {
...
+  >
+    <VideoPlayer.Source src="https://primer.github.io/brand/assets/example.mp4" type="video/mp4" />
+    <VideoPlayer.Track src="https://primer.github.io/brand/assets/example.vtt" />
+  </VideoPlayer>
+)
+
+export const WithCustomPlayButton = () => (
+  <VideoPlayer
+    title="GitHub media player"
+    renderPlayOverlay={() => <PlayIcon style={{width: 96, height: 96, opacity: 0.8}} />}
```
rezrah, line 70: This should be the default play button tbh 😅... looks much nicer

packages/react/src/VideoPlayer/VideoPlayer.module.css
```diff
@@ -321,16 +315,16 @@
 
 .VideoPlayer__tooltip {
   position: absolute;
+  left: 50%;
   bottom: 100%;
   z-index: 2;
   width: 10px;
   height: 10px;
```
rezrah, line 322: Looks like there were quite a few hard coded px values here, could we get these cleaned up while we're at this refactor too please? Preference is to use rems where possible.

packages/react/src/VideoPlayer/VideoPlayer.tsx
```diff
@@ -14,195 +13,95 @@ import '@primer/brand-primitives/lib/design-tokens/css/tokens/functional/compone
...
-        poster={poster}
-        className={clsx(styles.VideoPlayer, className)}
-        onPlay={handleOnPlay}
-        onPause={handleOnPause}
-        onPlaying={handleOnPlaying}
-        onLoadedMetadata={handleOnLoadedMetadata}
-        onTimeUpdate={handleOnTimeUpdate}
-        {...rest}
-      >
+      {/* eslint-disable-next-line jsx-a11y/media-has-caption */}
```
rezrah, line 58: This will need to be removed as there's an externally-managed static analysis audit that runs on this repo to check if we've disabled any `jsx-a11y` rules.

packages/react/src/VideoPlayer/VideoPlayer.tsx
```diff
@@ -14,195 +13,95 @@ import '@primer/brand-primitives/lib/design-tokens/css/tokens/functional/compone
...
-        <track kind="captions" />
       </video>
-      {title && (
-        <div className={styles.VideoPlayer__title}>
-          {branding && <MarkGithubIcon size={48} />}
-          <Text size="500" weight="medium" className={styles.VideoPlayer__controlTextColor}>
+      <div className={styles.VideoPlayer__title}>
+        {showBranding && <MarkGithubIcon size={40} />}
+        {showTitle && (
+          <Text size="400" weight="medium" className={styles.VideoPlayer__controlTextColor}>
```
rezrah, line 65: Nice 👍, looks a lot better smaller

packages/react/src/VideoPlayer/VideoPlayer.tsx
```diff
@@ -14,195 +13,95 @@ import '@primer/brand-primitives/lib/design-tokens/css/tokens/functional/compone
...
-            </svg>
-          </span>
         )}
-        <span className="visually-hidden">Play</span>
+      </div>
+      <button
+        className={styles.VideoPlayer__playButton}
+        onClick={() => {
+          isPlaying ? pause() : play()
+        }}
```
rezrah, line 74: Seeing this in a few places, but could you please hoist these inline anonymous functions outside of the return please into their own named handlers. They should be passed through a `useCallback` function too.

packages/react/src/VideoPlayer/components/Captions/Captions.module.css
```diff
@@ -0,0 +1,17 @@
...
+  margin: var(--brand-VideoPlayer-closedCaption-margin);
+  display: grid;
+  justify-content: center;
+  pointer-events: none;
+}
+
+.Captions__text {
+  color: var(--base-color-scale-white-0);
+  text-align: center;
+  background: rgba(0, 0, 0, 0.75);
```
rezrah, line 13: Could we make this a CSS var / token please?

packages/react/src/VideoPlayer/components/VideoTooltip/VideoTooltip.tsx
```diff
@@ -0,0 +1,17 @@
+import React from 'react'
+import clsx from 'clsx'
+
+import {Text} from '../../../Text'
+import styles from '../../VideoPlayer.module.css'
+
+type VideoTooltipProps = React.HTMLAttributes<HTMLDivElement>
+
+export const VideoTooltip = ({children, className, ...rest}: VideoTooltipProps) => (
```
rezrah, line 9: Is it possible to use our [Tooltip](https://primer-2c317f85db-26139705.drafts.github.io/storybook/?path=/story/components-tooltip--default) component here instead of maintaining a custom one? It uses popover.

packages/react/src/VideoPlayer/components/VideoTooltip/VideoTooltip.tsx
```diff
@@ -0,0 +1,17 @@
...
+  <div className={clsx(styles.VideoPlayer__tooltip, className)} aria-hidden="true" {...rest}>
```
rezrah, line 10: Should this be `aria-hidden`?

packages/react/src/hooks/useRef.ts
```diff
@@ -1,15 +1,20 @@
...
-export function useProvidedRefOrCreate<TRef>(providedRef?: React.RefObject<TRef>): React.RefObject<TRef> {
-  const createdRef = useRef<TRef>(null)
 
-  return providedRef ?? createdRef
+export function useProvidedRefOrCreate<T>(providedRef?: null | RefObject<T> | ForwardedRef<T>): RefObject<T> {
+  const createdRef = useRef<T>(null)
+
+  if (providedRef) {
+    return typeof providedRef === 'function' ? createdRef : providedRef
+  }
```
rezrah, line 17: Did you run into type issues with this hook, prompting this change? I don't think we have test coverage on this file, but we use it in a few places. Are we confident this won't cause regressions as the logic here is now different.

apps/docs/content/components/VideoPlayer.mdx
joshfarrant: I went for the more composable approach as, imo, that offers a tonne of flexibility without requiring us to expand the `VideoPlayer` component's API too much now (with the `showPlayPause`, `showControls`, etc props) and defends against future use-cases we might not have considered at this point, eg attaching analytics events to the controls, modifying or adding additional controls, etc.

As with anything, I'm not rigid on this approach at all and if you really don't like it I can swap it out, but it was very much a conscious decision to take this approach and not just the first thing that came to mind. I do very much appreciate the flexibility that this pattern offers, and my experience is that it requires less maintenance in the long-run.

On your two bullet points:

> We require passing an anonymous inline function inside the return render statement, so you can't take advantage of memoization. We also haven't used this pattern in PB before, so this is a precedent.

Very true, I've not seen this explicit pattern in any other places in our codebase (we do have a couple of `renderXyz()` functions which return JSX, but none which are taken as props afaict) but I don't see this approach as particularly radical.

If someone wanted to memoize that function then they could.

```tsx
const MyComponent = () => {
  const controls = useCallback(() => (
    <ControlsBar>
      <PlayPauseButton />
      <SeekControl />
    </ControlsBar>
  ), [])

  return (
    <VideoPlayer
      renderControls={controls}
    />
  );

}
```

But to be honest, in that case I'd recommend that they just pull it out of the parent component altogether, removing the need for any memoization overhead at all. Then, their controls are defined outside of the component in their own function, a function which returns some JSX. A function which returns JSX is a component, so they'd end up with this.

```tsx
const Controls = () => (
  <ControlsBar>
    <PlayPauseButton />
    <SeekControl />
  </ControlsBar>
);

const MyComponent = () => {
  return (
    <VideoPlayer
      renderControls={Controls}
    />
  );
}
```

I'd be inclined to say that that would be an over-optimisation, but the option is there for those who want it.

I appreciate that elsewhere in the codebase this behaviour may be implemented more like this:

```tsx
const MyComponent = () => {
  return (
    <VideoPlayer>
      <VideoPlayer.Controls>
        <ControlsBar>
          <PlayPauseButton />
          <SeekControl />
        </ControlsBar>
      </VideoPlayer.Controls>
    </VideoPlayer>
  );
}
```

And I'm happy to go with that approach instead if you'd prefer. Both approaches achieve the same thing as far as I understand, and are really just a workaround for the fact that React doesn't have a standard way of supporting [slots](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/slot) as you'd see in Vue, Svelte, Web Components, etc.

I chose to use the `renderControls` function instead as that doesn't require iterating over the children in the `VideoPlayer` component and so in theory should be more performant and require less code to achieve the same effect.

> We're surfacing sub-components in the root of the package, some of which could overlap with future component names. E.g. ControlsBar.

More than happy to namespace these to the `VideoPlayer` component, that was an oversight on my part, thanks for the suggestion 🙂

apps/docs/content/components/VideoPlayer.mdx
joshfarrant: Great spot, thanks!

packages/react/src/VideoPlayer/VideoPlayer.features.stories.tsx
```diff
@@ -11,14 +12,85 @@ export default {
...
+
+export const HideControlsUntilPlaying = () => (
+  <VideoPlayer
+    title="GitHub media player"
+    renderControls={(controlsProps, {isPlaying}) => (
+      <Controls
+        style={{
+          opacity: isPlaying ? 1 : 0,
+          transition: 'opacity var(--brand-animation-duration-default) var(--brand-animation-easing-default)',
+        }}
```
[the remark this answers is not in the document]
joshfarrant: Yea absolutely. This example was mainly intended to show just how much control (bad pun, sorry) the implementor has over the controls. In practise I'd imagined that they'd toggle classes rather than use inline styles, but I wasn't sure how styling was set up in storybook.

Would you prefer that I added `hideControlsUnilPlaying` (I'll find a better name) as a prop, or that I update this example to remove the inline styling and instead toggle classes based on `isPlaying`?

packages/react/src/VideoPlayer/VideoPlayer.features.stories.tsx
joshfarrant: Good point, it's poorly named and not very clear. I was trying to show that the user can mix and match the controls by omitting some, including some, and even writing their own. I think that's a valuable thing to demo, but I'll look at making that clearer and also add in a no controls example too

packages/react/src/VideoPlayer/VideoPlayer.features.stories.tsx
joshfarrant: Great suggestion, thank you

packages/react/src/VideoPlayer/VideoPlayer.features.stories.tsx
joshfarrant: I agree 😄

Reckon I should change it?

packages/react/src/VideoPlayer/VideoPlayer.tsx
joshfarrant: Good point.

I can add `<track kind="captions" />` back in, but it's a bit annoying as it's completely redundant.

Really, we need a linter rule to make sure that you can't use `VideoPlayer` without passing a captions track instead, or we iterate through the children each render and yell at the user if they don't pass a track 😅 (I'm not really advocating for that, don't worry)

packages/react/src/VideoPlayer/VideoPlayer.tsx
joshfarrant: Ah, thank you, I meant to tidy those up.

I've made the changes in 469903a60b588ad03519c97df644a120374c4b9d which ditches the anonymous function entirely so we don't need a useCallback, lmk if that works for you

apps/docs/content/components/VideoPlayer.mdx
joshfarrant: Resolved in 248ae2817b68adcdc9f2f920decfe7a88c37999c

packages/react/src/VideoPlayer/VideoPlayer.features.stories.tsx
joshfarrant: Resolved in b1d23e389a959cc9ff4802e07f9bc05645f7771d

packages/react/src/VideoPlayer/VideoPlayer.tsx
joshfarrant: Resolved in 3d65fee00e176151ec35c579d2bfe1f6a2ec3e31

packages/react/src/VideoPlayer/components/VideoTooltip/VideoTooltip.tsx
joshfarrant: Great suggestion, thanks. I'm looking into this now

packages/react/src/VideoPlayer/components/VideoTooltip/VideoTooltip.tsx
joshfarrant: Hopefully we can ditch this entirely and just use `Tooltip`, in which case this file can be deleted

packages/react/src/hooks/useRef.ts
joshfarrant: Yea I did run into some type issues, and it looks like we've run into the same ones elsewhere too. In all of the other places that we use `useProvidedRefOrCreate` we override the types with something like

```tsx
useProvidedRefOrCreate(ref as RefObject<HTMLDivElement>)
```

which suggests to me that the types of `useProvidedRefOrCreate` aren't right to start with, which is why I wanted to improve the types.

It's a good point that I shouldn't change the implementation though, and to be honest I probably shouldn't touch it in this PR at all.

I'll dig a bit deeper into this and raise a follow-up PR to address this

joshfarrant: Reverted in `8102bdee4e858f8c34ef957d0d677e023601cbd8`

apps/docs/content/components/VideoPlayer.mdx
joshfarrant: This is a good question and I went back and forth on this based on reading I'd done.

I think for now I'm going to disable the keyboard interaction until I've had a discussion with the a11y team

packages/react/src/VideoPlayer/components/Captions/Captions.module.css
joshfarrant: Resolved in c891b32a48dd0f6c452a2af6e50d5d398384923a

packages/react/src/VideoPlayer/components/VideoTooltip/VideoTooltip.tsx
joshfarrant: This works, but there's a bit of weirdness with the popover positioning. I'm not 100% sure yet, but I think we're fighting over the popover's position. The range input is trying to position it by the cursor, and the popover is trying to position itself centrally. 

Would you be happy for this to go in as part of a subsequent PR?

joshfarrant: (A video of a video player is slightly confusing 😅)

packages/react/src/VideoPlayer/components/VideoTooltip/VideoTooltip.tsx
joshfarrant: I don't know whether this will have had sign-off previously? I don't think I've really modified this file, just moved it.

apps/docs/content/components/VideoPlayer.mdx
joshfarrant: Resolving second point ad5863aa264c892bcb1864556e19d512f0913d2b

packages/react/src/VideoPlayer/VideoPlayer.features.stories.tsx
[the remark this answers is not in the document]
rezrah: > I wasn't sure how styling was set up in storybook. 

You can create a new co-located file called `VideoPlayer.features.module.css` and import that into your Storybook file. It's then separate from the main stylesheet, and we use this approach for other components too.

> Would you prefer that I added hideControlsUnilPlaying (I'll find a better name) as a prop

👍 Yeah that would be great. It's a cool feature, so we should treat it as such by giving it it's own prop. Would you agree, or see issues doing that?

packages/react/src/VideoPlayer/VideoPlayer.features.stories.tsx
rezrah: I think so... 

@simmonsjenna @danielguillan Cool if we change the main play icon?

 
<table>
<tr>
</tr>
<tr>
<td valign="top">

Screenshot 2024-07-16 at 16 06 03


 </td>
<td valign="top">

Screenshot 2024-07-16 at 16 06 08


</td>
</tr>
</table>

packages/react/src/VideoPlayer/components/VideoTooltip/VideoTooltip.tsx
rezrah: > Would you be happy for this to go in as part of a subsequent PR?

Yeah, sounds good 👍

packages/react/src/VideoPlayer/components/VideoTooltip/VideoTooltip.tsx
rezrah: > I don't know whether this will have had sign-off previously?

I don't think it did. If you're keeping the custom tooltip though, are you still removing the aria-hidden? I wasn't sure what it was doing or why it was needed.

packages/react/src/VideoPlayer/VideoPlayer.features.stories.tsx
simmonsjenna: @rezrah @joshfarrant Agreed — let's change it.

simmonsjenna: @rezrah @joshfarrant As a follow up, the newer one is a bit large — can we resize it across certain key breakpoints so it's more proportional as the video player gets smaller?

apps/docs/content/components/VideoPlayer.mdx
joshfarrant: I've modified my approach in 4873e4713bad9c9ce1780008cf36c994d2f7053d, let me know if you prefer it this way

apps/docs/content/components/VideoPlayer.mdx
joshfarrant: Resolved in 9944caca8bae4068b0ed3b6d8b2ced8b5308c58b

packages/react/src/VideoPlayer/VideoPlayer.features.stories.tsx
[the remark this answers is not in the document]
joshfarrant: Resolved in 4873e4713bad9c9ce1780008cf36c994d2f7053d

packages/react/src/VideoPlayer/VideoPlayer.features.stories.tsx
joshfarrant: Resolved in 4873e4713bad9c9ce1780008cf36c994d2f7053d

packages/react/src/VideoPlayer/VideoPlayer.module.css
joshfarrant: Resolved in 4873e4713bad9c9ce1780008cf36c994d2f7053d

packages/react/src/VideoPlayer/components/VideoTooltip/VideoTooltip.tsx
joshfarrant: Cool, thanks for the info. Yes it's removed 👍

apps/docs/content/components/VideoPlayer.mdx
rezrah: This looks good, thanks for explaining the thought process too @joshfarrant 👍 

To address a few of your points...

> defends against future use-cases we might not have considered at this point, eg attaching analytics events to the controls, modifying or adding additional controls, etc.

Yep, agree. In that instance, it would be preferable to use namespaced children like `VideoPlayer.SeekControl` as we can maintain sufficient guardrails internally to the component, whereas an anonymous function would behave more like an escape hatch (e.g. trailingComponent)

> I've not seen this explicit pattern in any other places in our codebase (we do have a couple of renderXyz() functions which return JSX, but none which are taken as props afaict) 

We haven't - to date - made a `renderFoo` slot available on the public API. Though this is more due to naming. Our version of this is leadingComponent and trailingComponent, and we try to be really careful about when/where to use it. This is because we've historically seen a lot of reliance on escape hatches, to the point where we couldn't keep up with all the different variations of components.

.changeset/kind-foxes-begin.md
```diff
@@ -0,0 +1,13 @@
+---
+'@primer/react-brand': minor
+---
+
+Refactored `VideoPlayer` component to make it more modular and customisable.
+
+- The `showTitle` prop has been renamed to `visuallyHiddenTitle` and inverted.
```
rezrah, line 7: Let's flag clearly this is a breaking change. This is how we usually do it...

```suggestion

⚠️ This update contains a breaking change.

Refactored `VideoPlayer` component to make it more modular and customisable.

- The `showTitle` prop has been renamed to `visuallyHiddenTitle` and inverted.
```

.changeset/kind-foxes-begin.md
```diff
@@ -0,0 +1,13 @@
...
+  - This prop hides the title visually, but keeps it accessible to screen readers.
+  - Where you prviosuly passed `showTitle={false}`, you should now pass `visuallyHiddenTitle={true}`.
+- The `branding` prop has been renamed to `showBranding`.
+- Video controls are now modular and can be customised via the `renderControls` prop. This allows you to hide or show specific controls, or provide your own custom controls. See Storybook and the updated documentation for more information.
```
rezrah, line 11: This needs to be revised now that the API has changed?

.changeset/kind-foxes-begin.md
```diff
@@ -0,0 +1,13 @@
...
+- A custom play overlay can now be provided via the `renderPlayOverlay` prop. This allows you to customise the appearance of the overlay that appears when the video is paused. See Storybook and the updated documentation for more information.
+- Videos now support YouTube-like keyboard shortcuts to control video playback. See the updated documentation for specific shortcuts.
```
rezrah, line 13: Have you removed the shortcuts now?

apps/docs/content/components/VideoPlayer.mdx
rezrah: Just noticed we've dropped the ability to customize the Play icon now that we've moved to a prop based flag. I think that was actually a nice feature, but should be done as a named child with an `icon` prop like we do in `Card`, `Pillar` and `EyebrowBanner`.

E.g.

```jsx
<VideoPlayer>
  <VideoPlayer.PlayPauseControl icon={<svg />} />
</VideoPlayer>
```

apps/docs/content/components/VideoPlayer.mdx
```diff
@@ -42,12 +38,55 @@ TODO: Fix asset loading in environments that don't use `pathPrefix`. This is a t
 ## Without branding
 
 ```jsx live
-<VideoPlayer branding={false} title="GitHub media player">
+<VideoPlayer hasBranding={false} title="GitHub media player">
+  <VideoPlayer.Source src="https://primer.github.io/brand/assets/example.mp4" />
+  <VideoPlayer.Track src="https://primer.github.io/brand/assets/example.vtt" />
+</VideoPlayer>
+```
```
rezrah, line 45: FYI, I'm still seeing the branding here

apps/docs/content/components/VideoPlayer.mdx
```diff
@@ -42,12 +38,55 @@ TODO: Fix asset loading in environments that don't use `pathPrefix`. This is a t
...
+  <VideoPlayer.Source
+    src="https://primer.github.io/brand/assets/example.mp4"
+    type="video/mp4"
+  />
+  <VideoPlayer.Track src="https://primer.github.io/brand/assets/example.vtt" />
+</VideoPlayer>
+```
+
+## Minimal
+
```
rezrah, line 89: Missing the example for this one...

Screenshot 2024-07-17 at 15 04 54

packages/design-tokens/src/tokens/functional/components/video-player/base.json
```diff
@@ -6,20 +6,23 @@
           "value": "0 auto 15px auto"
         },
         "text": {
+          "color": {
+            "value": "var(--base-color-scale-white-0)"
+          },
+          "backgroundColor": {
```
rezrah, line 12: Just noticed this, the preference in Primitives v8 is to use `bgColor`

packages/design-tokens/src/tokens/functional/components/video-player/base.json
```diff
@@ -6,20 +6,23 @@
           "value": "0 auto 15px auto"
         },
         "text": {
+          "color": {
```
rezrah, line 9: and this one would be `fgColor`

packages/design-tokens/src/tokens/functional/components/video-player/base.json
```diff
@@ -6,20 +6,23 @@
...
         "height": {
-          "value": "24px"
+          "value": "96px"
         },
         "width": {
-          "value": "24px"
-        },
-        "padding": {
-          "value": "25px"
+          "value": "96px"
```
rezrah, line 25:
```suggestion
          "value": "var(--base-size-96)"
```

.changeset/kind-foxes-begin.md
joshfarrant: Thanks, good spot. Resolved in f082be6b29bb8a989630051920297aeba0122f52

.changeset/kind-foxes-begin.md
joshfarrant: Yes I have for now, I've updated the changeset in f082be6b29bb8a989630051920297aeba0122f52

apps/docs/content/components/VideoPlayer.mdx
joshfarrant: Oops, good spot, thank you!

Fixed in 054a5dbd7793479e803139843bbd5a8f947026d0

apps/docs/content/components/VideoPlayer.mdx
joshfarrant: Thanks!

44b758df161f2d3883c39cf3e20df8d8064dd45a

packages/react/src/VideoPlayer/VideoPlayer.tsx
```diff
@@ -14,195 +27,109 @@ import '@primer/brand-primitives/lib/design-tokens/css/tokens/functional/compone
...
-        videoRef={videoRef}
-        videoWrapperRef={videoWrapperRef}
-        totalTime={totalTime}
-        currentTime={playedTime}
-        playing={playing}
-        trackInformation={trackInformation}
-      />
+      <div className={styles.VideoPlayer__controls}>
+        {ccEnabled ? <Captions /> : null}
+        {hideControls ? null : (
```
rezrah, line 92: Nit but would this be more readable?

```suggestion
        {!hideControls && (
```

packages/design-tokens/src/tokens/functional/components/video-player/base.json
joshfarrant: Thanks! 947c395b5ba093459009c0c08adef9f67a31c797

packages/react/src/VideoPlayer/components/PauseIcon/PauseIcon.tsx
```diff
@@ -0,0 +1,8 @@
+import React, {type SVGAttributes} from 'react'
+
+export const PauseIcon = (props: SVGAttributes<SVGSVGElement>) => (
+  <svg xmlns="http://www.w3.org/2000/svg" width="24" height="24" viewBox="0 0 24 24" fill="none" {...props}>
```
rezrah, line 4: Is this an Octicon @joshfarrant? If so, can we use the one from the `@primer/octicons-react` package here instead of  hardcoding our own?

packages/react/src/VideoPlayer/components/SeekControl/SeekControl.tsx
```diff
@@ -0,0 +1,77 @@
...
+  return (
+    <div className={clsx(styles['VideoPlayer__seek'], className)} {...rest}>
+      <Range
+        type="range"
+        min="0"
+        max={duration}
+        step={0.0001}
+        onInput={e => {
+          seek(e.currentTarget.valueAsNumber)
+        }}
```
rezrah, line 62: Just an FYI that there's another inline, anonymous function here...

packages/react/src/VideoPlayer/index.ts
```diff
@@ -1 +1,2 @@
 export * from './VideoPlayer'
+export * from './components'
```
rezrah, line 2: Can we keep these as private components please? If we do this, these sub components will be surfaced in the root of the package without namespacing [because of this](https://github.com/primer/brand/blob/main/packages/react/src/index.ts#L35). If we need to expose certain components, it should be via the VideoPlayer Object.

apps/docs/content/components/VideoPlayer.mdx
```diff
@@ -42,12 +38,55 @@ TODO: Fix asset loading in environments that don't use `pathPrefix`. This is a t
...
+## With visually hidden title
+
+```jsx live
+<VideoPlayer visuallyHiddenTitle title="GitHub media player">
   <VideoPlayer.Source src="https://primer.github.io/brand/assets/example.mp4" />
   <VideoPlayer.Track src="https://primer.github.io/brand/assets/example.vtt" />
 </VideoPlayer>
 ```
 
+## With custom controls
```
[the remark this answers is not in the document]
rezrah: Yeah that sounds fine. Though not sure we'd want to advertise using a video player without controls... realistically wouldn't you just use the Minimal appearance and hide the track too? No strong opinions though, so feel free to leave in and rename it.

packages/react/src/VideoPlayer/VideoPlayer.tsx
joshfarrant: 37031f7e2b921e6708552ce76d27952885efdd4c

packages/react/src/VideoPlayer/components/PauseIcon/PauseIcon.tsx
joshfarrant: I'm not sure where this came from tbh, it's just moved over from what the VideoPlayer was using before. I'll have a look

packages/react/src/VideoPlayer/VideoPlayer.module.css
```diff
@@ -42,32 +43,25 @@
   left: 0;
   width: 100%;
   height: 100%;
-  background: transparent;
+  background: #00000033;
```
rezrah, line 46: Can we add a TODO comment on this hex value, indicating that we should switch to token once v8 migration is complete

packages/react/src/VideoPlayer/components/PauseIcon/PauseIcon.tsx
joshfarrant: Ah yea now I remember, there isn't a simple _Play_ Octicon and there's no pause one at all, hence why these were made in the first place.

packages/react/src/VideoPlayer/index.ts
joshfarrant: Oversight when I switched to namespacing them, good catch thanks

4f99158cb9e13673c7c8aecd345ab8a41b2f284d

packages/react/src/VideoPlayer/VideoPlayer.module.css
joshfarrant: 👍 0b67ffb8e05b4096a9abab5a10512585434064eb

apps/docs/content/components/VideoPlayer.mdx
joshfarrant: Previously I had the `renderPlayOverlay` on the `VideoPlayer` which was just used for the full-size play icon while the video was paused. 

I'd suggest that this — along with per-control icon customisation as you mentioned there — be brought in at a later point

apps/docs/content/components/VideoPlayer.mdx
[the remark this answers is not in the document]
joshfarrant: Left it in and renamed in 82d1165c380039a08d54c372db8210f3bc8a7765, otherwise we've got no examples of any of the `show[Control]` props

packages/react/src/VideoPlayer/components/PlayIcon/PlayIcon.tsx
```diff
@@ -0,0 +1,10 @@
+import React, {type SVGAttributes} from 'react'
+
+export const PlayIcon = (props: SVGAttributes<SVGSVGElement>) => (
+  <svg xmlns="http://www.w3.org/2000/svg" width="24" height="24" viewBox="0 0 24 24" fill="none" {...props}>
```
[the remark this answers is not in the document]
joshfarrant: See https://github.com/primer/brand/pull/634#discussion_r1681175851

packages/react/src/VideoPlayer/components/PauseIcon/PauseIcon.tsx
rezrah: The official one is this https://primer.style/foundations/icons/play-24

I think we should keep it custom for now, and follow up with a request to Octicons working group about a play icon that doesn't have the rounded border. In long term, we don't want to maintain any of our own icons.

cc. @simmonsjenna @danielguillan for second opinions... the official Octicon has a rounded border, so quite different to custom one used in video player right now. Let us know if you have have a preference.

packages/react/src/VideoPlayer/hooks/useVideoResizeObserver.ts
```diff
@@ -0,0 +1,46 @@
...
+}) => {
+  const [isSmall, setIsSmall] = useState(false)
+
+  useEffect(() => {
+    const handleResize = () => {
+      if (!videoWrapperRef.current) return
+
+      const breakpoint = videoWrapperRef.current.getBoundingClientRect().width
+
+      const isSmallVideo = Boolean(breakpoint && breakpoint < 650)
```
rezrah, line 19: Just spotted this... our small breakpoint is actually `544`...  you can either pull in the constant from the tokens package or hardcode it, but shouldn't be 650 unless there's a good reason for it (let's leave comment).

Screenshot 2024-07-18 at 10 49 05

apps/docs/content/components/VideoPlayer.mdx
```diff
@@ -42,12 +38,70 @@ TODO: Fix asset loading in environments that don't use `pathPrefix`. This is a t
...
+  showControlsWhenPaused={false}
+>
+  <VideoPlayer.Source
+    src="https://primer.github.io/brand/assets/example.mp4"
+    type="video/mp4"
+  />
+  <VideoPlayer.Track src="https://primer.github.io/brand/assets/example.vtt" />
+</VideoPlayer>
+```
+
```
rezrah, line 104: Just thought of another example to add (as a follow up or before this merges): Programmatic play/pause. This is a really common use-case, and we use this regularly with Youtube's embed API, so would be great to share how to do it with this (hook). If we don't do it now, can we add to follow up tasklist please? 🙏

packages/react/src/VideoPlayer/hooks/useVideoResizeObserver.ts
joshfarrant: This is definitely one that will benefit from a container query in the future as my guess is its going to be rare that videos are rendered at full width

apps/docs/content/components/VideoPlayer.mdx
joshfarrant: I'll add it to the list, thanks!

Issue #612, opened 2024-06-17
[Feature Request] Add ability to attach analytics attributes to FAQ tab buttons


joshvogel17:
**Problem**
We are currently trying to attach analytics attributes to all clickable elements in the SEO templates. While verifying the template, it was noticed the FAQ tab buttons do not have a way to attach analytics to them.

![Image](https://github.com/user-attachments/assets/113b5075-d3f7-42e5-ba9a-94b06f1892c3)

joshfarrant:
Hey @joshvogel17 👋 Thanks for the feature request!

This has been added and will ship in the next Primer Brand release 🚀

PR #659, opened 2024-07-23
Add `tabAttributes` prop to `FAQGroup` component

joshfarrant:
## Summary

Adds `tabAttributes` prop to `FAQGroup` component. This prop is used to set arbitrary attributes on the tabs rendered by the `FAQGroup` component.

For example, the below code will add `data-analytics="tab-0"` to the first tab and `data-analytics="tab-1"` to the second tab.

```tsx
<FAQGroup
  tabAttributes={(children, index) => ({
    'data-analytics': `tab-${index}`,
  })}
>
  <FAQGroup.Heading>Frequently asked questions</FAQGroup.Heading>
  <FAQ>
    <FAQ.Heading>Using GitHub Enterprise</FAQ.Heading>
    <FAQ.Item>...</FAQ.Item>
    <FAQ.Item>...</FAQ.Item>
    <FAQ.Item>...</FAQ.Item>
  </FAQ>

  <FAQ>
    <FAQ.Heading>About GitHub Enterprise</FAQ.Heading>
    <FAQ.Item>...</FAQ.Item>
    <FAQ.Item>...</FAQ.Item>
    <FAQ.Item>...</FAQ.Item>
  </FAQ>
</FAQGroup>
```

## List of notable changes:

- Adds `tabAttributes` prop to FAQGroup
- Adds `tabAttributes` tests to FAQGroup test suite
- Removes some unnecessary `async/await` usage from FAQGroup test suite
- Updates FAQGroup documentation

Ensure you're happy with the API, updated docs, and updated tests

- https://github.com/primer/brand/issues/612
- [API exploration](https://www.figma.com/board/FJhHnM9BznUtZ9XHCNLh2T/Brand---FAQGroup-analytics-props?node-id=215-185&t=53hWTCQwoeWUHYIx-11)

## Contributor checklist:

## Reviewer checklist:

### DIFF
--- a/apps/docs/content/components/FAQ/react.mdx
+++ b/apps/docs/content/components/FAQ/react.mdx
@@ -112,8 +112,14 @@ import {FAQ, FAQGroup} from '@primer/react-brand'
 
 Use `FAQGroup` to display multiple `FAQ` components together.
 
+Use `tabAttributes` to add arbitrary attributes to the tabs. This can be useful for tracking analytics or adding custom attributes.
+
 ```jsx live
-<FAQGroup>
+<FAQGroup
+  tabAttributes={(children, i) => ({
+    'data-analytics': `faq-tab-${i}`,
+  })}
+>
   <FAQGroup.Heading>
     Frequently asked <br /> questions
   </FAQGroup.Heading>
--- a/apps/docs/content/components/FAQ/react.mdx
+++ b/apps/docs/content/components/FAQ/react.mdx
@@ -8,13 +8,11 @@ description: Use the FAQ component to display content in a Q&A format.
 ---
 
 import ComponentLayout from '../../../src/layouts/component-layout'
-export default ComponentLayout
 
-import InlineCode from '@primer/gatsby-theme-doctocat/src/components/inline-code'
 import {HeadingTags} from '@primer/react-brand'
-import {Box as Container} from '@primer/react'
-import {PropTableValues} from '../../../src/components'
 import {Label} from '@primer/react'
+import {PropTableValues} from '../../../src/components'
+export default ComponentLayout
 
 ```js
 import {FAQ, FAQGroup} from '@primer/react-brand'
--- a/apps/docs/content/components/FAQ/react.mdx
+++ b/apps/docs/content/components/FAQ/react.mdx
@@ -433,12 +433,13 @@ render(<App />)
 
 ### FAQGroup
 
-| Name        | Type                                                                                                                           | Default | Description                              |
-| :---------- | :----------------------------------------------------------------------------------------------------------------------------- | :-----: | :--------------------------------------- |
-| `className` | `string`                                                                                                                       |         | Sets a custom class                      |
-| `id`        | `string`                                                                                                                       |         | Sets a custom id                         |
-| `ref`       | `React.RefObject`                                                                                                              |         | Forward a Ref to the underlying DOM node |
-| `children`  | <PropTableValues values={[<a href="#faq-required">FAQ</a>, <a href="#faqgroup-heading">FAQGroup.Heading</a>]} addLineBreaks /> |         | Root element for the FAQGroup component. |
+| Name            | Type                                                                                                                           | Default | Description                                                                   |
+| :-------------- | :----------------------------------------------------------------------------------------------------------------------------- | :-----: | :---------------------------------------------------------------------------- |
+| `className`     | `string`                                                                                                                       |         | Sets a custom class                                                           |
+| `id`            | `string`                                                                                                                       |         | Sets a custom id                                                              |
+| `ref`           | `React.RefObject`                                                                                                              |         | Forward a Ref to the underlying DOM node                                      |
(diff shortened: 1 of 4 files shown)


apps/docs/content/components/FAQ/react.mdx
```diff
@@ -433,12 +431,13 @@ render(<App />)
...
+| `className`     | `string`                                                                                                                       |         | Sets a custom class                                                           |
+| `id`            | `string`                                                                                                                       |         | Sets a custom id                                                              |
+| `ref`           | `React.RefObject`                                                                                                              |         | Forward a Ref to the underlying DOM node                                      |
+| `children`      | <PropTableValues values={[<a href="#faq-required">FAQ</a>, <a href="#faqgroup-heading">FAQGroup.Heading</a>]} addLineBreaks /> |         | Root element for the FAQGroup component.                                      |
+| `tabAttributes` | `(children: ReactElement, index: number) => Record<string, unknown>`                                                           |         | Spreads the returned attributes onto the tab that's rendered by the FAQGroup. |
```
rezrah, line 440:
```suggestion
| `tabAttributes` | `(children: ReactElement, index: number) => Record<string, string>`                                                           |         | Spreads the returned attributes onto the tab that's rendered by the FAQGroup. |
```

packages/react/src/FAQ/FAQGroup.tsx
```diff
@@ -73,8 +74,13 @@ function _FAQGroup({children, id, defaultSelectedIndex = 0, ...rest}: FAQGroupPr
         child => React.isValidElement(child) && child.type === FAQ.Heading,
       )
 
+      const tabContents = React.isValidElement(GroupHeadingChild) && GroupHeadingChild.props.children
+
+      const providedTabAttributes = tabAttributes?.(tabContents, index)
+
       return (
         <Button
+          {...providedTabAttributes}
```
rezrah, line 83: Was it intended to put this at the top vs bottom? I.e. to avoid overriding the default values? On one hand, I love the intrinsic guardrails, but also wondering if this is going to be confusing for anyone that is trying to unsuccessfully set override props

apps/docs/content/components/FAQ/react.mdx
rezrah: Could we also add an example of how to use this prop to either Storybook or Documentation please. So that we have a reference point for the user.

packages/react/src/FAQ/FAQGroup.tsx
joshfarrant: Yes it was intentional and I was just erring on the side of guardrails. 

I'm happy moving it to the bottom (or even to somewhere in the middle) if you think there's value in allowing some/all props to be overriden

apps/docs/content/components/FAQ/react.mdx
rezrah: > as users may want to pass values other than strings

Could you provide a few examples please? I would have thought attributes always need to interpolate back to strings?

Or maybe you're referring to event props? In which case can we be explicit about the exact type here. I.e. `() => void`. I think it would help to be a little clearer in the prop table.

packages/react/src/FAQ/FAQGroup.tsx
rezrah: all good, let's leave it like this. thanks for clarifying.. 👍

apps/docs/content/components/FAQ/react.mdx
joshfarrant: Attributes technically do always need to be strings, but as we're working with props in React-land then really the values in this object can be any type.

I don't know of any specific examples, but `unknown` feels like the most accurate type as, under the hood, the user can pass anything they like and it will get forwarded to the `Button`. Even if we make it `Record<string, string>` then the user can still pass whatever they like with a `@ts-ignore` and it will work. I think the fact that it will work makes me feel `unknown` felt accurate. If we were enforcing that the values were strings then `Record<string, string>` would be better in that case.

Using `Partial<ButtonProps>` instead of `Record<string, unknown>` would potentially be more accurate, but I ran into some type issues with the Button component that felt out-of-scope to fix in this PR.

As always, I'll gladly change it if you don't think it's the right choice, I just wanted to give my reasoning 🙂

> Could we also add an example of how to use this prop to either Storybook or Documentation please. So that we have a reference point for the user.

I somehow missed this when writing my above comment, I'll add an example to the docs shortly

rezrah: I see, thanks for clarifying. Let's keep as-is for now...

For additional context on original question though... I see it as two issues:

- We could implement some type narrowing to remove ambiguity around types and prevent need for `unknown`
- because of how manual the types are in our docs we have an opportunity to be more intentional about which types we want users to pass through. E.g. We can advertise that it accepts either a function or string, even if we don't implement narrowing and the implementation detail technically allows for more. 

The thing I'm highlighting is more of an optics issue, where `unknown` is a bit 'ugly' in documentation. I can totally see why we're going with this though, and I think the only option is to use type narrowing if we want to be 100% accurate.

PR #665, opened 2024-07-24
Upgrade Doctocat

pouretrebelle:
This resolves a content blowout layout bug on nav component pages - https://github.com/primer/doctocat/pull/739

We have to anchor to PRC v36.11.0 because @react-aria/ssr was [removed in v36.12.0](https://github.com/primer/react/pull/4350) but is [still used in doctocat](https://github.com/primer/doctocat/blob/%40primer/gatsby-theme-doctocat%405.1.1/theme/src/components/wrap-root-element.js#L1), which has a PRC dependency of ^v36.1.0, where @react-aria/ssr was still a dep.

## List of notable changes:

-
-

- 
-

## Steps to test:

1.
1.
1.

-
-

## Contributor checklist:

## Reviewer checklist:

## Screenshots:

<table>
<tr>
</tr>
<tr>
<td valign="top">

 </td>
<td valign="top">

</td>
</tr>
</table>


pouretrebelle:
@raytalks can I get your thoughts on this:

PRC in Doctocat [was bumped to v36](https://github.com/primer/doctocat/pull/681) in v5.1.0 which afaik no primer.style sites have updated to. Upgrading to v36 means that the UnderlineNav has been updated ([old](https://primer.style/components/underline-nav/react/deprecated), [new](https://primer.style/components/underline-nav/react/latest)), which changes the look of the Brand and About links in header. Due to the css setup in the header it defaults to a dropdown menu instead (see screenshots in description).

I tried overloading the header to use the old deprecated nav as a hotfix but couldn't get it working.  So the choices here are:
1. Merge this as is, accepting the header will look different between Brand and the rest of primer.style
2. Add a fix to Doctocat which will allow the UnderlineNav to actually render as an underline nav instead of the dropdown. It may still look a shade different to the old version (the deprecated one has slightly different spacing) but I don't think it'd be noticeable. I could also fix the @react-aria/ssr bug at the same time.
3. Close up this pull request and leave the broken layout bug 🤷

N.B. that not doing the second option means that no sites can make Doctocat upgrades without having this issue. Idk if this is a concern for life before nextocat. cc: @dipree

dipree:
Appreciate the eye to detail, no preference from my side, on product we are quite stuck with doing anything, so, focusing on the new version.

apps/docs/package.json
```diff
@@ -11,18 +11,18 @@
     "build:animation": "node scripts/animation-support.js"
   },
   "dependencies": {
-    "@primer/gatsby-theme-doctocat": "^4.6.3",
+    "@primer/gatsby-theme-doctocat": "^5.1.1",
```
rezrah, line 14: Let's pin the version here. We want to do this for all dependencies going forward.

```suggestion
    "@primer/gatsby-theme-doctocat": "5.1.1",
```

apps/docs/package.json
```diff
@@ -11,18 +11,18 @@
...
-    "@primer/react": "^35.1.0",
+    "@primer/react": "36.11.0",
     "color2k": "^2.0.0",
     "cross-env": "7.0.2",
     "gatsby": "^3.7.2",
     "gatsby-plugin-alias-imports": "^1.0.5",
     "gatsby-plugin-postcss": "^4.14.0",
     "gatsby-plugin-typescript": "^2.12.1",
-    "react": "^17.0.2",
+    "react": "^18.x",
```
rezrah, line 23: Can we pin this version too please and avoid wildcards. Guessing it will be `18.2.0` here?

rezrah:
Option 1 seems okay... reasons being:

- It's using the newer, preferred version of the underline nav pattern
- It's already part of the latest Doctocat release, and that decision (along with the differences it introduces) was presumably made with oversight from Primer LT. I'm also assuming that the upgrade contains various other improvements - particularly around security - that we will benefit from.
- It fixes a major layout issue in our docs right now (SubdomainNavBar overflow)
- Users are already on our site, so finding the Brand tab isn't that important (at this part of the funnel)

I would however, be a much more concerned **if product docs moves to this format, as the discoverability of the Brand site would be extremely compromised.** cc. @dipree 

Perhaps also worth noting that Design LT had previously requested that we avoid introducing notable design differences between the product and marketing documentation sites, and given the prominence of this nav at the top of the page, I'm uncertain if this fall under that scope. That said, there are still enough good reasons for Primer Brand docs to move ahead with this change.

PR #671, opened 2024-07-26
Update Storybook to 8.2.6 (latest), and move preview deploys to /brand

joshfarrant:
## Summary

Update Storybook to latest, along with associated dependencies. Also moves preview deploys to live under /brand to match production.

## List of notable changes:

- Packages
  - Updated Storybook (and its addons) from `7.6.17` to `8.2.6` (latest).
  - Updated TypeScript from `4.9.5` to `5.5.4` (latest).
  - Updated jest and ts-jest from `27.5.1` to `29.7.0` (latest).
  - Added jest-environment-jsdom as that's no longer included in jest and must be installed separately.
- Deploy preview builds under `/brand` to more closely match production and make asset handling simpler in preview deploys.
- Modified paths of static assets in docs and storybook to point to their respective local static directories
- Updated `axe-clean.spec.ts` to use new `index.json` format instead of deprecated `stories.json`
- Updated `playwright.generate-tests.ts` to use new `index.json` format instead of deprecated `stories.json`
  - This has caused the visual specs to be sorted in a slightly different order. This shouldn't have any actual effect, but that's the reason for the changes to the `*.visual.spec.ts` files. Because of this, it might be easier to [view those changes in isolation](https://github.com/primer/brand/pull/671/commits/de8971ffd30cea9f6c0afd87681fe045eaeecd8f) when reviewing this PR

- Verify that the docs and storybook work as expected on the preview deploy now that they're being served under /brand
- Check that all our tests pass and the package updates haven't introduced any regressions

## Steps to test:

- [Link to preview deployment](https://primer-bbfb92f725-26139705.drafts.github.io/brand)

-

## Contributor checklist:

## Reviewer checklist:


apps/storybook/.storybook/main.js
```diff
@@ -2,6 +2,7 @@ import {dirname, join} from 'path'
 
 module.exports = {
   stories: ['../../../packages/react/src/**/*.stories.@(js|jsx|ts|tsx)'],
+  staticDirs: ['../static', '../../docs/static'],
```
joshfarrant, line 5: Do I need to load the docs assets too, or can I remove this?

rezrah:
This is looking great @joshfarrant, just doing a sweep of the Storybook and will log my feedback in this comment:

- [ ] [Failing interaction test for ActionMenu](https://primer-bbfb92f725-26139705.drafts.github.io/brand/storybook/?path=/story/components-actionmenu-features--keyboard-navigation&globals=backgrounds.grid:!false;backgrounds.value:transparent)
- [ ] Story parameters being persisted across unrelated stories

  https://github.com/user-attachments/assets/053821dd-db89-4a0c-a724-0edc45816d02

joshfarrant:
@rezrah Good spot, thanks. I've fixed that test

Regarding the story parameters being persisted, it appears to be a bug (and maybe a #wontfix) in Storybook https://github.com/storybookjs/storybook/issues/27073

rezrah:
😩 that's frustrating. It's a pretty severe bug, and IMO breaks the entire Storybook experience as the user will end up viewing certain stories in ways they were never intended to. Did the [workaround](https://github.com/storybookjs/storybook/issues/27073#issuecomment-2225329662) in that Issue pan out?

packages/e2e/scripts/playwright/playwright.generate-tests.ts
```diff
@@ -1,7 +1,9 @@
 ;(function () {
   /* eslint import/no-nodejs-modules: ["error", {"allow": ["path", "fs"]}] */
   // eslint-disable-next-line @typescript-eslint/no-var-requires
-  const {stories} = require('../../../../apps/storybook/storybook-static/stories')
+  const {StoryIndex} = require('@storybook/types')
```
rezrah, line 4: Nice!

packages/react/package.json
```diff
@@ -64,7 +64,9 @@
     "color2k": "^2.0.2",
     "css-loader": "^6.7.4",
     "identity-obj-proxy": "^3.0.0",
+    "jest": "^29.7.0",
```
rezrah, line 67: Is adding this here intentional and/or needed? This is already set in the monorepo root, so this is a duplication. If you do need it, could you hoist it?

packages/react/src/VideoPlayer/VideoPlayer.stories.tsx
```diff
@@ -56,12 +56,14 @@ export default {
...
-} as Meta<typeof VideoPlayer>
+}
+
+export default meta
 
 export const Playground: StoryFn<typeof VideoPlayer> = args => (
   <VideoPlayer {...args}>
-    <VideoPlayer.Source src="https://primer.github.io/brand/assets/example.mp4" type="video/mp4" />
-    <VideoPlayer.Track src="https://primer.github.io/brand/assets/example.vtt" default />
+    <VideoPlayer.Source src="/brand/storybook/example.mp4" type="video/mp4" />
```
rezrah, line 65: This example isn't working in local environments and assets are 404'ing as the URL assumes you're viewing in staging or prod...

Screenshot 2024-07-31 at 17 27 05

Can we roll back any changes that aren't specific to the upgrade to rule out regressions like this? We can patch them up in a follow-up PR.

packages/react/package.json
joshfarrant: Not required sorry, must have slipped through, good spot 👍

packages/react/src/VideoPlayer/VideoPlayer.stories.tsx
joshfarrant: That's rolled back, I'll take a look in a subsequent PR

packages/e2e/package.json
```diff
@@ -41,7 +41,9 @@
   "devDependencies": {
     "@github/axe-github": "^0.5.0",
     "@playwright/test": "^1.34.3",
-    "axe-playwright": "^1.1.12"
+    "axe-playwright": "^1.1.12",
+    "jest": "^29.7.0",
+    "ts-jest": "^29.2.3"
```
rezrah, line 46: Do these two need to be added in `e2e` folder?

packages/react/package.json
```diff
@@ -65,6 +65,7 @@
     "css-loader": "^6.7.4",
     "identity-obj-proxy": "^3.0.0",
     "jest-axe": "^6.0.1",
+    "jest-environment-jsdom": "^29.7.0",
```
rezrah, line 68: @joshfarrant so I know, is this related to storybook upgrade, or a missing dependency in the project? I assumed jsdom environment was already set.

rezrah (approved):
This is looking great 🙌 

@joshfarrant - just two minor questions about dependency changes. Also could you please update the snapshots one more time to be sure there's no additional regressions with the latest changes please, seems like it hasn't run in a while.

packages/react/package.json
joshfarrant: Yes, it's related to the Storybook upgrade.

Storybook 8 requires Jest 28, and in Jest 28 they [stopped bundling `jest-environment-jsdom` as part of Jest](https://jestjs.io/blog/2022/04/25/jest-28#breaking-changes).

> Please note that both of the removed modules (jest-environment-jsdom and jest-jasmine2) are still actively maintained and tested in the same way, so the only breaking change here is that you'll need to explicitly install them.

packages/e2e/package.json
joshfarrant: Good spot, thanks b2af2adce54e885dff03949bc4d076428b2e485e

joshfarrant:
Thanks @rezrah, I've responded to and resolved both comments 

I've been running the snapshot update in docker locally and all's good, but I've re-applied the label again just to be sure 👍

PR #679, opened 2024-08-05
Make SubNav SubMenus accessible using keyboard

joshfarrant:
## List of notable changes:

-
-

- 
-

## Steps to test:

1.
1.
1.

- https://github.com/github/primer/issues/3672

## Contributor checklist:

## Reviewer checklist:

## Screenshots:

<table>
<tr>
</tr>
<tr>
<td valign="top">

 </td>
<td valign="top">

</td>
</tr>
</table>


packages/react/src/SubNav/SubNav.tsx
```diff
@@ -176,56 +177,94 @@ type SubNavLinkProps = {
...
+
+  const onKeyDown = useCallback((e: React.KeyboardEvent<HTMLButtonElement>) => {
+    if (['Enter', ' '].includes(e.key)) {
+      setIsExpanded(prev => !prev)
+    }
+  }, [])
+
+  return (
+    // Disabling as the focus and blur events are handled by the useIsChildFocused hook
+    // eslint-disable-next-line jsx-a11y/mouse-events-have-key-events
```
joshfarrant, line 207: @rezrah Do I remember correctly that we can't disable `jsx-a11y` rules?

It's slightly annoying in this case as it's trying to get me to add `onFocus` and `onBlur` events to the div, but that's not the behaviour we want, and it wouldn't be accessible even if we did as the `div` isn't focusable.

The `useIsChildFocused` hook handles the focus/blur so we've still got all of the functionality that's required, just not in the way that this eslint rule wants.

If we can't disable the rule as I've done here, then I could either include `onBlur` and `onFocus` as noops, eg

```tsx
onFocus={noop}
onBlur={noop}
```

or attach the `onMouseOver` and `onMouseOut` handlers using `addEventListener` inside a `useEffect` as I don't think eslint will pick up on that.

packages/react/src/hooks/useIsChildFocused.ts
```diff
@@ -0,0 +1,49 @@
+import {useCallback, useEffect, useRef, useState, type RefObject} from 'react'
+
+/**
+ * Determine if a child element of the provided ref is currently focussed.
+ * @param onFocusChange The callback to be called when the focus state changes.
+ * @type T The type of the container element.
+ * @returns The ref to be applied to the container element.
+ */
+export const useIsChildFocused = <T extends HTMLElement>(
```
rezrah, line 9: Are you seeing a lot of reuse of this hook @joshfarrant, to warrant it being being hoisted from the source folder of sub nav? 

I think its use-cases are a little limited, so if we're keeping this I'd be in favor of co-locating it to the sub nav folder and keeping it private. We do this for some other components like anchor nav. 

Arguably, you could also eliminate most of the lines here and put the useEffect back into SubNav? I think that's the best option, given there's only one effect and the state management is incidental.

packages/react/src/SubNav/SubNav.tsx
rezrah: AFAIK based on my current understanding, the a11y linter uses static code analysis and there isn't a way to ignore it (which would defeat its purpose). I think you'd either have to go with noop approach like you shared. If you need to set a value, can you just set undefined or null?

packages/react/src/SubNav/SubNav.tsx
```diff
@@ -176,56 +177,94 @@ type SubNavLinkProps = {
...
 }: SubNavLinkProps) => {
-  const hasSubMenu = Children.toArray(children).some(child => {
+  const [isExpanded, setIsExpanded] = useState(false)
+  const onFocusChange = useCallback((isFocused: boolean) => {
+    if (!isFocused) {
+      setIsExpanded(false)
+    }
+  }, [])
+  const submenuId = useId()
+  const ref = useIsChildFocused<HTMLDivElement>(onFocusChange)
```
rezrah, line 195: Related to https://github.com/primer/brand/pull/679/files#r1706597663

This is probably another reason to remove the hook, it doesn't behave like an actual hook where given the required argument needing to be declared first, it can't be called at the beginning of the closure. Surprised that didn't trigger the rules of hook linter warning 🤔

packages/react/src/SubNav/SubNav.tsx
joshfarrant: Cool, I've just set them to `onFocus={() => null}` for now and added a comment explaining the choice

packages/react/src/hooks/useIsChildFocused.ts
joshfarrant: Happy to co-locate. I think I had it alongside SubNav before but thought I'd pull it up as it doesn't do anything specific to SubNav. I'll move it back into the directory 👍

I'd advocate for keeping it as its own hook, purely because it groups all of the logic related to this functionality together into a single place, along with a descriptive name. What lines are you thinking could be removed?

packages/react/src/SubNav/SubNav.tsx
joshfarrant: Could you expand on this a bit please, I'm not sure I understand why this isn't behaving like a hook?

My understanding is that hooks are allowed to consume the results of other hooks (we do it all the time), and they don't need to be called at the beginning of the containing component, they just [need to be called at the top level](https://react.dev/reference/rules/rules-of-hooks#only-call-hooks-at-the-top-level).

Happy to have a call if that's easier 🙂

.changeset/wild-olives-sin.md
```diff
@@ -0,0 +1,5 @@
+---
+'@primer/react-brand': patch
+---
+
+Fixed an issue where SubNav submenus were not accessible through keyboard navigation
```
rezrah, line 5: So the component name stands out a bit more in the final changelog.

```suggestion
Fixed an issue where `SubNav` submenus were not accessible through keyboard navigation
```

rezrah (approved):
Looks great 🙌

PR #688, opened 2024-08-08
Add TextRevealAnimation component 

rezrah:
Part of https://github.com/github/brand-marketing-design/issues/1469

Adds a new, experimental `TextRevealAnimation` component, and applies it the solution templates and article pages.

- 📺 [Video walkthrough](https://github.slack.com/archives/C04E44ZPNEA/p1723211283254439)
- 🔗 [Component preview](https://primer-501d58a8d4-26139705.drafts.github.io/brand/storybook/?path=/story/components-textrevealanimation--playground)

## List of notable changes:

- **added** TextRevealAnimation component, including examples and small playground
- **added** ability to trigger effect on visibility
- **removed** previous, custom implementation from templates
- **added** component back to the Article and Solution pages
- **updated** Article page template to remove ToC into separate component. Reduces size of main template and removes a re-render issue that causes animation effect to trigger multiple times.

## Steps to test:

- [Component preview](https://primer-501d58a8d4-26139705.drafts.github.io/brand/storybook/?path=/story/components-textrevealanimation--playground)
- [Solution templates](https://primer-501d58a8d4-26139705.drafts.github.io/brand/storybook/?path=/story/recipes-solutions-solution-industry--maximum) (scroll to bottom)
- [Article templates](https://primer-501d58a8d4-26139705.drafts.github.io/brand/storybook/?path=/story/recipes-seo-article-page--playground)

## Contributor checklist:

## Reviewer checklist:

## Screenshots:

https://github.com/user-attachments/assets/98b75962-03b7-4591-8982-63602c4dbcea


nsolerieu (commented):
The viewport entry moment at `0:03` in your recording looks very abrupt - could you add a `fade-in` on the both text and wipe overlay?

This would be even more important in context with other animations. If only one start without a smooth fade in it may look jarring or like a UI bug (blinking)

packages/react/src/Testimonial/Testimonial.tsx
```diff
@@ -111,12 +111,9 @@ const Root = forwardRef(_Root)
...
  */
-type QuoteProps = {
-  children: string | Array<string | React.ReactElement<HTMLElement>>
-} & React.HTMLAttributes<HTMLElement> &
-  BaseProps<HTMLElement>
+type QuoteProps = React.HTMLAttributes<HTMLElement> & BaseProps<HTMLElement>
 
-function _Quote({children, className}: QuoteProps, ref) {
+function _Quote({children, className}: PropsWithChildren<QuoteProps>, ref) {
   const childrenArray = useMemo(() => React.Children.toArray(children), [children])
```
joshfarrant, line 117: Minor thing, but you don't need `PropsWithChildren` here as that's included in `HTMLAttributes`. You could also use `HTMLQuoteElement` over `HTMLElement` to be a bit more explicit, although it makes no difference afaik

```suggestion
type QuoteProps = React.HTMLAttributes<HTMLQuoteElement> & BaseProps<HTMLElement>

function _Quote({children, className}: QuoteProps, ref) {
```

packages/react/src/animation/TextRevealAnimation/TextRevealAnimation.visual.spec.ts-snapshots/Visual-Comparison-TextRevealAnimation-TextRevealAnimation-Animation-On-Scroll-1-linux.png
joshfarrant, line 1: Is there a way to improve/remove this snapshot, as it doesn't serve much purpose at the moment

packages/react/src/recipes/SolutionTemplates/SolutionPage/Industry/Industry.visual.spec.ts-snapshots/Visual-Comparison-Industry-Industry-Maximum-Light-1-linux.png
joshfarrant, line 1: Do you think a delay would be useful on this set of snapshots so we capture them once the animation has finished?

packages/react/src/recipes/SolutionTemplates/SolutionPage/UseCase/UseCase.visual.spec.ts-snapshots/Visual-Comparison-UseCase-UseCase-Maximum-Dark-1-linux.png
joshfarrant, line 1: Same comment as before regarding delaying these snapshots

packages/react/src/animation/TextRevealAnimation/useTextRevealAnimation.ts
```diff
@@ -1,7 +1,7 @@
 import {useRef, useEffect, useState} from 'react'
 
-export const useWrapLines = (text?: string) => {
-  const ref = useRef<HTMLHeadingElement>(null)
+export const useTextRevealAnimation = (text?: string) => {
```
joshfarrant, line 3: Minor thing, but I think the previous naming was more accurate here. This hook doesn't deal with any animation or revealing, it just allows you to wrap individual lines of text as you see fit. Maybe `useSplitByLine` or something similar could be another option.

packages/react/src/animation/TextRevealAnimation/useTextRevealAnimation.ts
```diff
@@ -1,7 +1,7 @@
...
+  const ref = useRef<HTMLElement>(null)
```
joshfarrant, line 4: Good spot 🙌

packages/react/src/animation/TextRevealAnimation/TextRevealAnimation.tsx
```diff
@@ -0,0 +1,72 @@
...
+
+  const onLineAnimationEnd = useCallback(
+    (i: number) => {
+      if (lines && i === lines.length - 1) {
+        setHasHeroWipeAnimated(true)
+      }
+    },
+    [lines],
+  )
+
```
joshfarrant, line 20: I think this might be a case where `useCallback` is going to be less-performant than not using it. 

`onLineAnimationEnd` is only passed to a span, and so no other component renders are tied to the identity of this function, meaning there's no benefit in keeping the identity of the function the same across renders. All we're doing is adding the runtime overhead of `useCallback`, along with the memory footprint of memoizing the function. Both of those are admittedly tiny, but mentioning as all we're doing here is adding code that doesn't give any benefit.

I think I wrote this originally, so this is a comment to my past self more than anything 😅

packages/react/src/animation/TextRevealAnimation/TextRevealAnimation.tsx
```diff
@@ -0,0 +1,72 @@
+import React, {useCallback, useEffect} from 'react'
+import clsx from 'clsx'
+import {useTextRevealAnimation} from './useTextRevealAnimation'
+
+import styles from './TextRevealAnimation.module.css'
+
+export function TextRevealAnimation({children, ...rest}) {
+  const [animationStarted, setAnimationStarted] = React.useState(false)
```
joshfarrant, line 8: I'm surprised that TS isn't complaining here, but we're missing a type on the props

```suggestion
type TextRevealAnimationProps = React.HTMLAttributes<HTMLSpanElement> & {children: string}
```

We'll either have to force `children` to be a string here, or accept the normal children and have the hook check that it's a string and nothing else

packages/react/src/animation/TextRevealAnimation/TextRevealAnimation.examples.stories.tsx
```diff
@@ -0,0 +1,44 @@
...
+  decorators: [
+    Story => (
+      <AnimationProvider>
+        <Box animate="fade-in">
+          <Story />
+        </Box>
+      </AnimationProvider>
+    ),
+  ],
+} as Meta<typeof TextRevealAnimation>
```
joshfarrant, line 18: Something to discuss in general, but we should probably discourage the use of `as` in the codebase.

In this example, you can put whatever you want into the above object and TS won't complain as we're overriding it with `as`.

If we switched to 

```ts
export const meta: Meta<typeof TextRevealAnimation> = {
    // ...
}
export default meta
```

Then TS will complain if we pass invalid keys

joshfarrant (commented):
Very happy to see this split out 🙌

packages/react/src/Testimonial/Testimonial.tsx
rezrah: Nice! Didn't know that extended DOMAttributes. Was mostly trying to get rid of the manual children declaration.

packages/react/src/animation/TextRevealAnimation/TextRevealAnimation.tsx
rezrah: Thanks. I've added a type declaration for TS users and a warning for JS., which will cover most use-cases. We don't use inline types, so exporting a dedicated declaration for it instead.

packages/react/src/animation/TextRevealAnimation/TextRevealAnimation.examples.stories.tsx
joshfarrant: Coincidentally I've just run into an example of this with `LogoSuite`.

We're using a type assertion (`as Meta`) in the stories there too and that hides the fact that `default` isn't a valid `marqueeSpeed`. The correct value should have been `normal`.

If I switch it to a type a type annotation (`: Meta`) then that error is correctly highlighted.

https://github.com/primer/brand/pull/691/commits/0d3b6b90c662560a77ce3bcbb1e7b3a841675692

rezrah:
> The viewport entry moment at `0:03` in your recording looks very abrupt - could you add a `fade-in` on the both text and wipe overlay?

@nsolerieu thanks for tip, added a fade-in which smoothens the effects appearance...

How's it look now?

https://github.com/user-attachments/assets/e654b469-aefc-4101-9985-7520dada8cb2

packages/react/src/animation/TextRevealAnimation/useTextRevealAnimation.ts
rezrah: Yeah that's fair. I called it this for DevExp reasons, so we can find it easily while doing file searches. It can also be used as a general hook as we expand capabilities of the component, which is what we do for some other components too. I can see it the other way too though, as it does return some specific information.

packages/react/src/animation/TextRevealAnimation/TextRevealAnimation.examples.stories.tsx
rezrah: I agree with this in principle. One edge case we have is where some stories need to extend the components props with story specific ones. This approach wouldn't directly work there, but we could create an intersection instead? How are you seeing that?

packages/react/src/animation/TextRevealAnimation/TextRevealAnimation.visual.spec.ts-snapshots/Visual-Comparison-TextRevealAnimation-TextRevealAnimation-Animation-On-Scroll-1-linux.png
rezrah: I'd say this is actually correct, it shows that the text/component is not visible on load.

packages/react/src/recipes/SolutionTemplates/SolutionPage/Industry/Industry.visual.spec.ts-snapshots/Visual-Comparison-Industry-Industry-Maximum-Light-1-linux.png
rezrah: A delay on its own wouldn't help here, because the visibility of the testimonial requires user interaction / scrolling. I'd treat this as correct for now, as it should be hidden on load. A potential improvement would probably be instructing headless browser to scroll to bottom of the page before taking a snapshot, but I can look into that separately..

packages/react/src/animation/TextRevealAnimation/TextRevealAnimation.examples.stories.tsx
joshfarrant: Yea that's a good point. I tend to prefer creating a new `ControlsProps` type that works pretty much as you'd expect.

I've got an example of that [here](https://github.com/primer/brand/blob/85bff7aa3d201e2caf5e2cca3c09c4a820e32e71/packages/react/src/LogoSuite/LogoSuite.stories.tsx)

packages/react/src/animation/TextRevealAnimation/useTextRevealAnimation.ts
joshfarrant: Cool, I can see the logic there.

My opinion would be that, since it's already co-located with the component (which tells us that it's related to the component), using a generic name instead of a descriptive one would make for a worse DevExp, as devs now have to read the implementation of the hook before they can understand what it does.

It also makes it more likely that additional, unrelated stuff will get piled into that hook in the future too. I take the point that it's convenient to have a single hook that contains all of a component's hook logic, but in the same way that we break big functions or components down into smaller, more focussed parts I'd tend towards doing the same with hooks. If a component ends up relying on a large number of those smaller hooks then that's the point I'd look to wrap up all the calls to those hooks into a single, component-named hook.

I'm happy with however you want to approach it though, just sharing my two cents 🙂

packages/react/src/recipes/SolutionTemplates/SolutionPage/Industry/Industry.visual.spec.ts-snapshots/Visual-Comparison-Industry-Industry-Maximum-Light-1-linux.png
joshfarrant: Ahh of course, gotcha.

Yea that could be an interesting follow-up at some point, but agreed that it's not something that needs addressing now

packages/react/src/animation/TextRevealAnimation/TextRevealAnimation.examples.stories.tsx
joshfarrant: You could you that same pattern I mentioned above [here](https://github.com/primer/brand/pull/688/files#diff-5a817314c9e27f624f129949bef8dd3a335fa52920ca35fe94c2b05a47e6799bR38) to remove the `as`, for example.

packages/react/src/animation/TextRevealAnimation/TextRevealAnimation.tsx
rezrah: 👍 yeah, fair points.. have removed the memoization. I think a similar issue is the recreation of the inline, anonymous function on each render, which would have been the one we wanted to memoize. Arguably negligible impact though, so will leave as-is.

joshfarrant: Great stuff, thanks 🙌

It's a bit counterintuitive, but the anonymous function will be recreated on every render regardless of whether we use `useCallback`.

Kent C Dodds' article breaks it down quite well ([link to relevant section](https://kentcdodds.com/blog/usememo-and-usecallback#why-is-usecallback-worse))

packages/react/src/animation/TextRevealAnimation/useTextRevealAnimation.ts
rezrah: 👍 yeah I agree with those points. I've already renamed the hook to append Lines at the end as it was returning something rather than a void... Still has the component name prepended so best of both worlds.

packages/react/src/animation/TextRevealAnimation/TextRevealAnimation.examples.stories.tsx
rezrah: 🤔 I see, you're using intersections but manually cherry-picking specific props instead of automatically inheriting and joining them. Is that what's happening? (I could totally be reading this wrong). I can see why we'd need to do that for compound APIs, but it feels a bit duplicative.

If we needed to create story props though, creating a new type makes sense. Could you call it StoryProps though please, we [have something similar for IDE already](https://github.com/primer/brand/blob/main/packages/react/src/IDE/IDE.stories.tsx#L75).

joshfarrant: I'm creating my own type by `Pick`ing the props I care about from the individual LogoSuite compound components here

```ts
type ControlsProps = Pick<LogoSuiteProps, 'align' | 'hasDivider'> &
  Pick<LogoSuiteHeadingProps, 'visuallyHidden'> &
  Pick<LogoSuiteLogoBarProps, 'marquee' | 'marqueeSpeed' | 'variant'>
```

Which is equivalent to this, just minus the duplication

```ts
type ControlsProps = {
  hasDivider?: boolean,
  align?: 'start' | 'center' | 'justify',
  visuallyHidden?: boolean,
  marquee?: boolean,
  marqueeSpeed?: 'slow' | 'normal' | 'idle',
  variant?: 'muted' | 'emphasis'
} 
```

I then grab those props and put them in the right place in the Playground story

```tsx
export const Playground: StoryFn<ControlsProps> = ({
  hasDivider,
  align,
  visuallyHidden,
  marquee,
  marqueeSpeed,
  variant,
}) => (
  <LogoSuite align={align} hasDivider={hasDivider}>
    <LogoSuite.Heading visuallyHidden={visuallyHidden}>Heading</LogoSuite.Heading>
    <LogoSuite.Description>Body text maximus ligula felis, non egestas dolor rutrum vel.</LogoSuite.Description>
    <LogoSuite.Logobar variant={variant} marquee={marquee} marqueeSpeed={marqueeSpeed}>
      {logos.slice(0, 5)}
    </LogoSuite.Logobar>
  </LogoSuite>
)
```

joshfarrant (approved):

packages/react/src/animation/TextRevealAnimation/TextRevealAnimation.tsx
```diff
@@ -0,0 +1,81 @@
...
+import {BaseProps} from '../../component-helpers'
+
+export type TextRevealAnimationProps = BaseProps<HTMLSpanElement> &
+  Omit<React.HTMLAttributes<HTMLSpanElement>, 'children'> & {
+    children: React.ReactChild
+  }
+
+export function TextRevealAnimation({children, ...rest}: TextRevealAnimationProps) {
+  const [animationStarted, setAnimationStarted] = React.useState(false)
+  const [hasHeroWipeAnimated, setHasHeroWipeAnimated] = React.useState(false)
```
danielguillan, line 15: Should we rename these to be component-agnostic? For example, `hasAnimated` and `setHasAnimated`.

danielguillan (approved):

packages/react/src/animation/TextRevealAnimation/TextRevealAnimation.tsx
rezrah: 🤦 good spot, i forgot to rename this after porting it from article template... will fix 👍

PR #616, opened 2024-06-18
Updated default river description color

rezrah:
## Summary

Closes https://github.com/github/primer/issues/3341

Updates text descriptions in River to use muted instead of default. 

## List of notable changes:

- **added** reminder about contrast to changeset
- **updated** default color value in River

## Steps to test:

1. Review snapshots and storybook

## Contributor checklist:

- [ ] All new and existing CI checks pass
- [ ] Tests prove that the feature works and covers both happy and unhappy paths
- [ ] Any drop in coverage, breaking changes or regressions have been documented above
- [ ] New visual snapshots have been generated / updated for any UI changes
- [ ] All developer debugging and non-functional logging has been removed
- [ ] Related issues have been referenced in the PR description

## Reviewer checklist:

- [ ] Check that pull request and proposed changes adhere to our [contribution guidelines](../../CONTRIBUTING.md) and [code of conduct](../../CODE_OF_CONDUCT.md)
- [ ] Check that tests prove the feature works and covers both happy and unhappy paths
- [ ] Check that there aren't other open Pull Requests for the same update/change

## Screenshots:

> Please try to provide before and after screenshots or videos

<table>
<tr>
<th> Before</th> <th> After </th>
</tr>
<tr>
<td valign="top">

image

 </td>
<td valign="top">

image

</td>
</tr>
</table>

<table>
<tr>
<th> Before</th> <th> After </th>
</tr>
<tr>
<td valign="top">

Screenshot 2024-06-18 at 14 45 52

 </td>
<td valign="top">

Screenshot 2024-06-18 at 14 46 05

</td>
</tr>
</table>

PR #691, opened 2024-08-12
Apply opinionated defaults to LogoSuite

joshfarrant:
## List of notable changes:

- Added new `default` variant to LogoSuite which:
  - Applies `emphasis` styles if there are 5 or fewer logos
  - Applied `muted` styles if there are 6 or more logos
- Reduced logo size on mobile
- Removed unnecessary (and technically unsupported) Fragments from LogoSuite Stories
- Tidied up LogoSuite stories
- Don't actively encourage stacking LogoSuite components in docs or Storybook
- Prefer examples with visually-hidden headings

- Make sure you're happy with the regenerated snapshots
- Does the functionality of the new `default` variant make sense and is it communicated appropriately in the docs?

- https://github.com/github/primer/issues/3290

## Contributor checklist:

## Reviewer checklist:

## Screenshots:

<table>
<tr>
</tr>
<tr>
<td valign="top">

image

 </td>
<td valign="top">

image

</td>
</tr>
</table>


packages/design-tokens/src/tokens/functional/components/logosuite/base.json
```diff
@@ -6,7 +6,12 @@
           "value": "var(--base-size-80)"
         },
         "rowGap": {
-          "value": "var(--base-size-40)"
+          "mobile": {
+            "value": "var(--base-size-40)"
+          },
+          "default": {
+            "value": "var(--base-size-80)"
+          }
```
rezrah, line 14: FYI, you don't need to create separate tokens for different viewports, as we have a custom token formatter that handles responsive behaviour on the same variable. [Example from river](https://github.com/primer/brand/blob/main/packages/design-tokens/src/tokens/functional/components/river/river.json#L7).

packages/react/src/LogoSuite/LogoSuite.visual.spec.ts-snapshots/Visual-Comparison-LogoSuite-LogoSuite-Visible-Heading-With-Description-1-linux.png
rezrah, line 1: The bottom border seems too close to the logos here

Screenshot 2024-08-15 at 17 19 36

packages/design-tokens/src/tokens/functional/components/logosuite/base.json
joshfarrant: Ah nice, thank you.

Is there something I need to do to enable it for a component?

I've updated my base.json to the following but I'm still not getting the responsive behaviour in the outputted `/lib/design-tokens/css/tokens/functional/components/logosuite/base.css` file

```json
{
  "brand": {
    "LogoSuite": {
      "logobar": {
        "columnGap": {
          "value": "var(--base-size-80)"
        },
        "rowGap": {
          "value": "var(--base-size-40)",
          "responsive": {
            "768px": {
              "value": "var(--base-size-80)"
            }
          }
        },
        "marquee": {
          "default": {
            "value": "30s"
          },
          "slow": {
            "value": "60s"
          },
          "gap": {
            "value": "var(--base-size-96)"
          }
        }
      }
    }
  }
}
```

packages/react/src/LogoSuite/LogoSuite.visual.spec.ts-snapshots/Visual-Comparison-LogoSuite-LogoSuite-Visible-Heading-With-Description-1-linux.png
joshfarrant: Hmm, I'm not entirely sure what was going on here. I wasn't able to reproduce this locally, and regenerating the snapshots seems to have corrected it ¯\\\_(ツ)_/¯

joshfarrant: Hmm, scratch that. It looks like it's like that in the deployed version, just not locally. That would explain the snapshot discrepancy too.

Let me look into it. I might nuke my local env too just to be safe

packages/design-tokens/src/tokens/functional/components/logosuite/base.json
danielguillan: Yes, you have to add it to the [`build-tokens` script file](https://github.com/primer/brand/blob/main/packages/design-tokens/scripts/build-tokens.js#L161-L171).

joshfarrant: Amazing, thanks @danielguillan 💖

danielguillan (approved):

packages/react/src/LogoSuite/LogoSuite.visual.spec.ts-snapshots/Visual-Comparison-LogoSuite-LogoSuite-Visible-Heading-With-Description-1-linux.png
joshfarrant: @rezrah I'm not sure what the cause of the discrepancy was but it seems to be happy now

packages/react/src/LogoSuite/LogoSuite.tsx
```diff
@@ -157,16 +157,26 @@ type LogoSuiteLogoBarProps = BaseProps<HTMLDivElement> & {
   /**
    * The stylistic variant of the LogoBar.
    */
-  variant?: 'muted' | 'emphasis'
+  variant?: 'muted' | 'emphasis' | 'default'
```
rezrah, line 160: Do we need 'default'? Given the color is dynamic and we only seem to alternate between muted and emphasis colors, can we drop this?

.changeset/silent-scissors-suffer.md
```diff
@@ -0,0 +1,10 @@
+---
+'@primer/brand-primitives': minor
+'@primer/react-brand': minor
+---
+
+⚠️ This update contains a breaking change.
+
+- Added new `default` variant to `LogoSuite` component which automatically applies `emphasis` or `muted` styles depending on the number of logos in the suite. 5 or fewer uses `emphasis`, 6 or more uses `muted`.
+- Reduced `LogoSuite` size on mobile viewports.
+- Removed `--brand-LogoSuite-logobar-rowGap` token and replaced it with two new tokens. `--brand-LogoSuite-logobar-rowGap-mobile` is used on mobile viewports, and `--brand-LogoSuite-logobar-rowGap-default` is used on all other viewports.
```
rezrah, line 10: This is no longer relevant?

```suggestion
```

packages/react/src/LogoSuite/LogoSuite.tsx
joshfarrant: On reflection potentially not, it doesn't do anything different to just leaving `variant` as `undefined`, so in hindsight it doesn't offer a lot of benefit.

apps/docs/content/components/LogoSuite/index.mdx
```diff
@@ -22,7 +21,7 @@ Use the logo suite component to showcase a list of logos, such as for clients, s
 
 A logo suite is intended to build credibility, trust, and provide social proof by displaying consistent logos.
 
-Logo suite should be used for groups of logos more than 4 and less than 12. In scenarios hoping to showcase more than 12 logos, considering stacking two components together. To create the best visual consistency and balance, rearrange and resize the logos when necessary to create similar area distribution and attention to each of the logos.
+Logo suite should be used for groups of logos more than 4 and less than 12. To create the best visual consistency and balance, rearrange and resize the logos when necessary to create similar area distribution and attention to each of the logos.
 
 ### Arrangement
```
rezrah, line 27: Unable to add these comments as these sections haven't changed, but will need to...

1. Can we remove section for `Contextual information` as it still recommends adding a heading a description
  Screenshot 2024-08-16 at 15 15 28
2. Remove 'default muted', unless we are dropping new `default` variant and sticking with muted..
  Screenshot 2024-08-16 at 15 30 27
3. Could we please update this example to use the description instead of Heading now. This will help us a) match the marquee example further above it, but also reinforce that we shouldn't use LogoSuite.Heading arbitrarily
  Screenshot 2024-08-16 at 15 32 26

rezrah: 4. In the `React` tab, the "Default" example still includes a heading and description but these are now optional. Can we remove those in the "Default" and move this example to appear immediately  after and call it "Optional heading and description". This should make it super obvious what the minimum expectations are, and match [what we do in other components](https://primer-90ea61f842-26139705.drafts.github.io/brand/components/BreakoutBanner/react#optional-description).

Screenshot 2024-08-16 at 15 35 26

apps/docs/content/components/LogoSuite/react.mdx
```diff
@@ -411,6 +329,88 @@ Logobars using the `emphasis` variant will appear with relatively higher contras
     </svg>
   </LogoSuite.Logobar>
 </LogoSuite>
+<br />
+<LogoSuite>
+  <LogoSuite.Heading visuallyHidden>Muted</LogoSuite.Heading>
```
rezrah, line 334: To avoid it looking like we're stacking logobars again, can we make this look more like the examples above and use the description to explain what's happening?

E.g.

```suggestion
  <LogoSuite.Heading visuallyHidden>Muted example</LogoSuite.Heading>
  <LogoSuite.Description>Muted:</LogoSuite.Description>
```

rezrah: Same would also need to happen for Emphasis example

packages/react/src/LogoSuite/LogoSuite.tsx
rezrah: 👍 cool, after seeing the related code, I think it behaves more like an `auto` variant, which better describes what it does.. but also agree it's probably better to just remove it altogether. Just for future reference, this is a good candidate for an quick API exploration (assuming it hasn't already happened)

apps/docs/content/components/LogoSuite/index.mdx
joshfarrant: All updated (along with a small copy tweak too) in b6616fc962b20c019bfb709a89b452f1f8f4ddd9

For further context, Dani and I agreed that the docs images will be updated in a subsequent PR once he's updated the LogoSuite component in Figma

joshfarrant: In that last suggestion (4) I've updated it but left the heading in the Default example and just made it `visuallyHidden` if that works for you?

packages/react/src/LogoSuite/LogoSuite.tsx
joshfarrant: Yea that's worth keeping in mind in the future, will do, thanks 👍

.changeset/silent-scissors-suffer.md
```diff
@@ -0,0 +1,9 @@
+---
+'@primer/brand-primitives': minor
+'@primer/react-brand': minor
+---
+
+⚠️ This update contains a breaking change.
```
rezrah, line 6: Is the breaking change still valid following recent changes? If it is, could we be more specific about whether it's a visual one vs API? Guessing it's the former, in which case we should give the user a nudge about what to look for. 

Example you can reference from another recent change: https://github.com/primer/brand/pull/616/files#diff-bf4a5c10ff09f2964e747a5873b4c1b2124a813202d314f69effbc3de1faf3dfR7-R8

.changeset/silent-scissors-suffer.md
```diff
@@ -0,0 +1,9 @@
...
+
+- Added new default behaviour to the `variant` prop of the `LogoSuite` component. If `variant` is not provided then either `emphasis` or `muted` styles are automatically applied depending on the number of logos in the suite. 5 or fewer uses `emphasis`, 6 or more uses `muted`.
```
rezrah, line 8: Minor stylistic/grammatical change around using numerals, particularly at the start of a sentence.

```suggestion
- Added new default behaviour to the `variant` prop of the `LogoSuite` component. If `variant` is `undefined` then either `emphasis` or `muted` styles are automatically applied depending on the number of logos in the `LogoBar`. Five or fewer logos apply the `emphasis` style, while six or more use the `muted` style.

```

packages/react/src/LogoSuite/LogoSuite.stories.tsx
```diff
@@ -1,16 +1,32 @@
...
-import {LogoSuite} from './LogoSuite'
+import type {Meta, StoryFn} from '@storybook/react'
+import {LogoSuite, type LogoSuiteHeadingProps, type LogoSuiteLogoBarProps, type LogoSuiteProps} from './LogoSuite'
+import {logos} from './LogoSuite.fixtures'
 
-export default {
+type StoryProps = Required<
+  Pick<LogoSuiteProps, 'align' | 'hasDivider'> &
+    Pick<LogoSuiteHeadingProps, 'visuallyHidden'> &
+    Pick<LogoSuiteLogoBarProps, 'marquee' | 'marqueeSpeed' | 'variant'>
```
rezrah, line 10: This is cool, better than what's currently here so let's go with it 👍 

Leaving some additional thoughts for posterity too, so we can come back to use of Omit vs Pick.

I could be reading this wrong, but when adding a new prop to LogoSuite we unfortunately won't get it for 'free' here unless we opt in. I think we're doing this because we want to avoid a conflict in names between named children, but that's potentially an edge case. I'd assume in most cases we'd want to benefit from auto opt-in where possible. We could instead manually opt-out those conflicting props as they are really only solving the need to flatten Storybook arguments and nothing to do with the component itself.

packages/react/src/LogoSuite/LogoSuite.tsx
```diff
@@ -162,11 +162,14 @@ type LogoSuiteLogoBarProps = BaseProps<HTMLDivElement> & {
...
   (
-    {className, children, marquee = false, marqueeSpeed = 'normal', variant = 'muted', ...props}: LogoSuiteLogoBarProps,
+    {className, children, marquee = false, marqueeSpeed = 'normal', variant, ...props}: LogoSuiteLogoBarProps,
     ref: Ref<HTMLDivElement>,
   ) => {
     const [prefersReducedMotion, setPrefersReducedMotion] = useState(false)
 
+    const childrenCount = React.Children.toArray(children).length
+    variant ??= childrenCount <= 5 ? 'emphasis' : 'muted'
+
```
rezrah, line 172: As this is the main logic being added, could we add a test case for this please? Ideally in this PR, or in a follow up would be great 🙏.

rezrah (approved):
Thanks @joshfarrant, appreciate you going through all that feedback. LogoSuite is looking really cool 👍

packages/react/src/LogoSuite/LogoSuite.tsx
joshfarrant: Added, great suggestion thanks!

packages/react/src/LogoSuite/LogoSuite.stories.tsx
joshfarrant: Yes that's exactly right. I thought this behaviour would be more predictable and intuitive.

> _You want to add a new control to the story? Just add the type to the top of the file too._

As opposed to

> _You want to add a new prop to one of the LogoSuite components? It might cause conflicts in the Storybook types, regardless of whether this prop is actually used in Storybook._

It's just a matter of preference, but I prefer the explicit, opt-in approach to the more magic, opt-out approach. Adding a type to cover new functionality feels much more natural to me than having types be automatically added for me and then potentially having to manage conflicts. If we adopt this approach everywhere then conflicts might not be that uncommon as we reuse prop names fairly regularly across components.

.changeset/silent-scissors-suffer.md
joshfarrant: Added, thanks!

.changeset/silent-scissors-suffer.md
joshfarrant: I've updated the changeset, let me know what you think 👍

PR #695, opened 2024-08-14
Add visually-hidden text to headers of ComparisonTable feature columns

joshfarrant:
## List of notable changes:

- Featured ComparisonTable columns now feature the visually-hidden text `(featured)` after the column name 

- Check that this is read correctly by screen readers

- https://github.com/github/primer/issues/3752

## Contributor checklist:

## Reviewer checklist:


packages/react/src/ComparisonTable/ComparisonTable.tsx
```diff
@@ -72,6 +73,7 @@ export const _ComparisonTable = forwardRef(
                       )}
                     >
                       {rowChild.props.children}
+                      {isFeatured && <span className="visually-hidden">(featured)</span>}
```
rezrah, line 76: qq. is there a way we can make this customizable by end users? This is currently a hard coded English string, which we'll need to eventually change for localization reasons, but there's also an edge case where content may need to change to better describe the table cell header more accurately (feature / highlighted / spotlighted). I think the the flexibility would be nice to offer on the API.. Not something we need to solve now if it balloons the scope though.

rezrah (approved):
Works as expected to resolve the Issue.

Highlighted one item that we'll need to come back to at a later date, but we can defer to a follow up. 

Thanks for fixing this @joshfarrant

joshfarrant:
@rezrah Great shout, we could just add a `featuredColumnAccessibleLabel` required prop that defaults to `featured` which we always render inside a pair of brackets. wdyt?

joshfarrant:
I've added that @rezrah, let me know your thoughts

rezrah:
> @rezrah Great shout, we could just add a `featuredColumnAccessibleLabel` required prop that defaults to `(featured)`. wdyt?

Yep sounds good, your name suggestion makes sense too, though wondering if we can shorten it, or make it more consistent with how we represent visually hidden content. 

Some other suggestions:

- `featuredLabel`
- `featuredColumnLabel`
- `visuallyHiddenFeaturedLabel` similar to [this](https://github.com/primer/brand/blob/main/packages/react/src/VideoPlayer/VideoPlayer.tsx#L31)

packages/react/src/ComparisonTable/ComparisonTable.tsx
```diff
@@ -29,6 +29,7 @@ export const _ComparisonTable = forwardRef(
       className,
       heading,
       featuredColumn = 1,
+      featuredColumnAccessibleLabel = '(featured)',
```
rezrah, line 32: Could we apply the parenthesis internally please, and not leave to end user to customize in case they forget?

joshfarrant: 👍 794484bd7f9779e9bda0d2697580567c9ba4eeef

apps/docs/content/components/ComparisonTable.mdx
```diff
@@ -124,6 +124,10 @@ For granular control over heading markup or appearance, we recommend using [`Hea
 
 By default, the first data column will be marked as `featured`. This is to draw attention to a particular product as the basis of a comparison against adjacent columns.
 
+Featured columns will have the visually-hidden text `featured` appended to their column title. This text is used to identify the featured column to screen reader users and can be customized using the `visuallyHiddenFeaturedLabel` prop.
+
+```jsx live
```
rezrah, line 129: FYI This isn't closed properly so is breaking docs. Guessing it's not needed?

```suggestion
```

Screenshot 2024-08-14 at 15 21 14

apps/docs/content/components/ComparisonTable.mdx
```diff
@@ -356,12 +360,13 @@ This variant is suitable for embedding into long-form content, or in situations
...
+| `heading`                     | `string`                                                                                | undefined   | `false`  | Optional heading that appears above table, preconfigured as a `h3`.                                                                                       |
+| `featuredColumn`              | `number`                                                                                | `1`         | `false`  | Indicates that the data column at the specified index is visually important and unique in relation to its adjacent cells, and will be styled accordingly. |
+| `visuallyHiddenFeaturedLabel` | `string`                                                                                | `featured`  | `false`  | A visually-hidden label which is appended to the end of the column title. Used to identify the featured column to screen reader users.                    |
```
rezrah, line 368:
```suggestion
| `visuallyHiddenFeaturedLabel` | `string`                                                                                | `'featured'`  | `false`  | A visually-hidden label which is appended to the end of the column title. Used to identify the featured column to screen reader users.                    |
```

rezrah (approved):
Looks great, just two more changes before it's merged (one is super important)

apps/docs/content/components/ComparisonTable.mdx
joshfarrant: Ah sorry, Copilot has been really eager to add this every opportunity it gets. I keep deleting them but obviously I missed it this time 🤦‍♂️

apps/docs/content/components/ComparisonTable.mdx
joshfarrant: Good spot

PR #700, opened 2024-08-16
ActionMenu text reflow fix

rezrah:
## Summary

Resolves https://github.com/github/primer/issues/3758 (SEV-1 ⚠️ )

Removes truncation of long text in `ActionMenu` button and lists, allowing it to instead reflow [per WCAG guidelines](https://www.w3.org/TR/WCAG22/#reflow).

## List of notable changes:

- Removed fixed height, text truncation and ellipsis on buttons to allow text to flow onto multiple lines if needed
- Fixed issues with inherited paddings, which weren't previously accurate
- Removed truncation and ellipsis on popup list items
- Widths on action list items are now dynamically calculated

## Steps to test:

1. [Use this visual diff](https://github.com/primer/brand/compare/rezrah/action-menu-a11y-reflow?expand=1#diff-9aece6da8765c3bade01d153faea0fae7d95ea602525d171f3bf1baae0c35266) to see the before/after of this change. The story name will change after reviews are in, because the diff won't work if that's done first.

- https://github.com/github/primer/issues/3758

## Contributor checklist:

## Reviewer checklist:

## Screenshots:

<table>
<tr>
</tr>
<tr>
<td valign="top">

Screenshot 2024-08-16 at 17 35 16

 </td>
<td valign="top">

Screenshot 2024-08-16 at 17 35 00

</td>
</tr>
</table>


joshfarrant (approved):
LGTM!

There's the one remaining TODO, but I'm guessing that's just there to make the before/after visual diff possible and will be fixed before merge? 🙂

.changeset/quick-poems-know.md
```diff
@@ -0,0 +1,5 @@
+---
+'@primer/react-brand': patch
+---
+
+`ActionMenu` button no longer truncates longer labels. Now follows WCAG criterion around text reflow.
```
danielguillan, line 5: This change may introduce a visual breaking change in some situations. Is this something we should warn users about in the changeset?

danielguillan (approved):

simmonsjenna (commented):
@rezrah Can we have the text within `Select` be left-aligned instead of center-aligned?

.changeset/quick-poems-know.md
rezrah: Thanks for checking in. In this case I would say no. 

1. Most of component changes affect the visuals in some way, so question is what's the threshold to issue a warning
1. We risk diluting the importance and impact of these warnings if we add them too frequently, so they should be used sparingly
1. I recommend adding them when they either pose an immediate risk of accessibility regressions  downstream (e.g poor contrast) **or** when design at-large will deviate considerably and we want folks to not blindly upgrade without checking.

I wouldn't see this change falling under either of those categories ☝️. The only potential risk _could_ have been the height changing, which leads to misalignment with adjacent controls but ActionMenu is not used adjacent to other form controls so we're good there. Hope that makes sense! Let me know if you have any counter considerations though @danielguillan.

rezrah:
> @rezrah Can we have the text within `Select` be left-aligned instead of center-aligned?

@simmonsjenna 👋 Are you referring to the text inside the button or the popup menu?

If button, this is how it would look:

Screenshot 2024-08-20 at 09 49 59

If menu items, those are already left aligned and have are offset by default to make room for the selection arrow. This helps mitigate layout shift.

simmonsjenna:
> > @rezrah Can we have the text within `Select` be left-aligned instead of center-aligned?
> 
> @simmonsjenna 👋 Are you referring to the text inside the button or the popup menu?
> 
> If button, this is how it would look:
> 
> Screenshot 2024-08-20 at 09 49 59

Yes! Inside the button itself as in the example you provided above.

danielguillan:
@simmonsjenna I understand your proposal to left-align the label but I believe keeping it centered is a more consistent option for a couple of reasons:
1. Custom widths: if users set a custom width for the button, a left-aligned label might look awkward, especially if the label is much shorter than the button.
2. Right-aligned action menu: when the control is right-aligned in a UI, a left-aligned label could apper misaligned and visually unbalanced.

I think we cannot easily test those two scenarios and switch from left to center alignment accordingly so the best option might be to keep the default centered alignment.

simmonsjenna:
@danielguillan Thanks for the callout — I dug into it a bit more to understand how things are currently structured.

I didn't realize that the downward chevron was right next to the Button copy:
Screenshot 2024-08-21 at 6 46 42 AM

I was expecting it to be more like this with `justify-content: space-between` applied:
Screenshot 2024-08-21 at 6 46 30 AM

Screenshot 2024-08-21 at 6 50 36 AM


We can keep it as center aligned based on the variations that could occur 👍

PR #704, opened 2024-08-20
Fix static assets dir in preview/production Storybook and docs

joshfarrant:
## Summary

There are currently some issues with the static directories in both the preview/production Storybook and the docs which result in assets not loading correctly in certain environments. The only component which consumes these assets is `VideoPlayer`, so the issue hasn't been particularly noticeable up to now.

This PR resolves those issues and allows us to pull static assets from the asset directory of the current application, without having to pull all assets from the production site, as we were doing before.

> ⚠️ NB This PR uses the `cp` method on Node's `fs` API — in development only — to move the docs assets into the expected directory.  The `cp` method was added in Node 20, so this PR won't work on older versions of Node. If that's an issue, we can use a different method to move the assets, but it will be slightly more complex.
> 
> Node 20 is now LTS however, so I think it's reasonable for us to require this version for development. If you disagree, let me know and I can look into other options.

## List of notable changes:

- Added an `onPostBootstrap` hook to our Gatsby config to move assets into the expected directory (`/brand/assets`) in development. This means that, in both dev and prod environments, the assets are in the same relative location. This means we don't need to add any special logic to the codebase which determines asset location based on environment.

- Check stories and documentation which use the `VideoPlayer` component to ensure that the video loads as expected

- https://github.com/github/primer/issues/3772

## Contributor checklist:

## Reviewer checklist:

### DIFF
--- a/apps/docs/gatsby-node.js
+++ b/apps/docs/gatsby-node.js
@@ -41,15 +41,15 @@ exports.onPostBootstrap = async ({reporter}) => {
     return
   }
 
-  const staticDir = path.resolve(__dirname, 'static', 'assets')
-  const staticOutDir = path.resolve(__dirname, 'public', 'brand', 'assets')
+  const srcDir = path.resolve(__dirname, 'static', 'assets')
+  const outDir = path.resolve(__dirname, 'public', 'brand', 'assets')
 
   try {
     // eslint-disable-next-line i18n-text/no-en
-    reporter.info(`Copying static assets from ${staticDir} to ${staticOutDir}`)
-    await fs.mkdir(staticOutDir, {recursive: true})
+    reporter.info(`Copying static assets from ${srcDir} to ${outDir}`)
+    await fs.mkdir(outDir, {recursive: true})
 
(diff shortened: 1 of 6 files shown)


apps/docs/gatsby-node.js
```diff
@@ -28,3 +30,27 @@ exports.onCreateWebpackConfig = ({actions, loaders, plugins, getConfig}) => {
...
+  // eslint-disable-next-line i18n-text/no-en
+  reporter.info('Running post-bootstrap tasks...')
+
+  if (process.env.NODE_ENV === 'production') {
+    // eslint-disable-next-line i18n-text/no-en
+    reporter.info('No post-bootstrap tasks to run in production environment.')
+    return
+  }
+
+  const staticDir = path.resolve(__dirname, 'static', 'assets')
```
rezrah, line 44: nit, but could we call this `srcDir` and the other `outDir` please to make it immediately obvious what these are...

rezrah (approved):
Looks good ... also fine to use features in the LTS 👍

PR #707, opened 2024-08-21
Add footnotes component

rezrah:
## Summary

Part of https://github.com/github/primer/issues/3673

Adds a new Footnotes component to the library.

Displays two types of footnotes content: `citation` and `disclaimer`.

🔗  [Documentation preview](https://primer-565d161980-26139705.drafts.github.io/brand/components/Footnotes#ai)
🔗  [Real-world example](https://primer-565d161980-26139705.drafts.github.io/brand/storybook/?path=/story/components-footnotes-examples--rivers-with-citations)
🔗  [Citation examples in Storybook](https://primer-565d161980-26139705.drafts.github.io/brand/storybook/?path=/story/components-footnotes-features--citations-with-link)
🔗  [Paragraph (disclaimer) examples in Storybook](https://primer-565d161980-26139705.drafts.github.io/brand/storybook/?path=/story/components-footnotes-features--disclaimer-variant)

## List of notable changes:

- added component
- added documentation
- added storybook stories
- added unit tests
- added changeset
- added VRT

- Does it [look like the designs?](https://www.figma.com/design/BJ95AjraesmRCWsKA013GS/Primer-Brand?node-id=14530-1570&m=dev) 
- Does it work as expected?

## Steps to test:

Use the following links:

-  [Documentation preview](https://primer-565d161980-26139705.drafts.github.io/brand/components/Footnotes#ai)
-   [Real-world example](https://primer-565d161980-26139705.drafts.github.io/brand/storybook/?path=/story/components-footnotes-examples--rivers-with-citations)
-   [Citation examples in Storybook](https://primer-565d161980-26139705.drafts.github.io/brand/storybook/?path=/story/components-footnotes-features--citations-with-link)
-   [Paragraph (disclaimer) examples in Storybook](https://primer-565d161980-26139705.drafts.github.io/brand/storybook/?path=/story/components-footnotes-features--disclaimer-variant)

- [Design file](https://www.figma.com/design/BJ95AjraesmRCWsKA013GS/Primer-Brand?node-id=14530-1570&m=dev)
- [Tracking issue](https://github.com/github/primer/issues/3673)

## Contributor checklist:

## Reviewer checklist:

## Screenshots:

Screenshot 2024-08-21 at 17 28 22

Screenshot 2024-08-21 at 17 28 48


simmonsjenna (approved):
LGTM — thank you! 🙌

packages/react/src/Footnotes/Footnotes.tsx
```diff
@@ -0,0 +1,112 @@
...
+
+/** * Main Stylesheet (as a CSS Module) */
+import styles from './Footnotes.module.css'
+
+export const FootnotesTags = ['div', 'ol'] as const
+
+type FootnotesRestrictedPolymorphism =
+  | (React.HTMLAttributes<HTMLDivElement> & BaseProps<HTMLDivElement> & {as?: 'div'})
+  | (React.HTMLAttributes<HTMLParagraphElement> & BaseProps<HTMLOListElement> & {as?: 'ol'})
+
```
joshfarrant, line 21: I don't think we need `BaseProps` here do we? I don't think it's accurate either as we don't support the `animate` prop.

packages/react/src/Footnotes/Footnotes.tsx
```diff
@@ -0,0 +1,112 @@
...
+}
+
+type CitationFootnotesItemProps = {
+  variant?: 'citation'
+  href?: string
+}
+
+type FootnotesItemProps = (DefaultFootnotesItemProps | CitationFootnotesItemProps) &
+  React.HTMLAttributes<HTMLParagraphElement> &
+  BaseProps<HTMLParagraphElement>
```
joshfarrant, line 82: Same here, don't think we need `BaseProps`

packages/react/src/Footnotes/Footnotes.tsx
```diff
@@ -0,0 +1,112 @@
...
+  return (
+    <>
+      <Heading className="visually-hidden" as="h2" id="footnote-label">
+        {visuallyHiddenHeading}
+      </Heading>
+      <ol
+        className={clsx(styles.Footnotes, styles['Footnotes--variant-citations'], className)}
+        {...(rest as BaseProps<HTMLOListElement>)}
+      >
+        {children}
```
joshfarrant, line 63: Do you think we need to guard against someone passing `as="ol"` with `variant="disclaimer"`? That would render an `<ol>` with `<Text>` children.

packages/react/src/Footnotes/Footnotes.test.tsx
```diff
@@ -0,0 +1,110 @@
...
+    const {getByRole} = render(
+      <Footnotes>
+        <Footnotes.Item>{mockItemText}</Footnotes.Item>
+      </Footnotes>,
+    )
+
+    const visuallyHiddenHeading = getByRole('heading', {name: expectedHeading})
+    expect(visuallyHiddenHeading).toBeInTheDocument()
+    expect(visuallyHiddenHeading.tagName).toEqual('H2')
+    expect(visuallyHiddenHeading.className).toContain('visually-hidden')
```
joshfarrant, line 87: This can be condensed slightly

```suggestion
    const visuallyHiddenHeading = getByRole('heading', {name: expectedHeading, level: 2})
    expect(visuallyHiddenHeading).toBeInTheDocument()
    expect(visuallyHiddenHeading.className).toContain('visually-hidden')
```

packages/react/src/Footnotes/Footnotes.test.tsx
```diff
@@ -0,0 +1,110 @@
...
+
+    const iconLinkEl = itemEl.querySelector('a')
+    expect(iconLinkEl).toBeInTheDocument()
+
+    expect(iconLinkEl).toHaveAttribute('href', mockHref)
+    expect(iconLinkEl).toHaveAttribute('aria-label', 'Back to content')
+
+    const iconEl = iconLinkEl?.querySelector('svg')
+    expect(iconEl).toBeInTheDocument()
+  })
```
joshfarrant, line 109: Just a suggestion, but you could leverage testing library a bit more in this test and have the test more closely match how the user would interact with the page.


```tsx
  it('appends an icon with an aria-label when href is passed', async () => {
    const mockHref = 'https://github.com'
    const {getByText, getByRole} = render(
      <Footnotes>
        <Footnotes.Item href={mockHref}>{mockItemText} </Footnotes.Item>
      </Footnotes>,
    )

    const itemEl = getByText(mockItemText)
    expect(itemEl).toBeInTheDocument()

    const backLink = getByRole('link', {name: 'Back to content'})
    expect(backLink).toBeInTheDocument()

    expect(backLink).toHaveAttribute('href', mockHref)

    const iconEl = getByRole('img', {hidden: true})
    expect(iconEl).toBeInTheDocument()
  })
```

packages/react/src/Footnotes/Footnotes.tsx
```diff
@@ -0,0 +1,112 @@
...
+type FootnotesRestrictedPolymorphism =
+  | (React.HTMLAttributes<HTMLDivElement> & BaseProps<HTMLDivElement> & {as?: 'div'})
+  | (React.HTMLAttributes<HTMLParagraphElement> & BaseProps<HTMLOListElement> & {as?: 'ol'})
+
+type FootnotesProps = {
+  as?: (typeof FootnotesTags)[number]
+  visuallyHiddenHeading?: string
+} & FootnotesRestrictedPolymorphism
+
+function Root({as = 'ol', className, children, visuallyHiddenHeading = 'Footnotes', ...rest}: FootnotesProps) {
```
joshfarrant, line 27: Just a suggestion, but do you think this component should return a `<section>`, rather than just a heading adjacent to a `div`/`ol`?

joshfarrant (commented):
Looking great!

packages/react/src/Footnotes/Footnotes.test.tsx
```diff
@@ -0,0 +1,110 @@
...
+
+    let index = 0
+    for (const item of items) {
+      const itemEl = getByTestId(`item-${index}`)
+      expect(itemEl).toHaveTextContent(item)
+      expect(itemEl).toBeInTheDocument()
+      expect(itemEl.tagName).toEqual('LI')
+      index++
+    }
+  })
```
joshfarrant, line 48: Just a suggestion, but you could get rid of test-ids completely here. Testing Library lists them as their [lowest-priority queries](https://testing-library.com/docs/queries/about/#priority) and as such they recommend avoiding them unless absolutely necessary

```tsx
  it('renders the default footnotes correctly', async () => {
    const items = ['Citation 1', 'Citation 2', 'Citation 3']

    const {getByRole, container} = render(
      <Footnotes>
        {items.map((citation, index) => (
          <Footnotes.Item data-testid={`item-${index}`} key={citation}>
            {citation}
          </Footnotes.Item>
        ))}
      </Footnotes>,
    )

    const results = await axe(container)
    expect(results).toHaveNoViolations()

    const listEl = getByRole('list')

    expect(listEl).toBeInTheDocument()
    expect(listEl.tagName).toEqual('OL')

    for (const item of items) {
      expect(getByRole('listitem', {name: item})).toBeInTheDocument()
    }
  })
```

packages/react/src/Footnotes/Footnotes.tsx
rezrah: BaseProps is a general-purpose wrapper around all components in Primer Brand, so I'd prefer to leave that in unless it's causing problems. It gives us the options to centralise certain props in future. Fair point around `animate`. The context there is that _every_ Primer Brand component was meant to animate in the past, but at some point we decided that shouldn't be the case. I think we should consider deprecating it in future. Will add to maintainer sync agenda.

packages/react/src/Footnotes/Footnotes.tsx
rezrah: Responded [here](https://github.com/primer/brand/pull/707#discussion_r1727166349)

packages/react/src/Footnotes/Footnotes.test.tsx
rezrah: Thanks, gone with a hybrid approach.. Used the byRole matcher for the main link but prefer using query selector for targeting the svg because it's more specific about the element type (should be svg instead of img).

packages/react/src/Footnotes/Footnotes.tsx
rezrah: I wondered the same but went with a fragment because these footnotes are intended to be used inside other semantically descriptive elements like `<footer>`, and not used standalone. (It will eventually be used in MinimalFooter). I think it makes the markup more verbose and complicated, but happy to add this in if there's a technical reason for needing it like improved accessibility.

packages/react/src/Footnotes/Footnotes.tsx
joshfarrant: Interesting, thanks. I had wondered about `BaseProps` in the past because it doesn't really seem to do anything other than add the `animate` prop. Assuming all of our components extend `HTMLAttributes<HTMLWhateverElement>` then they already have `id`, `className`, and `ref` types.

packages/react/src/Footnotes/Footnotes.tsx
joshfarrant: I was thinking mainly on the accessibility front, but testing with VoiceOver on Mac (admittedly it's not the greatest) it doesn't identify a `section` as a landmark. From my searching it looks like it's the same for JAWS, so happy to leave as is for now 🙂

packages/react/src/Footnotes/Footnotes.test.tsx
rezrah: Thanks.

`expect(getByRole('listitem', {name: item})).toBeInTheDocument()`

☝️ doesn't work because of the way the content is nested inside another tag of the listitem. Relevant WCAG guidelines [here](https://www.w3.org/TR/wai-aria-1.2/#namecalculation). I've instead used the `textContent` property on the listitem to infer the correct value. IMO it's a bit more convoluted now, but happy to keep it like this. 

If you're wondering why the assertion against tagName is still there, it's because the `listitem` role can be manually assigned to a different tag, so this makes sure that doesn't happen/

packages/react/src/Footnotes/Footnotes.tsx
rezrah: Good question. variant isn't documented anywhere outside of source code, and I'm using it as an internal-only prop. Bit of a trade-off between adding more code to guard it vs the managing the complexity of adding them.

In this case, because its undocumented I think i prefer to not add the code, though I have now prepended `_` to the start to make it clearer for TS users. Maybe this is enough of a guard? Could also go more heavy handed, add warnings, pass the `as` prop down and add conditional logic to prevent this configuration from being possible but we could always add that down the line too if we're seeing a lot of misuse.

joshfarrant (approved):
Looking great!

danielguillan:
Footnotes highlight

From a design perspective, the footnotes highlight doesn't really work. In dark mode, the contrast between the text and background is insufficient (4.22). Additionally, the yellow in dark mode clashes with the rest of the page. Finally, the highlight should only wrap the text, including the counter, instead of going full-width. 

To unblock this PR. I suggest we remove the highlight from this initial version of the component while we figure out a better visual treatment.

rezrah:
> To unblock this PR. I suggest we remove the highlight from this initial version of the component while we figure out a better visual treatment.

@danielguillan the target highlight has been removed

danielguillan (approved):

PR #709, opened 2024-08-22
Render SubNav.Heading as a heading

joshfarrant:
## List of notable changes:

- Updated `SubNav.Heading` to render a `h2`, rather than a `div`
- Added `as` prop to `SubNav.Heading`
- Updated `SubNav.Heading` documentation to include `as` prop
- Fixed an issue where the submenu toggle was rendered as an empty button in the mobile SubNav

- https://github.com/github/primer/issues/3801

## Contributor checklist:

## Reviewer checklist:

## Screenshots:

<table>
<tr>
</tr>
<tr>
<td valign="top">

image

 </td>
<td valign="top">

image

</td>
</tr>
</table>


danielguillan (commented):
It's looking great!

Should we include some guidance around how to use the correct heading level for this component? I'm thinking about [this example](https://primer-aace050369-26139705.drafts.github.io/brand/storybook/?path=/story/recipes-feature-previews-level-2--level-two-point-four) where SubNav (using h2) appears before a Hero (using h1).

apps/docs/content/components/SubNav.mdx
```diff
@@ -99,9 +99,10 @@ import {SubNav} from '@primer/react-brand'
 
 ### SubNav.Heading <Label>Required</Label>
 
-| name   | type     | default | required | description                      |
-| ------ | -------- | ------- | -------- | -------------------------------- |
-| `href` | `string` |         |          | URL of the page the link goes to |
+| name   | type                                                                            | default | required | description                      |
+| ------ | ------------------------------------------------------------------------------- | ------- | -------- | -------------------------------- |
+| `href` | `string`                                                                        |         | `true`   | URL of the page the link goes to |
+| `as`   | <PropTableValues values={['h1', 'h2', 'h3', 'h4', 'h5', 'h6']} addLineBreaks /> | `'h2'`  | `false`  | Heading level                    |
```
rezrah, line 105: Can you spread the values of [HeadingTags](https://github.com/primer/brand/blob/main/packages/react/src/Heading/Heading.tsx#L8) in here?

packages/react/src/SubNav/SubNav.tsx
```diff
@@ -158,19 +158,24 @@ const _SubNavRoot = memo(({id, children, className, 'data-testid': testId, fullW
 type SubNavHeadingProps = {
   href: string
   'data-testid'?: string
+  as?: 'h1' | 'h2' | 'h3' | 'h4' | 'h5' | 'h6'
```
rezrah, line 161: You could Pick the tags from Heading, as they're already exported there?

https://github.com/primer/brand/blob/main/packages/react/src/Heading/Heading.tsx#L8

packages/react/src/SubNav/SubNav.tsx
```diff
@@ -158,19 +158,24 @@ const _SubNavRoot = memo(({id, children, className, 'data-testid': testId, fullW
...
 } & PropsWithChildren<React.HTMLProps<HTMLAnchorElement>> &
   BaseProps<HTMLAnchorElement>
```
rezrah, line 163: We might need to refactor this too? The element type is still a HTMLAnchorElement, but the root element isn't. This might create some confusion, particularly with the introduction of an `as` prop which specifies heading tags.

packages/react/src/SubNav/SubNav.module.css
```diff
@@ -23,6 +23,8 @@
   z-index: 9998;
   display: inline-block;
   margin-inline-end: var(--base-size-16);
+  font-size: var(--base-size-16);
+  line-height: var(--base-size-24);
```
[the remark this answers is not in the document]
joshfarrant: Great spot, thanks!

packages/react/src/SubNav/SubNav.tsx
joshfarrant: Good suggestion, thanks

packages/react/src/SubNav/SubNav.tsx
joshfarrant: What are you thinking it should be changed to?

The props get spread onto an `<a>` element regardless of what the `as` prop is set to, so I think `HTMLProps<HTMLAnchorElement>` is correct in this case.

rezrah: > The props get spread onto an <a> element regardless of what the as prop is set to

That's implementation detail. The user may find it confusing because the API now let's you do this:

```jsx
<SubNav.Heading href="#" as="h1" />
```

<h2> is now the new root element, not a link. Which in hindsight, how it probably should have been done in the first place 😅 

Happy to jam on some ideas with you over a call if easier..

in the meantime, some ideas:

- change to HTMLHeadingElement and spread the props into the root / heading instead
  - Makes more sense given the child is named `Heading`
  - add `href` as an optional, additional prop here, and leave a description / comment that it sets the href on the nested, optional link
  - This is gonna be a breaking change
  
- switch to `headingAs` and leave the types as is (not ideal, and I think sets a precedent so I think first option is better)

joshfarrant: Take a look at the approach in https://github.com/primer/brand/pull/709/commits/d46affe04f561eb0c3512c0ccb64a6ff2407348c

This leaves the rendering of the link up to the consumer, meaning we don't have to worry about supporting anchor link props now, or in the future. Consumers can add a `href`, `target`, `onClick`, `data-*` attribute, or anything else they like without us needing to modify the `SubNav.Heading` component.

```tsx
<SubNav.Heading as="h3">
  <a
    href="#"
    target="_blank"
    rel="noopener noreferrer"
    onClick={sendAnalytics}
    data-analytics="nav-heading"
    {/* or anything else they'd like to pass... */}
  >
    GitHub
  </a>
</SubNav.Heading>
```

PR #706, opened 2024-08-21
Fix a11y contrast of duotone text across components

danielguillan:
## Summary

This pull request guarantees a minimum text contrast between emphasized and muted text on components using a duotone text treatment.

## List of notable changes:

- Adds a new `gray-10` token (`gray-13` on the new Primitives scale) to increase contrast and achieve a 3.1:1 contrast ratio agains `gray-5` (subtle). 
- Adds a new `emphasis` text color token used on `em` elements in duotone text.
- Updates `gray-5` to achieve 4.5:1 contrast ratio against a `gray-1` background (subtle).
- Updates components with duotone styling (Bento, SectionIntro, Testimonial, Timeline and RiverBreakout)  to use `text-subtle` and `text-emphasis` on light color modes to reach 3.1:1 contrast ratio.

## Steps to test:

1. Review the [token updates](https://github.com/primer/brand/pull/706/files#diff-bd84586063ff835c411a57b5622e94342ec94cafe9d3e6417ded065b94f60791)
2. Review the [Bento heading with emphasized text story](https://primer-cf011bbb41-26139705.drafts.github.io/brand/storybook/?path=/story/components-bento-item--heading-with-emphasized-text)
3. Review the [SectionIntro with emphasized text story](https://primer-cf011bbb41-26139705.drafts.github.io/brand/storybook/?path=/story/components-sectionintro-features--with-emphasized-text)
4. Review the [Testimonial large highlighted portion story](https://primer-cf011bbb41-26139705.drafts.github.io/brand/storybook/?path=/story/components-testimonial--large-highlighted-portion)
5. Review the [Timeline with emphasis story](https://primer-cf011bbb41-26139705.drafts.github.io/brand/storybook/?path=/story/components-timeline-features--with-emphasis)

## Supporting resources (related issues, external links, etc):

- Fixes https://github.com/github/primer/issues/3791
- Fixes https://github.com/github/primer/issues/3740

## Contributor checklist:

- [x] All new and existing CI checks pass
- [x] Tests prove that the feature works and covers both happy and unhappy paths
- [x] Any drop in coverage, breaking changes or regressions have been documented above
- [ ] New visual snapshots have been generated / updated for any UI changes
- [x] All developer debugging and non-functional logging has been removed
- [x] Related issues have been referenced in the PR description

## Reviewer checklist:

- [ ] Check that pull request and proposed changes adhere to our [contribution guidelines](../../CONTRIBUTING.md) and [code of conduct](../../CODE_OF_CONDUCT.md)
[...]

PR #722, opened 2024-08-30
Add support for duotone text with <b> elements by default

danielguillan:
## Summary

Components using duotone text treatments now use the `<b>` element by default. The `<em>` element is still supported to avoid breaking changes, but the documentation and stories have been updated to only reference the use of `<b>`.

- 🔗 [Docs](https://primer-6ab1159603-26139705.drafts.github.io/brand/)
- 🔗 [Storybook](https://primer-6ab1159603-26139705.drafts.github.io/brand/storybook/)

## List of notable changes:

- Bento, RiverBreakout, SectionIntro, Timeline, and Testimonial now support `b` elements to enable the duotone text treatment.
- The documentation has been updated to recommend using `b` exclusively.
- Stories now use `b` instead of `em`.
- Tests have been updated to expect both `b` and `em` to enable duotone text.

_Please note that we'll be able to support a CSS-only solution for duotone text when we are able to leverage the :has() pseudo-class. In the meantime, and to avoid breaking changes, we'll continue to support `em` elements at the component-level but only recommend using `b`._

- Verify that duotone text stories use `b` instead of `em`.
- Verify that the documentation and examples have been updated to only reference `b` usage.
- Ensure that duotone text is working for both `b` and `em` elements.

- Replaces #706 
- Closes https://github.com/github/primer/issues/3791
- Closes https://github.com/github/primer/issues/3740 

## Contributor checklist:

## Reviewer checklist:


joshfarrant (approved):
Looks great!

Interesting to see the discussion in https://github.com/github/primer/issues/3740 about whether to `<em>` or not too 👍

packages/react/src/DuotoneText/DuotoneText.module.css
```diff
@@ -0,0 +1,7 @@
+.DuotoneText {
+  color: var(--duotone-color, var(--brand-DuotoneText-text-color-muted));
```
rezrah, line 2: Why do we need `--duotone-color`? I can't see a reference to it anywhere. We probably also want to fix this to `var(--brand-DuotoneText-text-color-muted)` anyway?

packages/react/src/DuotoneText/DuotoneText.module.css
```diff
@@ -0,0 +1,7 @@
...
+}
+
+.DuotoneText__emphasis {
+  color: var(--duotone-color, var(--brand-DuotoneText-text-color-emphasis, red));
```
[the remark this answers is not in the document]
rezrah: or assuming `--duotone-color` is needed 

```suggestion
  color: var(--duotone-color, var(--brand-DuotoneText-text-color-emphasis));
```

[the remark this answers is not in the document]
rezrah: I might totally be jumping the gun here though, as this is still a draft 😅.. so happy to defer a proper code review till later, lmk

apps/docs/content/components/DuotoneText/index.mdx
```diff
@@ -0,0 +1,42 @@
...
+  </Do>
+  <Dont>
+    <img alt="" src={dontEmphasis} />
+    <Caption>Don't emphasize other parts of the text.</Caption>
+  </Dont>
+</DoDontContainer>
+
+### Supported components
+
+We recommend using duotone text exclusively on components that have been designed to support this treatment:
```
rezrah, line 36: It's going to be really difficult to enforce this IMO, as you can use DuotoneText inside any Text or Heading component. Is this something we need to be careful about? The alternative is to only enable this pattern to work inside of the components you've mentioned.

apps/docs/content/components/DuotoneText/react.mdx
```diff
@@ -0,0 +1,135 @@
...
+  <RiverBreakout.Content>
+    <Text>
+      <DuotoneText>
+        <DuotoneText.Emphasis>Lorem ipsum dolor sit amet,</DuotoneText.Emphasis>{' '}
+        consectetur adipiscing elit. In sapien sit ullamcorper id. Aliquam
+        luctus sed turpis felis nam pulvinar risus elementum.
+      </DuotoneText>
+    </Text>
+    <Link href="#">Call to action</Link>
+  </RiverBreakout.Content>
```
rezrah, line 76: Does DuotoneText need to be a child of Text? Could we instead do something like this:

```suggestion
 <RiverBreakout.Content>
    <DuotoneText>
      <DuotoneText.Emphasis>Lorem ipsum dolor sit amet,</DuotoneText.Emphasis>{' '}
      consectetur adipiscing elit. In sapien sit ullamcorper id. Aliquam
      luctus sed turpis felis nam pulvinar risus elementum.
    </DuotoneText>
    <Link href="#">Call to action</Link>
  </RiverBreakout.Content>
```

This would make DuotoneText functionally equivalent to Text, which i'm not actually sure is a good thing, but it could cut down on the amount of code needed to make the pattern work.

I'm not 100% on this idea 😄 

Another idea:

```jsx
<Text>
  <Text.Emphasis>Lorem ipsum dolor sit amet,</Text.Emphasis>{" "}
   consectetur adipiscing elit. In sapien sit ullamcorper id. Aliquam luctus sed turpis felis nam pulvinar risus elementum.
</Text>
```

But at that point, it's probably easier to just stick with a visually-oriented HTML child like `b` or `strong` 🤷

joshfarrant (approved):

.changeset/nervous-bees-guess.md
```diff
@@ -0,0 +1,18 @@
+---
+'@primer/react-brand': patch
+---
+
+Added support for emphasized text using `<b>` elements to Bento, RiverBreakout, SectionIntro, Timeline, and Testimonial components.
```
rezrah, line 5: Could we mention here that we've done this to ensure the emphasised text is a purely visual treatment and won't be picked up by assistive tech? It might not be immediately clear to reads of the change log entry why we've done this otherwise, and why it's important.

Something like..

```suggestion
Added support for `<b>` elements to `Bento`, `RiverBreakout`, `SectionIntro`, `Timeline`, and `Testimonial` components, as `<em>` is semantically interpreted by screen readers for emphasis, while `<b>` allows duo-tone text to function purely as a visual decoration.
```

.changeset/nervous-bees-guess.md
```diff
@@ -0,0 +1,18 @@
...
+
+```jsx
+<SectionIntro>
+  <SectionIntro.Heading>
+    <b>Expressive headline</b> about an exclusive set of features.
+  </SectionIntro.Heading>
+</SectionIntro>
+```
+
+> **Warning**
```
rezrah, line 17:
```suggestion
> **Important**
```

rezrah (approved):
Nice one @danielguillan, this is great 🙌

PR #726, opened 2024-09-03
Using upstream config from primer/primitives

lukasoppermann:
## Summary

This fixes https://github.com/github/primer/issues/3891

## List of notable changes:

-
-

- 
-

## Steps to test:

1.
1.
1.

-
-

## Contributor checklist:

## Reviewer checklist:

## Screenshots:

<table>
<tr>
</tr>
<tr>
<td valign="top">

 </td>
<td valign="top">

</td>
</tr>
</table>

### DIFF
--- a/packages/design-tokens/package.json
+++ b/packages/design-tokens/package.json
@@ -33,7 +33,9 @@
     "test": "jest"
   },
   "devDependencies": {
-    "@primer/primitives": "9.0.2",
+    "@primer/primitives": "9.1.0",
+    "prettier": "^2.8.8",
+    "@prettier/sync": "^0.5.2",
     "color2k": "^2.0.2",
     "fast-glob": "3.3.2",
     "style-dictionary": "3.8.0"
--- a/packages/design-tokens/src/tokens/functional/size/viewport.json
+++ b/packages/design-tokens/src/tokens/functional/size/viewport.json
@@ -1,27 +1,27 @@
 {
   "brand": {
     "viewportRange": {
-      "narrow": {
+      "narrow-viewport": {
         "$value": "(max-width: calc({brand.breakpoint.medium} - 0.02px))",
         "$type": "custom-viewportRange"
       },
-      "narrowLandscape": {
+      "narrowLandscape-viewport": {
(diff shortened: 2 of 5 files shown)


packages/design-tokens/src/tokens/functional/size/viewport.json
```diff
@@ -1,27 +1,27 @@
 {
   "brand": {
     "viewportRange": {
-      "narrow": {
+      "narrow-viewport": {
```
lukasoppermann, line 4: @rezrah this is added to be able to replace your custom media format. 

However, I think it would be better to just remove the `-viewport` part. You already have `viewportRange` before the name, so this should be enough.

packages/design-tokens/package.json
```diff
@@ -33,7 +33,9 @@
     "test": "jest"
   },
   "devDependencies": {
-    "@primer/primitives": "9.0.2",
+    "@primer/primitives": "9.1.0",
+    "prettier": "^2.8.8",
```
rezrah, line 37: We should already have this available in the root for all workspaces: https://github.com/primer/brand/blob/main/package.json#L68

You can keep `@prettier/sync` though, could that be a dependency of `primer/primitives` instead? Reason being that I'm not seeing any direct imports to that library, which indicates that it's an 3rd party dependency.

packages/design-tokens/src/tokens/functional/size/viewport.json
rezrah: That seems reasonable @lukasoppermann. We'd need to just add a changeset to indicate this change. 

Ideally we just grab this file from `primer/primitives` though, is that still the plan and just out of scope for this PR?

rezrah:
Thanks @lukasoppermann 🙌  👨‍🍳 

Screenshot 2024-09-03 at 09 34 58

and

Screenshot 2024-09-03 at 09 31 07

👏 👏 👏 

(will follow up in DMs with some other minor comments, but this is looking soo good)

rezrah (approved):

packages/design-tokens/package.json
lukasoppermann: https://github.com/github/primer/issues/3903

packages/design-tokens/src/tokens/functional/size/viewport.json
lukasoppermann: Yes, this would be optimal.

Issue #203, opened 2023-03-09
🐛 [BUG - React] - FormControl.Validation layout problems with Checkbox


rezrah:
The following snippet:

```jsx
<FormControl validationStatus='error'>
  <FormControl.Label>Day</FormControl.Label>
  <Checkbox name="day" value="" />
  <FormControl.Validation className="test">Please select at least one day</FormControl.Validation>
</FormControl>
```

Produces this:

![validation message appearing before checkbox on the same line](https://user-images.githubusercontent.com/13340707/224012648-1e95b010-f071-4707-9d42-1def86677879.png)

This happens because of `row-reverse` flex layout method. 

### Expected behavior

- Validation appears below the input and label per the usual validation layout: https://primer.style/brand/components/FormControl#validation

MadMango:
![image](https://github.com/primer/brand/assets/7106600/02b8a804-748c-495b-8f69-4d56ea0f16b3)

I had a quick stab at this in #513

MelissaPastore:
Closing this out since it should be fixed by the changes in https://github.com/primer/brand/pull/728

PR #728, opened 2024-09-03
Fix checkbox input validation positioning 

MelissaPastore:
## Summary

Ref: https://github.com/primer/brand/issues/203

This PR fixes a bug with the positioning of the validation message for checkbox inputs inside a FormControl component.

## List of notable changes:

Styling and layout updates:

* [`packages/react/src/forms/FormControl/FormControl.module.css`](diffhunk://#diff-66f88db62ba93924cf6de6ae216861c98149f0122e02c93173b5778c43cd4b57L13-R14): Changed `.FormControl--checkbox` to use a grid layout instead of flex and added `.FormControl-validation-checkbox` to span columns. [[1]](diffhunk://#diff-66f88db62ba93924cf6de6ae216861c98149f0122e02c93173b5778c43cd4b57L13-R14) [[2]](diffhunk://#diff-66f88db62ba93924cf6de6ae216861c98149f0122e02c93173b5778c43cd4b57R117-R120)
* [`packages/react/src/forms/Checkbox/Checkbox.module.css`](diffhunk://#diff-73b35ce349fc37e97410e2b629ab71dc02d3268812d5e4eed53fa33aa18c3ad7R6): Added `grid-row: 1` to `.Checkbox-wrapper`.
* [`packages/react/src/forms/Radio/Radio.module.css`](diffhunk://#diff-cebde5d1553a837694f74ac3f39e23d3cecbdac76855ed93e741be0df483ed5dR11): Added `grid-row: 1` to the radio button styling.

Storybook updates:

* [`packages/react/src/forms/FormControl/FormControl.stories.tsx`](diffhunk://#diff-27807d60e3cf04e8fc7c732bc664f14b6ee97086aad369c1d6accd00ea7c2bc8R337-R343): Added a new story `ErrorValidationWithCheckbox` for error validation with checkboxes.

Are there any unintended regressions this might cause? 

I looked at the current storybook example form here, as well as the checkbox and radio storybook examples and they all seem to have the intended behavior after my changes, but let me know if I missed anything! 

## Steps to test:

1. Open the FormControl component in CI-deployed preview environment
2. Go to  'w/ Checkbox with Error Validation - Playground' in Storybook
3. Verify that the validation now appears below the checkbox instead of next to them. 

- Related issue: https://github.com/primer/brand/issues/203
- Previous PR that attempted to fix this issue: https://github.com/primer/brand/pull/513

## Contributor checklist:

## Reviewer checklist:

## Screenshots:

<table>
<tr>
</tr>
<tr>
<td valign="top">

 </td>
<td valign="top">

</td>
</tr>
</table>


packages/react/src/forms/FormControl/FormControl.module.css
```diff
@@ -116,3 +116,8 @@
     transform: translateY(0);
   }
 }
+
+.FormControl-validation-checkbox {
+  position: absolute;
+  top: 45px;
+}
```
rezrah, line 123: 👋 @MelissaPastore 

> What do you think of this approach of using CSS to fix this issue? 
> I'm not sure if it's good practice to have a position: absolute element within a flex container?

Thanks for calling this out ✨. While it does work with `position: absolute`, you could also use a grid to achieve a similar effect without needing to manually set a the vertical offset, or change the markup.

E.g. something like this...

<table>

<tr>
<td valign="top">


Screenshot 2024-09-09 at 11 06 37


 </td>
<td valign="top">


Screenshot 2024-09-09 at 11 07 36
</td>
</tr>
</table>

You'll notice that you still need to apply some optical alignment to the checkbox, but this should only be around 1-2px. Another advantage to this approach is that it will preserve the `gap` that's already set in `FormControl`

Here's the CSS I used in the inspector to achieve the layout above, but it can likely be optimized further...

```css
.formcontrol-parent {
  display: grid;
  grid-template-columns: auto 1fr;
}

label {
  grid-column: 2; 
  grid-row: 1;
}

input {
  grid-row: 1; 
}

.validation {
  grid-column: 1 / 3;
}
```

Hope this helps!

packages/react/src/forms/FormControl/FormControl.stories.tsx
```diff
@@ -334,3 +334,25 @@ export const ErrorValidation = () => {
...
+export const ErrorValidationWithRadio = () => {
+  return (
+    <FormControl validationStatus="error">
+      <FormControl.Label>Day</FormControl.Label>
+      <Radio name="day" value="" />
+      <FormControl.Validation>Please select at least one day</FormControl.Validation>
+    </FormControl>
+  )
+}
+ErrorValidationWithRadio.storyName = 'w/ Error Validation with Radio - Playground'
```
rezrah, line 358: As the JSX is almost identical to the existing `CheckboxPlayground` story, it might be helpful to reuse it with different arguments. This should save a few lines, and keep all checkbox stories consistent.

E.g. 

```jsx 
export const ErrorValidationWithCheckbox = args => <CheckboxPlayground {...args} />
ErrorValidationWithCheckbox.storyName = 'w/ Error Validation with Checkbox - Playground'
ErrorValidationWithCheckbox.args = {
  validationStatus: 'error',
}
```

packages/react/src/forms/FormControl/FormControl.module.css
MelissaPastore: Thanks for the feedback. I'll try this approach!

packages/react/src/forms/FormControl/FormControl.stories.tsx
MelissaPastore: Good suggestion - I'll update that!

packages/react/src/forms/FormControl/FormControl.module.css
joshfarrant: +1 to @rezrah's grid approach 🙂

Looks great though @MelissaPastore! ✨

packages/react/src/forms/Checkbox/Checkbox.module.css
```diff
@@ -5,6 +5,13 @@
   position: relative;
 }
 
+
+[class*="FormControl"] .Checkbox-wrapper,
```
MelissaPastore, line 9: This might be overkill but I wasn't sure if we wanted to always apply the grid-row to the wrapper for the input or only do this when its parent is a form control element?

packages/react/src/forms/FormControl/FormControl.stories.tsx
```diff
@@ -296,12 +296,24 @@ export const CheckboxPlayground = args => {
 
 CheckboxPlayground.storyName = 'w/ Checkbox - Playground'
 
-export const RadioPlayground = () => (
+export const RadioPlayground = args => (
   <>
     <Stack direction={{narrow: 'vertical', regular: 'horizontal'}} gap="condensed" padding="condensed">
-      <FormControl>
+      <FormControl
```
MelissaPastore, line 302: This is pretty much borrowed from the CheckboxPlayground implementation so I could re-use this component for the example with a validation message.

packages/react/src/forms/Checkbox/Checkbox.module.css
```diff
@@ -5,6 +5,11 @@
   position: relative;
 }
 
+[class*='FormControl'] .Checkbox-wrapper,
```
MelissaPastore, line 8: This might be overkill but I wasn't sure if we wanted to always apply the grid-row to the wrapper for the input or only do this when its parent is a form control element?

packages/react/src/forms/Radio/Radio.module.css
```diff
@@ -8,6 +8,7 @@
   -ms-user-select: none;
   user-select: none;
   cursor: pointer;
+  grid-row: 1;
```
MelissaPastore, line 11: It's necessary to add this here because we updated the styles in the FormControl--checkbox class to use flex-grid and this class gets applied to both Checkbox and Radio inputs that are children of a FormControl component.

packages/react/src/forms/Radio/Radio.module.css
```diff
@@ -8,6 +8,7 @@
...
```
MelissaPastore, line 11: It's necessary to add this here because we updated the styles in the FormControl--checkbox class to use flex-grid and this class gets applied to both Checkbox and Radio inputs that are children of a FormControl component.

MelissaPastore:
@rezrah @joshfarrant I think this is ready for another review. @rezrah and I discussed isolating the changes to just the checkbox inputs, so I made those changes and incorporated all of the previous feedback as well.

.changeset/loud-phones-exercise.md
```diff
@@ -0,0 +1,5 @@
+---
+'@primer/react-brand': patch
+---
+
+Updates the styling of validation messages for checkbox inputs to use flex-grid to fix a bug and adds a storybook example for the scenario.
```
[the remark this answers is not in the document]
MelissaPastore: Thanks for the context @joshfarrant! I'll keep that in mind going forward.

joshfarrant:
Thanks again @MelissaPastore!

PR #736, opened 2024-09-09
Use dimension/px transformer instead of custom

lukasoppermann:
## Summary

Replace custom transform with upstream equivalent

### DIFF
--- a/packages/design-tokens/src/tokens/functional/components/video-player/base.json
+++ b/packages/design-tokens/src/tokens/functional/components/video-player/base.json
@@ -13,7 +13,8 @@
             "value": "rgba(0, 0, 0, 0.75)"
           },
           "padding": {
-            "value": "10px"
+            "value": "10px",
+            "$type": "dimension"
           }
         }
       },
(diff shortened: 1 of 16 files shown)


packages/design-tokens/src/tokens/functional/components/video-player/base.json
```diff
@@ -27,21 +28,31 @@
...
         "padding": {
-          "value": "5px 10px"
+          "block": {
+            "value": "10px",
+            "$type": "dimension"
+          },
+          "inline": {
+            "value": "5px",
+            "$type": "dimension"
+          }
```
rezrah, line 55: <table>
<tr>
</tr>
<tr>
<td valign="top">

Screenshot 2024-09-09 at 12 44 38


 </td>
<td valign="top">

Screenshot 2024-09-09 at 12 47 18


</td>
</tr>
</table>


Let's set these both at 5px for now, to eliminate a regression.

```suggestion
          "block": {
            "value": "5px",
            "$type": "dimension"
          },
          "inline": {
            "value": "5px",
            "$type": "dimension"
          }
```

rezrah (approved):
LGTM, thanks @lukasoppermann

PR #753, opened 2024-09-16
Add Figma code connect forn River component

danielguillan:
## Summary

Adds basic Figma code connect to the River component. 

## List of notable changes:

- Adds Figma code connect file for the River component
- Disables `i18n-text/no-en` in Figma code connect files to support some Figma layer names

- Test River examples in dev mode on the `main` branch in Figma. 

## Screenshots:

image

### DIFF
--- a/packages/react/src/river/River.figma.tsx
+++ b/packages/react/src/river/River.figma.tsx
@@ -1,0 +1,40 @@
+import React from 'react'
+import {Heading, Link, River, Text} from '../'
+import figma from '@figma/code-connect'
+
+figma.connect(River, 'https://www.figma.com/design/BJ95AjraesmRCWsKA013GS/Primer-Brand?node-id=375-8463', {
+  props: {
+    align: figma.enum('align', {
+      start: 'start',
+      center: 'center',
+      end: 'end',
+    }),
+    imageTextRatio: figma.enum('imageTextRatio', {
+      '50:50': '50:50',
+      '60:40': '60:40',
+    }),
+    content: figma.nestedProps('Content', {
+      headingText: figma.textContent('Heading'),
+      descriptionText: figma.textContent('Description'),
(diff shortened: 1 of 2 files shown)


packages/react/src/river/River.figma.tsx
```diff
@@ -0,0 +1,40 @@
...
+        true: <Link href="#">Link text</Link>,
+        false: undefined,
+      }),
+    }),
+  },
+
+  example: ({align, imageTextRatio, content}) => (
+    <River align={align} imageTextRatio={imageTextRatio}>
+      <River.Content>
+        <>
```
danielguillan, line 29: This React fragment is required to prevent a type error for `River.Contrnt` children (i.e. `{content.link}`)

joshfarrant: Thanks for explaining this, makes sense 👍

It kinda makes me think that `RiverContentProps['children']` should include `null` and `undefined` in its type so the fragment isn't necessary. For example:

```tsx
export type RiverContentProps = {
  // ...
  children:
    | React.ReactElement<TextProps>
    | (React.ReactElement<HeadingProps | TextProps | LinkProps> | null | undefined)[]
}
```

That would allow you to ditch the fragment, and I think make the API a bit friendlier to consumers.

```tsx
<River.Content>
  <Heading>{content.headingText}</Heading>
  <Text>{content.descriptionText}</Text>
  {content.link}
</River.Content>
```

Not something I'm suggesting we change now though, but maybe something to think about in the future and implement library-wide as we continue this pattern of limiting the type of `children`.

joshfarrant (approved):
Looks great! I've still got a lot to learn with Figma, so I'd appreciate a run-through of how this works on the Figma-side at some point if you don't mind? 🙏🙂

Issue #773, opened 2024-09-27
[Modification Request] Update the icon in `Bento.Item` to appear the same as `Card` (with a background)


rfearing:
### Summary

Currently the Bento's `LeadingVisual` is an `Octicon` with no background.

There are now new design directions about how icons should render in Bento boxes, the same way they appear in cards. (c.c. @nsolerieu): 
![Figma bento design](https://github.com/user-attachments/assets/bf1634bd-1685-451e-8ceb-ff5bab6e8d67)

Corresponding issue: https://github.com/github/marketing-platform-services/issues/3553

#### Requested update:

There be an option for the circular icon as a child of `Bento.Item`

joshfarrant:
Hey @rfearing 👋

Following on from the discussions in yesterday's Primer Brand Office Hours, @danielguillan and I think the best way to proceed is with a dedicated `Icon` component.

Adding a new component to the library typically comprises of a few steps:

1. API exploration
2. Figma implementation
3. React implementation

If you're still happy to work on the implementation of this yourself — which we'd be very grateful for — then here's some information on each of those steps.


## API exploration
The purpose of an API Exploration is to flesh out the API of new components with the Primer Brand maintainers. This is where we can discuss component naming, what props the component will accept (and the types of those props), and highlight anything that is out-of-scope for this component.

As a starting point, you might want to take a look at the [`Card.Icon` component](https://github.com/primer/brand/blob/main/packages/react/src/Card/Card.tsx#L187-L209) as that already implements some props (`icon`, `color`, `hasBackground`) that we'll likely want to support in this new component.

Typically these API explorations take place in a FigJam, however we can do it any way you prefer. If you'd like, you can instead share your proposed types for the props that the component will accept as a comment in this issue. I've got some initial thoughts on this, so let me know if you'd like some support with this 🙂


## Figma implementation

After speaking with @danielguillan, I believe there is already [a private component in Figma](https://www.figma.com/design/BJ95AjraesmRCWsKA013GS/Primer-Brand?node-id=6872-32721&t=AiyaJwHNTGZ0eCPl-4) which is used in the Card component. This will need to be made into a public-facing component so it can be consumed alongside other Primer Brand components in Figma.

@danielguillan is the best person to speak to for support on this.


## React implementation

[...]

PR #796, opened 2024-10-23
Add `size` prop to `Icon` component

joshfarrant:
## Summary

Adds `size` prop to `Icon` component following @simmonsjenna's [suggested sizing](https://github.com/primer/brand/issues/773#issuecomment-2429832267), and @danielguillan's [Figma designs](https://www.figma.com/design/BJ95AjraesmRCWsKA013GS/branch/9CEUShb0wiFWbpQYpQMdpk/Primer-Brand?node-id=15397-35&node-type=canvas&t=DEcuCT9yVFeK8Aua-0).

## List of notable changes:

- Added `size` prop
- Added tests
- Updates stories
- Condensed styles

- There's an `!important` in the CSS. If you're not happy with that then I'm happy to tweak the approach to use specificity to override the background styles instead.
- Happy with the new stories?

## Steps to test:

- Play around in the Storybook playground

- Works towards https://github.com/primer/brand/issues/773

## Contributor checklist:

## Reviewer checklist:

## Screenshots:

image


joshfarrant:
NextJS seem to be in the process of messing around with their React version (they're currently on a React 19 RC), which is the underlying cause of the pipeline failure.

If you're happy to just ignore the failing check for now @rezrah then we can come back to it once it's settled down.

packages/react/src/Icon/Icon.module.css
```diff
@@ -1,121 +1,95 @@
...
+.Icon {
+  box-sizing: content-box;
+  border-radius: 9999px;
   padding: var(--base-size-12);
 }
 
-.Icon--color-default {
-  color: var(--brand-Icon-color-default);
+.Icon--noBackground {
+  background-color: initial !important;
```
joshfarrant, line 8: Cool with this `!important`?

packages/react/src/Icon/Icon.tsx
```diff
@@ -1,33 +1,81 @@
...
- * Design Tokens
- */
 import '@primer/brand-primitives/lib/design-tokens/css/tokens/functional/components/icon/colors.css'
 
-export const IconColors = Colors
-export const defaultIconColor = IconColors[0]
+export const namedIconSizes = ['small', 'medium', 'large'] as const
+export type NamedIconSize = (typeof namedIconSizes)[number]
+
+export const numericIconSizes = [12, 16, 20, 24, 28, 32, 36, 40, 44, 48, 52] as const
```
rezrah, line 12: Can we max out at `44` please [per this guidance
](https://github.com/primer/brand/issues/773#issuecomment-2429832267)

```suggestion
export const numericIconSizes = [20, 24, 28, 32, 36, 40, 44] as const
```

packages/react/src/Icon/Icon.module.css
```diff
@@ -1,121 +1,95 @@
-.Icon--badge {
-  width: var(--base-size-44);
-  height: var(--base-size-44);
-  border-radius: 50%;
-  display: flex;
-  align-items: center;
-  justify-content: center;
+.Icon {
+  box-sizing: content-box;
+  border-radius: 9999px;
```
rezrah, line 3:
```suggestion
  border-radius: var(--brand-borderRadius-full);
```

packages/react/src/Icon/Icon.module.css
rezrah: It looks like we need this because we currently apply a background color by default to all Icons, even when background is not required or `false`, is that correct?

If so, this feels a little inverted, because we shouldn't apply that style unless we need it. In an edge case it could also cause a specificity issue if someone tried to override the background color, though they could workaround that too.

If we did invert it, presumably we'd need to add a `with-background-color` class for each of the colors? In terms of lesser evils, keeping the `!important` might be okay. Any other ways to solve this @joshfarrant?

packages/react/src/Icon/Icon.module.css
```diff
@@ -1,121 +1,95 @@
...
-.Icon--color-lemon.Icon--badge {
   background-color: var(--brand-Icon-background-lemon);
 }
 
 .Icon--color-lime {
-  color: var(--brand-Label-color-lime);
-}
-
-.Icon--color-lime.Icon--badge {
+  color: var(--brand-Icon-color-lime);
```
rezrah, line 63: Nice catch!

packages/react/src/Icon/Icon.module.css
rezrah: Just noticed we used to have a separate rule for badges

`.Icon--color-{hue}.Icon--badge`

I think that's easier to reason about tbh. What was the motivation behind removing that?

packages/react/src/Icon/Icon.tsx
joshfarrant: Happy to do that if @danielguillan is happy.

These numbers have come from [Dani's FIgma designs](https://www.figma.com/design/BJ95AjraesmRCWsKA013GS/branch/9CEUShb0wiFWbpQYpQMdpk/Primer-Brand?node-id=15397-35&node-type=canvas&t=DEcuCT9yVFeK8Aua-0)

packages/react/src/Icon/Icon.module.css
joshfarrant: Ah nice, thanks!

packages/react/src/Icon/Icon.module.css
joshfarrant: Yea I'm happy to revert that change. Keeping both the `color` and `background-color` together made more sense to me and cut down on the amount of CSS in the file, but I don't have strong opinions so I'm happy to revert it

packages/react/src/Icon/Icon.tsx
rezrah: Ah cool, that's good to know. We'll need to follow up In Figma after too, as the Site guidance supersedes that. Particularly in the React version, we'd want to provide the correct guardrails so recommend we go ahead with the 20-44 range.

joshfarrant: That's updated

rezrah (approved):
Looks great! Visual snapshots will need an update, but otherwise :shipit:

packages/react/src/Icon/Icon.tsx
danielguillan: I'm happy to limit the scale and expand it later if needed. I just want to check with @simmonsjenna to confirm that we've considered scenarios beyond Bento, Card, and Pillar in making this decision. For example, I'm thinking about icons that align with the default body text (16px) in Primer Brand components like Button, Label, and UnorderedList, among others. I'm not suggesting we necessarily use Icon in those, but those are potential use cases. Also custom elements might benefit from the expanded scale as there are several places in dotcom where use icons smaller than 20 and larger than 44.

danielguillan (approved):
Looks great! I left one comment about the numeric size scale, but it's non-blocking.

packages/react/src/Icon/Icon.tsx
danielguillan: Sounds good!

I believe there's only one icon larger than 44 in landing pages but [there are some more](https://github.com/search?q=org%3Agithub++%2F%3C.*Icon+.*size%3D%5C%7B%28%5B4-9%5D%5Cd%7C%5Cd%7B3%2C%7D%29%5C%7D%2F&type=code&p=1) on product pages that could potentially use Primer Brand components at some point (e.g., upsell). That said, we don't need to support those larger sizes if we want to discourage that. Octicons aren't actually designed for larger sizes either way. We can create some guardrails around that.

PR #798, opened 2024-10-25
Integrate Icon component into Bento and Card

joshfarrant:
## Summary

Replace custom Icon implementations in `Bento` and `Card` components with new `Icon` component.

## List of notable changes:

- Replace custom Icon implementations in `Bento` and `Card` components with new `Icon` component.
  - This also resolves a bug in the `Bento` component where the icon's default colour would be black, rather than the gray as it should have been. [See here](https://github.com/primer/brand/pull/798/files#diff-0f4f07cc6bd6d0b115bcd92a3b811badf30f2793abd4d1b46a43906c23bf8e95L70-L71) for the previous issue (variable should use `base` instead of `brand`).
- Update examples in docs to pass Icon component, rather than rendered element.
  - At one point I reference the not-yet-created Icon docs. These will be added in a subsequent PR.

- Happy with any visual changes?
  - Note that the increased spacing between icon/label and heading in the `Card` component is an intentional change to align with the Figma `Card`.

- Closes https://github.com/primer/brand/issues/773

## Contributor checklist:

## Reviewer checklist:

## Screenshots:

<table>
<tr>
</tr>
<tr>
<td valign="top">

 </td>
<td valign="top">

</td>
</tr>
</table>


apps/docs/content/components/Card/react.mdx
```diff
@@ -83,7 +83,7 @@ You can add an `icon` to enhance the visual context. We recommend using an [Octi
 
 ```jsx live
 <Card href="https://github.com">
-  <Card.Icon icon={<CopilotIcon />} color="purple" hasBackground />
+  <Card.Icon icon={CopilotIcon} color="purple" hasBackground />
```
rezrah, line 86: Does Icon still support passing the instantiated component, or only reference? I.e. if user wanted to customize values here, can they still do it?

packages/design-tokens/src/tokens/functional/components/card/colors.json
```diff
@@ -5,124 +5,6 @@
         "value": "var(--brand-color-canvas-default)",
         "dark": "var(--brand-color-canvas-subtle)"
       }
-    },
-    "icon": {
```
rezrah, line 9: We're missing a changeset in this PR, could we get one added please, and also capture in the Primitives one that we've removed these tokens?

packages/react/src/Bento/Bento.visual.spec.ts-snapshots/Visual-Comparison-Bento-Bento-Default-1-linux.png
rezrah, line 1: FYI the icon here seems to be muted now. Is that intended or a regression?

packages/react/src/Card/Card.features.stories.tsx
```diff
@@ -120,20 +121,21 @@ export const IconColors: StoryFn<typeof Card> = () => {
 export const WithIconSVG = () => (
   <Card href="https://github.com">
     <Card.Icon
-      icon={
+      icon={props => (
```
rezrah, line 124: Is `props` doing something specific here or can we keep it the same as before?

packages/react/src/Card/Card.module.css
```diff
@@ -111,21 +111,12 @@
 }
 
 .Card__icon {
-  margin-bottom: var(--base-size-16);
+  margin-bottom: var(--base-size-24);
```
rezrah, line 114: Was this extra spacing an intended design change? It's increasing the height of the card

packages/react/src/Card/Card.features.stories.tsx
```diff
@@ -148,7 +150,7 @@ WithIconSVG.storyName = 'Icon (native)'
 export const IconAndLabel: StoryFn<typeof Card> = () => {
   return (
     <Card href="https://github.com">
-      <Card.Icon icon={<GitBranchIcon />} color="purple" hasBackground />
+      <Card.Icon icon={GitBranchIcon} color="purple" hasBackground />
```
rezrah, line 153: FYI there's a noticeable visual difference now, with the overall icon dimensions being smaller. I _think_ it's a bit cleaner now tbh, but wanted to check this is something Site intended for? cc. @danielguillan

https://github.com/user-attachments/assets/d14b296c-9e53-477f-a17d-254b7a9d49f6

rezrah: Related, the reduced spacing here is different to how we show the same UI in the README project pages. E.g. 

Screenshot 2024-10-29 at 12 27 36

packages/react/src/Bento/Bento.visual.spec.ts-snapshots/Visual-Comparison-Bento-Bento-Default-1-linux.png
joshfarrant: I called this out in the PR description above, but tl;dr is that the icon shouldn't have been black before, it was a bug in the previous implementation

packages/design-tokens/src/tokens/functional/components/card/colors.json
joshfarrant: Good spot, thank you

packages/react/src/Bento/Bento.visual.spec.ts-snapshots/Visual-Comparison-Bento-Bento-Default-1-linux.png
rezrah: ah, okay thanks for confirming 🙇 (should have checked that 😓, my bad)

packages/react/src/Card/Card.module.css
joshfarrant: Yes intentional. Called this out in the PR description, but this aligns the component with the Figma

packages/react/src/Card/Card.features.stories.tsx
joshfarrant: `props` includes the classNames, without it the Icon styles won't be applied to the svg

apps/docs/content/components/Card/react.mdx
joshfarrant: No, that was removed by design after speaking with @danielguillan to avoid situations like this:

```tsx
<Card.Icon icon={<CopilotIcon size={20} />} size={44} />
```

Since `Card.Icon` forwards all of its props to the provided icon, you could easily end up in these awkward situations where two props clash. 

Do you think this should be called out as a breaking change in the changeset?

packages/react/src/Card/Card.features.stories.tsx
joshfarrant: Defer to @danielguillan on this. The behaviour in the PR matches the Figma, but I'd also like confirmation from Dani that we're happy with this change.

packages/react/src/Card/Card.module.css
danielguillan: It seems we've had this discrepancy between Figma and React for some time. The `24px` spacing has always been present in the Figma implementation (see interface guidelines images, for example) and was not introduced with the new Icon component update. 

I can confirm that 24 is the intended spacing here.

packages/react/src/Card/Card.features.stories.tsx
danielguillan: I agree, this doesn't look quite right. We should go with `12px` paddings instead of the current `8px` on sizes `20` and `24` on the Icon component. [Updated Figma](https://www.figma.com/design/BJ95AjraesmRCWsKA013GS/Primer-Brand?node-id=15444-214&node-type=frame&t=JKBjUZCPyhUCNrQv-11)

joshfarrant: Thanks @danielguillan — that's updated and I've regenerated the snapshots, let me know what you both think 
👍

danielguillan (approved):
Looking great! It just needs the adjustment to the default icon color we discussed in Slack.

apps/docs/content/components/Card/react.mdx
rezrah: Thanks for this example. Makes sense 👍 

This is a breaking change if we don't allow instantiated JSX anymore, yeah. Unless there's a really important reason for it, we should continue to support both ways, as we have _many_ instances in dotcom and elsewhere that use that approach and we can more easily work around it on our side than the users. In many of those instances, the user doesn't apply size too, so this is less of a risk IMO.

E.g.

Screenshot 2024-10-31 at 10 44 51

Suggestion:

Update our docs and examples to show the reference approach (happy path), and ensure that instantiated JSX can continue to be passed for continuity. How does that sound?

packages/react/src/Card/Card.features.stories.tsx
rezrah: I'm guessing this was needed because we only accepted function references? What would happen if someone forgot to spread the props?

For now, I've rolled this back to allow passing the SVG directly as a prop. We can figure out if the props argument is still needed when you're back.

apps/docs/content/components/Card/react.mdx
rezrah: Update for posterity.. I've pushed the changes ☝️ in [this commit](https://github.com/primer/brand/pull/798/commits/6acc73575c059c8da2e8f2b55a6027d6e5c50092)

- passing JSX is now working
- removed breaking change notice
- added a new story to track future regressions

rezrah (approved):

PR #822, opened 2024-11-18
Replace remote GitHub avatars with local assets

joshfarrant:
## Summary

- Removed remote usage of https://github.com/mona.png and replaced with local asset instead.
- Modified types in `IDE` component to remove unused `avatar` property from assistant message type
  - This allowed me to remove the unused assistant avatars from fixture data. They were set to https://github.com/copilot.png, which doesn't resolve anyway.

- Check I've not missed any instances of remote GitHub avatars
- Check that nothing has changed visually in the docs or Storybook

## Steps to test:

1.
1.
1.

-
-

## Contributor checklist:

## Reviewer checklist:


danielguillan (approved):
Thanks for fixing this!

packages/react/src/IDE/IDE.tsx
```diff
@@ -129,17 +129,25 @@ type IDEChatProps = {
...
-  highlighter?: 'hljs' // add additional highlighters as needed
+  highlighter?: 'hljs'
+}
+
+type IDEChatMessageAssistant = {
+  role: 'assistant'
+  handle: string
+  message: string
+  codeSnippet?: string
+  highlighter?: 'hljs'
```
rezrah, line 146: Nit... this seems a little duplicative. Took me a second to spot the difference . Worth creating an intersection with a shared message type for `handle`, `message`,  `codeSnippet` and `highlighter`?

packages/react/src/IDE/IDE.tsx
```diff
@@ -129,17 +129,25 @@ type IDEChatProps = {
...
 }
 
+export type IDEChatMessage = IDEChatMessageUser | IDEChatMessageAssistant
```
rezrah, line 149: We'll need a changeset for this update as it's a material change to the library. Although, it doesn't need to say too much. 

Maybe something like: Updated `IDEChatMessage` type declaration to differentiate between assistant and user messages. WDYT?

rezrah (approved):
Thanks @joshfarrant. [Left a comment about a missing changeset](https://github.com/primer/brand/pull/822#discussion_r1848129208), but otherwise LGTM. Will pre-emptively ✅ so you can merge in when ready.

packages/react/src/IDE/IDE.tsx
joshfarrant: Good shout, added 👍

packages/react/src/IDE/IDE.tsx
joshfarrant: Works for me, updated 🙂

joshfarrant:
Thanks @rezrah 🙌

Just made those changes, will merge once the checks have finished :shipit:

Issue #846, opened 2024-12-02
[a11y] Issue with `Button` and `Link` outlines on Windows 11 high contrast themes


stamat:
### 1. Issue with the `Button` component outline in high contrast themes

For some reason there is this code, that draws the outline if `@media (forced-colors: active)`. So on focus, the outline doesn't change, it's always visible.

https://github.com/primer/brand/blob/7473042c4ea4130f2d3fe9e4de88dd28899ce727/packages/react/src/Button/Button.module.css#L211-L216

This caused the issue 👇  to be reported:

- https://github.com/github/accessibility-audits/issues/9846
- https://github.com/github/accessibility-audits/issues/9849
- https://github.com/github/accessibility-audits/issues/9850

### 2. Issue with the `Link` component outline in high contrast themes

Link has the native outline disabled and `box-shadow` used as the outline for some reason unknown to me. In `@media (forced-colors: active)` all of the `box-shadows` are removed and all of the backgrounds are set to the same color.

https://github.com/primer/brand/blob/7473042c4ea4130f2d3fe9e4de88dd28899ce727/packages/react/src/Link/Link.module.css#L40-L44

Also Link needs the `outline-offset` set to `2px`

It's just something I picked up along the way.

**All high contrast issues are Sev 4** - so it's not a hurry on this. But could be cool for this to be sorted out and be consistent. 😘

stamat:
@rezrah @joshfarrant bros, I've made the update [here](https://github.com/primer/brand/pull/847), tested it out in the high contrast themes, it works now 💖 . Kinda need this in the new update 🙏

rezrah:
Closing as the fix has shipped in `v0.44.0`

PR #847, opened 2024-12-02
[a11y] `Button` and `Link` high contrast outline issue fix

stamat:
## Summary

Outline of the `Button` component was always visible in high contrast themes which was causing a11y issues.

Outline of the `Link` component was invisible in the high contrast themes which is also an a11y issue.

- Issue: https://github.com/primer/brand/issues/846

## List of notable changes:

- **removed** styles that were forcing the outline of the `Button` component to be always visible regardless of the focus in high contrast themes
- **updated** outline of the `Link` component to use the native outline instead of box-shadow, since box shadow is invisible in high contrast themes
- **updated** underline of the `Link` component to use border instead of background so the underline will be visible in high contrast themes

- Outline of the `Button` component should only be visible when the component is focused in high contrast themes
- Outline of the `Link` component should be visible when the component is focused in high contrast themes
- Underline of the `Link` component should be visible on hover in high contrast themes

## Steps to test:

1. `npm run start`
2. Make the storybook port public
3. Test in high contrast theme enabled Windows 11 by focusing on Button and Link with keyboard

- https://github.com/github/accessibility-audits/issues/9849
- https://github.com/github/accessibility-audits/issues/9850

## Contributor checklist:

## Reviewer checklist:

## Screenshots:

<table>
<tr>
</tr>
<tr>
<td valign="top">

Screenshot 2024-12-02 at 21 45 57

 </td>
<td valign="top">

Screenshot 2024-12-02 at 22 00 02

</td>
</tr>

<tr>
<td valign="top">

Screenshot 2024-12-02 at 21 53 45

 </td>
<td valign="top">

Screenshot 2024-12-02 at 21 58 36

</td>
</tr>

<tr>
<td valign="top">

Screenshot 2024-12-02 at 21 53 34

 </td>
<td valign="top">

Screenshot 2024-12-02 at 21 58 44

</td>
</tr>
</table>


joshfarrant:
Looks great @stamat, thank you!

Thanks for raising this, I'd not actually tested with Windows high-contrast themes before. It looks like our icons could do with switching to a different colour in some scenarios. Are you seeing the same thing that I'm seeing here (dark grey arrow icon is barely visible)?

image

Could you also add the _Update snapshots_ label to this PR to regenerate the snapshots please?

stamat:
@joshfarrant I'll take a deeper look into this man in few moments! I didn't know there was an icon lol...

joshfarrant:
Sorry @stamat, I wasn't suggesting this was something you had to fix 😅

I just wanted to ask out of curiosity as I've not tested with Windows high-contrast themes before and I wanted to make sure I wasn't missing something obvious. I'm happy to take a look through all our components at some point in the future and make sure they're working as well as possible in high-contrast.

joshfarrant (approved):

packages/react/src/Link/Link.module.css
```diff
@@ -18,10 +17,9 @@
   bottom: -0.15em;
   left: 0;
   width: calc(100% - var(--arrow-width));
-  height: 2px;
   pointer-events: none;
   content: '';
-  background-color: currentColor;
+  border: 1px solid currentColor;
```
rezrah, line 22: We have a css var for this:

```suggestion
  border: var(--brand-borderWidth-thin) solid currentColor;
```

packages/react/src/Link/Link.module.css
```diff
@@ -38,9 +36,8 @@
 }
 
 .Link:focus:focus-visible {
-  outline: none;
-  box-shadow: 0 0 0 2px var(--brand-color-canvas-default), 0 0 2px 5px var(--brand-color-focus);
-  transition: none;
+  outline: 2px solid var(--brand-color-focus);
+  outline-offset: 2px;
```
rezrah, line 40:
```suggestion
  outline: var(--brand-borderWidth-thick) solid var(--brand-color-focus);
  outline-offset: var(--base-size-2);
```

rezrah (commented):
Thanks for helping to fix this @stamat 💖 ... I've left you some minor feedback. Would love your thoughts [on this in particular](https://github.com/primer/brand/pull/847/files#r1867539330). 

A few other things before this lands:

- We'll need a changeset added to this PR to describe what changed in this PR for our release change log. You can do it by clicking this link on [this comment](https://github.com/primer/brand/pull/847#issuecomment-2512477235) 👇 
  
  Screenshot 2024-12-03 at 11 29 39

- Updated snapshots would be useful to verify there aren't any visual regressions. You can add the 'update snapshots' label in this PR to run the tests.

packages/react/src/Link/Link.module.css
stamat: "There is an app for that" pun huh? 😆

packages/react/src/Button/Button.module.css
```diff
@@ -197,24 +197,10 @@
 }
 
 @media (forced-colors: active) {
-  .Button--primary,
-  .Button--secondary,
-  .Button--disabled:hover {
-    outline: var(--brand-borderWidth-thick) solid transparent;
-    outline-offset: 2px;
-  }
-
```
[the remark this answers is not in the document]
stamat: @rezrah I completely failed to understand how these button updates that I removed helped accessibility, seemed they had an opposite effect based on my understanding... Yeah it's a good call to ask @TylerJDev to shed some light on this. I'm all 👂 ! 🙏 ❤️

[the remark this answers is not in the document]
stamat: @rezrah I think I know what happened... Our old(er) buttons probably didn't have a border but used an inset box-shadow... https://github.com/primer/brand/blob/d763e2fb6584489f1277370457b4c2ebded2fd82/packages/react/src/Button/Button.module.css (This is the state from the era the change was made)

So this "fix" used the outline to draw the border in high contrast themes since backgrounds and shadows are disabled - instead of just adding the border!?

stamat:
Also a note, I believe `aria-disabled="true"` is assumed if the button has `disabled` attribute: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-disabled

>When needing to disable native HTML form controls, developers will need to specify the disabled attribute, as it provides all of the generally expected features of disabling a control by default. However, there can be instances where elements need to be exposed as disabled, but are still available for users to find when navigating via the Tab key. Doing so can improve their discoverability as they will not be removed from the focus order of the web page, as aria-disabled does not change the focusability of such elements, nor will the elements be dimmed by default browser styling, making them easier to read.

I don't think this applies to buttons, for the buttons to be disabled we need to use `disabled` attr...

stamat:
@joshfarrant @rezrah one final ✅ from you 🙏 ! Also should I up a patch version? 😇

packages/react/src/Button/Button.module.css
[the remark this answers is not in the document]
TylerJDev: Hello 👋 

Looking at the change, and my own change, this was a time before borders were used for buttons, so we needed to add something to imitate the styles of box-shadow for HCM. 

Since we've added borders to the button, we don't need to imitate the borders anymore, since those always show in HCM, so removing this is okay since it, and the focus styles. Right now it seems like the two are competing against each other, the outline provided in HCM and the border of the button.

As for the outline, this was more or less interchangeable with the border in HCM. I'm not too sure if there was another driving factor here, since it's been a minute 😅, but having there be some sort of presence of a border was the main goal to differentiate between what is/isn't a button.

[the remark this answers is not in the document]
stamat: @TylerJDev thanks for the reply dude! 💖 Yeah I figured it was something like that! ☺️

packages/react/src/Hero/Hero.visual.spec.ts-snapshots/Visual-Comparison-Hero-Mobile-viewport-test-for-Narrow-View-Hero-Narrow-View-1-linux.png
rezrah, line 1: @stamat heads up that there's a visual regression on the arrow here, which appears blue instead of black. Can repro this in the preview URL here: https://primer-a794edc343-26139705.drafts.github.io/brand/storybook/?path=/story/components-button-features--secondary-with-hover-interaction

Looking into why now (unless you already know?), and will also add a changeset for you in this PR.

packages/react/src/Button/Button.module.css
```diff
@@ -118,11 +117,6 @@
   color: var(--brand-button-secondary-fgColor-rest);
 }
 
-.Button--secondary .Button__icon-visual,
-.Button--secondary .Button-arrow {
-  color: var(--brand-button-secondary-fgColor-rest);
-}
-
```
rezrah, line 125: @stamat do we need to remove these styles? Doesn't seem related to HCM (might be wrong?) It's caused a visual regression whereby the arrow is now blue:

Screenshot 2024-12-04 at 12 43 50

.changeset/itchy-apes-peel.md
```diff
@@ -0,0 +1,5 @@
+---
+'@primer/react-brand': patch
+---
+
+Improved accessibility of `Button` and `Link` components on Windows-based high contrast themes. Outlines are now visible only when focused, and border underlines will appear correctly.
```
rezrah, line 5: @stamat I've added a changelog entry for this update ☝️, which hopefully summarizes what these changes do. Could you please check its accurate on your end, and that I haven't missed anything?

rezrah (approved):
This is looking great @stamat, thanks for making these changes.

FYI, I've pushed a few changes to get this PR mergeable:

- Added a changeset for the changelog entry
- Reset your visual snapshots, as they were showing false positives

stamat:
@rezrah thanks for the your help and extra eyes! 💖 Everything is good with the changeset! I'm merging it now!

Issue #829, opened 2024-11-21
[Feature request] Add "boxed" variant to the Testimonial


seangolob:
Hi PB Team!

I have noticed a few pages that have testimonials that have a boxed wrapper around them:

Single Testimonial:

https://github.com/features/actions
https://github.com/features/issues

![Image](https://github.com/user-attachments/assets/589d9a45-1051-46f4-873d-f17d42fc8511)
![Image](https://github.com/user-attachments/assets/d5542fa3-6f0a-4610-98fc-5f7f6c314d12)

Two Testimonials:

https://github.com/features/codespaces
https://github.com/features/code-search

![Image](https://github.com/user-attachments/assets/297b860a-2b73-4fab-9088-f3ddeb8b9cd3)
![Image](https://github.com/user-attachments/assets/7a061e08-0e04-40a0-b7e6-d7f2bb3e8872)

Can we consider adding a "boxed" variant to the Testimonial component to support these use cases? 

The Single Testimonial page examples have a somewhat transparent background with a blur which should probably exist at the implementation level. However, it would be nice to have the spacing defined in the PB library to have future testimonials share the same spacing.

CC @simmonsjenna CC @jesussandreas

Thank you for your help!

rezrah:
Thanks for opening this feature request @seangolo ❤ 

A couple of options here:

1. Add a Storybook example for your team to reference as the design can be achieved using the `Box` component today
2. Add a `variant` prop to the `Testimonial` component to add first-class support

My preference would be the latter (option 2) assuming Site Design plan to continue using this design approach in future.

There's a _lot_ of detail here that would be difficult to expect users to implement on their own each time. E.g. responsive paddings, responsive filter effects, etc.

I think it's reasonable to add first-class support to Testimonial for this using the variant prop.

E.g. of new API

```typescript
type TesimonialProps {
  variant?: 'default' | 'minimal' | 'frost'
}
```

Like the Card component, 'minimal' would be the correspond to the current version of the Testimonial and be made default to avoid a breaking change. 'default' value would be inclusive of a background color and padded, and frost being the same with a background filter applied.

Final API will likely differ, but I'll bring this to other maintainers and provide an update here when that's available.

seangolob:
@rezrah Thank you for your feedback! ❤ 

Depending on the priority of this issue for the SWP team, we may build this internally and potentially port over to the PB API if/when it becomes available.

[...]

PR #863, opened 2024-12-13
Add Testimonial variants and FrostedGlassVFX component

rezrah:
## Summary

🔗 [Preview updated documentation for Testimonial](https://primer-7be784b7d2-26139705.drafts.github.io/brand/components/Testimonial/react#variants)
🔗 [Preview Storybook examples for new Testimonial variants](https://primer-7be784b7d2-26139705.drafts.github.io/brand/storybook/?path=/story/components-testimonial-features--testimonial-variants)
🔗 [Preview frosted glass effect with Testimonial](https://primer-7be784b7d2-26139705.drafts.github.io/brand/storybook/?path=/story/components-testimonial-examples--with-frosted-glass)

Part of https://github.com/primer/brand/issues/829

## List of notable changes:

- Added new, experimental FrostedGlassVFX component - no docs at this time
- Added default and subtle variants to Testimonial
- Refactored Testimonial stories into /features and /example subfolders for consistency 
- Added Pillar examples using frost effect

- Reviewing border colors on new variants
- Reviewing padding and spacing
- Are we happy to add three intensity levels to FrostGlassVFX?

## Steps to test:

1. Use the links at the top 
1. Review the code in files diffed

-[ Issues page prior art](https://github.com/features/issues)

## Contributor checklist:

## Reviewer checklist:

## Screenshots:

Visual-Comparison-Testimonial-Testimonial-Testimonial-Variants-1-linux

Visual-Comparison-Testimonial-Testimonial-With-Frosted-Glass-1-linux


apps/docs/content/components/Testimonial/index.mdx
joshfarrant, line 1: I might be missing something here, but have these docs intentionally been deleted? 😅

packages/react/src/Testimonial/Testimonial.features.stories.tsx
```diff
@@ -0,0 +1,168 @@
...
+          first line of code we&apos;re writing.
+        </Testimonial.Quote>
+        <Testimonial.Name position="Staff Security Engineer">David Ross</Testimonial.Name>
+        <Testimonial.Avatar src={monaAvatar} alt="Circular avatar from David Ross's GitHub profile" />
+      </Testimonial>
+    </div>
+  ),
+}
+
+export const TestimonialVariants: Story = {
```
joshfarrant, line 84: Worth adding `variant` to the Playground story do you think?

packages/react/src/Pillar/Pillar.features.stories.tsx
```diff
@@ -199,3 +199,48 @@ export const StackedWithLink: StoryFn<typeof Pillar> = () => {
     </Stack>
   )
 }
+
+export const FrostedGlassEffect: StoryFn<typeof Pillar> = () => {
```
joshfarrant, line 203: The colour contrast of the grey description text in this example drops quite low in places (2.9:1 on the left-most Pillar when over the dark-blue).

Of course this is only an example of component usage, but I think it might still be worth thinking about.

A few of the a11y issues that were raised in the recent audit related to inaccessible examples in our docs and storybook. Essentially, the goal is that all of our examples should promote accessibility best-practises since consumers may use our examples to inform their own designs and decisions. I'm not suggesting that anyone will copy this particular background image (hopefully!) but I do think that using a background image which doesn't result in areas of low contrast would allow us to lead by example.

The darker the background is, the lower the contrast drops, so maybe a vibrant, light image would be a better choice.

wdyt?

joshfarrant (commented):
Looking good!! ❄️🪟

I know you touched on this in the PR description, but I think that this component would benefit from having some documentation available right as it's released, even if it's really lightweight, as we need to be careful with to not unintentionally promote inaccessible experiences.

My feeling is that we’d benefit from some clear guidance on how to ensure that the component is being used in an accessible way since the contrast of the text depends so heavily on the colour behind it. In a worst-case scenario, using a pillar over a black background results in a contrast of 2.2:1, which falls quite a way short of the minimum 4.5:1 contrast that's required.

The tricky thing here is that this component is kind of an accessibility footgun; the nature of it is such that we can't completely prevent consumers from being able to use it in an inaccessible way. We can, however, help them understand what sort of things they need to consider and test when using this component, eg testing colour contrast of text across the whole background and across all viewports.

Even basic documentation would be super helpful here, but providing guidance on how to use the component in an accessible way feels quite important.

What are your thoughts?

natalie-iv:
@rezrah the first docs link isn’t working—unsure if that’s intentional.

The borders don’t look right to me. I’m using the features pages as a reference—it seems to reflect our latest component designs (though I’m not sure if it’s the official source).:

1. FrostGlass border does not align with the one we use on [features/issues page](https://github.com/features/issues). Here is what was used in [Figma](https://www.figma.com/design/SSBWMP19m11Fi3tQ7eYHE7/Features?node-id=1216-30072&t=iRMRKl976TGWc5Pq-1).
Screenshot 2024-12-16 at 11 14 02

2. Border of light version of testimonial does not align with  what we used [in Figma](https://www.figma.com/design/SSBWMP19m11Fi3tQ7eYHE7/Features?node-id=1039-22599&t=iRMRKl976TGWc5Pq-1). 

Screenshot 2024-12-16 at 11 13 43

However on the life website there **are no borders** at all for light testimonial cards (ex. https://github.com/features/code-search). @jesussandreas:  Was this our agreed general approach for all light versions of testimonials, or is our general direction with a border as in Reza's previews or in Figma?

Screenshot 2024-12-16 at 10 56 09

apps/docs/content/components/Testimonial/index.mdx
rezrah: 😱 oh no, that's definitely not intentional. restored. Thanks for spotting that.

packages/react/src/Pillar/Pillar.features.stories.tsx
rezrah: Yeah, this is a fair point. That background is a stress test for internal use (VRT). It's not intended to be  end-user documentation. I'm wondering if we should have another 'subfolder' called tests / experiments, which auditors know to ignore. Ideally they wouldn't look at Storybook at all, as the public docs is our store front.

rezrah: Separately to the above, I've updated the pillar example now to use the same animated background as testimonial. It achieves the same thing from a VRT perspective and is more consistent with how we expect end-users to utilize the component.

packages/react/src/FrostedGlassVFX/FrostedGlassVFX.tsx
```diff
@@ -0,0 +1,54 @@
...
+export const FrostedGlassVFXIntensity = ['low', 'medium', 'high'] as const
+export type FrostedGlassVFXIntensity = (typeof FrostedGlassVFXIntensity)[number]
+export const defaultFrostedGlassVFXIntensity: FrostedGlassVFXIntensity = 'medium'
+
+export type FrostedGlassVFXProps = BaseProps<HTMLDivElement> &
+  React.HTMLAttributes<HTMLDivElement> & {
+    /**
+     * Applies rounded corners
+     */
+    roundedCorners?: boolean
```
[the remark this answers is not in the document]
rezrah: 👍 Agreed on inconsistency. For posterity, we don't want to support _n_ string values here, as the configurability of corners is binary. Updated to be `hasBorderRadius` for consistency with other existing component API's.

rezrah:
Update for frosted glass effect:

- Borders now always present
- Exclusion mask used to enable a gradient
- Light mode gradient matches Figma file
- Dark mode copied from production

preview links:

- [Testimonial (light mode)](https://primer-7be784b7d2-26139705.drafts.github.io/brand/storybook/?path=/story/components-testimonial-examples--with-frosted-glass)
- [Testimonial (dark mode)](https://primer-7be784b7d2-26139705.drafts.github.io/brand/storybook/?path=/story/components-testimonial-examples--with-frosted-glass-dark)
- [Pillars](https://primer-7be784b7d2-26139705.drafts.github.io/brand/storybook/?path=/story/components-pillar-features--frosted-glass-effect)

packages/react/src/FrostedGlassVFX/FrostedGlassVFX.module.css
```diff
@@ -0,0 +1,63 @@
...
+  border-radius: var(--brand-borderRadius-xlarge);
+}
+
+.FrostedGlassVFX > * {
+  background-color: transparent;
+}
+
+.FrostedGlassVFX > *::before {
+  border-color: transparent;
+}
```
danielguillan, line 63:
```suggestion
.FrostedGlassVFX > *
.FrostedGlassVFX > *::before {
  border-color: transparent;
}
```

We may also want to override the border if it is set in the element itself rather than in the `::before` pseudo-element.

danielguillan (approved):
LGTM! 🚀

.changeset/smooth-ways-retire.md
```diff
@@ -0,0 +1,24 @@
...
+      GitHub helps us ensure that we have our security controls baked into our pipelines all the way from the first line
+      of code we&apos;re writing.
+    </Testimonial.Quote>
+    <Testimonial.Name />
+    <Testimonial.Avatar />
+  </Testimonial>
+</FrostedGlassVFX>
+```
+
+🔗 (See Storybook for an example)[https://primer.style/brand/storybook/?path=/story/components-testimonial-examples--with-frosted-glass]
```
joshfarrant, line 24:
```suggestion
🔗 (See Storybook for an example)[https://primer.style/brand/storybook/?path=/story/components-testimonial-examples--with-frosted-glass]

⚠️  When using this component, ensure sufficient contrast between text and the background behind `FrostedGlassVFX`. The frosted glass effect can reduce contrast, especially:
- In light mode: Dark background areas may make dark text hard to read
- In dark mode: Light background areas may make light text hard to read

Test contrast across all viewport sizes when using `FrostedGlassVFX` over varied or multicoloured backgrounds.
```

packages/react/src/FrostedGlassVFX/FrostedGlassVFX.module.css
```diff
@@ -0,0 +1,63 @@
...
+    content: '';
+    background-color: var(--brand-color-canvas-default);
+    z-index: 0;
+    border-radius: var(--brand-borderRadius-xlarge);
+  }
+
+  .FrostedGlassVFX > * {
+    z-index: 1;
+  }
+}
```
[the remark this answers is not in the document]
joshfarrant: +1, support looks really good. Also, it looks like the only browsers that don't support `backdrop-filter` also don't support `@supports`, so agree that this can probably be safely removed.

- [Can I use backdrop-filter](https://caniuse.com/css-backdrop-filter)
- [Can I use @supports](https://caniuse.com/?search=@supports)

packages/react/src/Testimonial/Testimonial.features.stories.tsx
danielguillan, line 84: Can we name this story "Variants" for simplicity and consistency? I believe we generally do not repeat the name of the component in the feature stories.

rezrah: Yeah will do, not sure why I did it that way tbh 😕

packages/react/src/FrostedGlassVFX/FrostedGlassVFX.module.css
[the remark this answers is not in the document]
rezrah: Yeah that's fair. Examples of backdrop-filter seemed to include it but we don't do this in dotcom either, so removing it. Thanks 👍

.changeset/smooth-ways-retire.md
rezrah: Thanks @joshfarrant - I've adapted your suggestion and put it earlier in the log entry.

PR #882, opened 2025-01-08
Add audio description example to VideoPlayer docs

joshfarrant:
## Summary

Adds an audio description example to the VideoPlayer documentation, and fixes styling of custom play icon example.

## List of notable changes:

- Added an audio description example to the VideoPlayer documentation
- Changed the fill of the custom play icon in the VideoPlayer documentation

- Make sure you're happy with the example and phrasing.

## Steps to test:

1. Take a look at the updated docs

- Closes https://github.com/github/primer/issues/3771

## Contributor checklist:

## Reviewer checklist:


apps/docs/content/components/VideoPlayer.mdx
```diff
@@ -22,6 +23,22 @@ import {VideoPlayer} from '@primer/react-brand'
...
+The audio description can be provided as part of the existing soundtrack, or via a link to an alternate audio-described version.
+
+```jsx live
+<Stack gap="normal">
+  <VideoPlayer title="GitHub media player">
+    <VideoPlayer.Source src="/brand/assets/example.mp4" />
+    <VideoPlayer.Track src="/brand/assets/example.vtt" default />
+  </VideoPlayer>
+  <Link href="#">Watch with audio description</Link>
+</Stack>
```
rezrah, line 39: Screenshot 2025-01-08 at 12 49 55


The link hover state is taking up the full width of the container. Quick fix below 👇 


```suggestion
<Stack gap="normal" alignItems="center">
  <VideoPlayer title="GitHub media player">
    <VideoPlayer.Source src="/brand/assets/example.mp4" />
    <VideoPlayer.Track src="/brand/assets/example.vtt" default />
  </VideoPlayer>
  <Link href="#">Watch with audio description</Link>
</Stack>
```

rezrah: Also could this be a button instead of a link?

apps/docs/content/components/VideoPlayer.mdx
```diff
@@ -111,7 +128,7 @@ The `VideoPlayer` component exposes a `useVideo` hook that can be used to contro
 Full documentation for the `useVideo` hook can be found [below](#usevideo-context).
 
 ```tsx live
-() => {
+;() => {
```
rezrah, line 131: Looks like a typo? Could we revert this? 

Screenshot 2025-01-08 at 12 53 18

joshfarrant: Good spot, thanks. My IDE likes to put a semicolon there on save for some reason.

I've updated the example to use the `javascript live noinline` tags and refactored it slightly to prevent this happening again.

apps/docs/content/components/VideoPlayer.mdx
joshfarrant: The intention here is to demonstrate the second pattern described above.

> or via a link to an alternate audio-described version

As we don't have an audio-described version of this video I opted to just use a href of "#". If there's another approach you'd prefer just let me know.

Good spot on the width issue too, that's now fixed.

rezrah: Gotcha, thanks @joshfarrant. Could we instead use a no-op link, or open an alert saying something like "Audio description should begin playing" (Assuming that's what's supposed to happen?). I think the scrolling is a quite distracting right now, what do you think? 

I've not seen this audio description link on any of our marketing videos, so wonder how it's supposed to look. We should ask Site at some point for guidance on whether it should be button or link. Until then it's fine to keep as a link.

joshfarrant: Happy to update it, but just want to make sure we're on the same page about the intended behaviour here. 

The intention is that the text "Watch with audio description" will link to an alternative video on some other page; that video will have audio description embedded in it. That's why this can only be a link and not a button.

So if we're keeping it as a link then should we just keep it as `href="#"` since we do that in quite a few other places in the docs already?

image

I've asked around and I can't find any self-hosted videos which have an optional audio description, so there might not be an existing pattern for this. My goal here is to take the simplest possible approach and just ask consumers to provide a link to an audio described version.

Hopefully that makes sense

rezrah: Thanks Josh, makes sense to keep it a link. How about we add an external link icon next to it then, to make it obvious _before_ someone clicks on it, that it should eventually go to an external location. Otherwise it still feels both a little broken with the scrolling and confusing on should happen.  

```jsx
<Stack gap="normal" alignItems="center">
  <VideoPlayer title="GitHub media player">
    <VideoPlayer.Source src="/brand/assets/example.mp4" />
    <VideoPlayer.Track src="/brand/assets/example.vtt" default />
  </VideoPlayer>
  <Stack direction="horizontal" gap={12} padding="none" alignItems="center">
    <Link href="#" arrowDirection="none">Watch with audio description</Link>
    <LinkExternalIcon size={16} />
  </Stack>
</Stack>
```

Should produce:
Screenshot 2025-01-09 at 10 16 17

joshfarrant: Works for me, thanks!

joshfarrant: Just spotted a style issue on hover again, let me fix that and I'll ping you when it's ready

rezrah (approved):

PR #889, opened 2025-01-13
Add `green-blue-purple` gradient color variant to `Label` component

danielguillan:
## Summary

Solves https://github.com/github/primer/issues/4583

## List of notable changes:

- New `green-blue-purple` option available for Label `color`.

- Gradient matches custom implementation in https://github.com/features/copilot
- See comment about custom colors in https://github.com/github/primer/issues/4583#issuecomment-2586989493

## Steps to test:

- [Storybook](https://primer-d99d6b9082-26139705.drafts.github.io/brand/storybook/?path=/story/components-label-features--color)

## Contributor checklist:

## Reviewer checklist:


packages/react/src/IDE/IDE.visual.spec.ts-snapshots/Visual-Comparison-IDE-IDE-Playground-1-linux.png
joshfarrant, line 1: Strange — any idea why the `JS` icon has disappeared from the file tab?

packages/react/src/Label/Label.module.css
```diff
@@ -1,10 +1,24 @@
 .Label {
-  --brand-Label-gradient: radial-gradient(
```
joshfarrant, line 2: I was going to ask whether you think we should continue to support this CSS variable, but I've had a look in github/github and I can't see it used, so it's probably alright to remove.

Might be worth calling it out in the changelog though, wdyt?

packages/react/src/IDE/IDE.visual.spec.ts-snapshots/Visual-Comparison-IDE-IDE-Playground-1-linux.png
joshfarrant: Ah interesting, it looks like the SVG takes a second to load and the snapshot is taken before it does.

Rather than dig into that, you can probably just add a short timeout (idk what value, maybe like 1000ms or 1500ms?) to this snapshot here:

https://github.com/primer/brand/blob/a5b6800a8316c1a26732b678e91b6d2def5445e1/packages/e2e/scripts/playwright/playwright.generate-tests.ts#L23-L100

joshfarrant (approved):
🔥

packages/react/src/IDE/IDE.visual.spec.ts-snapshots/Visual-Comparison-IDE-IDE-Playground-1-linux.png
danielguillan: Ah! Thanks for looking into this, @joshfarrant! Regenerating the snapshots now.

packages/react/src/Label/Label.module.css
danielguillan: Hmm, these CSS variables are actually internal to the component and not meant for external use. Perhaps we should consider a different naming convention for them?

joshfarrant: Ah gotcha, disregard this suggestion in that case 🙂

PR #896, opened 2025-01-21
Baseline repo to React 18 ahead of 19 upgrade

rezrah:
…each/auto-id

## List of notable changes:

-
-

-
-

## Steps to test:

1.
1.
1.

-
-

## Contributor checklist:

## Reviewer checklist:

## Screenshots:

<table>
<tr>
</tr>
<tr>
<td valign="top">

 </td>
<td valign="top">

</td>
</tr>
</table>


packages/react/src/Button/Button.test.tsx
```diff
@@ -63,44 +63,47 @@ describe('Button', () => {
...
-    const {getByRole} = render(
+    render(
       // eslint-disable-next-line no-console
       <Button variant="primary" size="medium" aria-disabled="false" onFocus={() => console.log('Focused')}>
         Primary Button
       </Button>,
     )
-    const btnEl = getByRole('button')
-    btnEl.focus()
+    await act(async () => {
```
joshfarrant, line 72: Happy for this to be addressed separately, but you _shouldn't_ need `act` here, curious to know what happens if we omit it.

Another [recommendation](https://testing-library.com/docs/user-event/intro/#writing-tests-with-userevent) with user-events v14 is that we switch to using `userEvent.setup()` rather than having it be called implicitly when we call click/hover/tab/etc

joshfarrant (approved):
Awesome!

packages/react/src/Button/Button.test.tsx
rezrah: pretty sure i needed to add this because we have [state being set on mouseover](https://github.com/primer/brand/blob/main/packages/react/src/Button/Button.tsx#L123)... it was producing a lot of linter warnings about adding act.

PR #898, opened 2025-01-21
Allow inline ControlGroup examples to reflow

joshfarrant:
## Summary

Updates the inline `CheckboxGroup` and `RadioGroup` examples to reflow on narrow viewports.

There's quite a large row gap between lines. This could be solved by adding `style={{ rowGap: 'var(--base-size-8)' }}` to the `Stack` used in the example. Let me know if that would be preferred.

- Closes https://github.com/github/primer/issues/4672
- Closes https://github.com/github/primer/issues/4673

## Screenshots:

<table>
<tr>
</tr>
<tr>
<td valign="top">

image

 </td>
<td valign="top">

image

</td>
</tr>
</table>

Having to add an inline style to the example isn't ideal, but without it the rowGap is way too big

image

### DIFF
--- a/apps/docs/content/components/CheckboxGroup/react.mdx
+++ b/apps/docs/content/components/CheckboxGroup/react.mdx
@@ -133,7 +133,13 @@ When space is limited, checkboxes can be arranged horizontally using the [Stack]
   <CheckboxGroup.Caption>
     Some inline checkboxes with a visually hidden label
   </CheckboxGroup.Caption>
-  <Stack direction="horizontal" gap="normal" padding="none" flexWrap="wrap">
+  <Stack
+    direction="horizontal"
+    gap="normal"
+    padding="none"
+    flexWrap="wrap"
+    style={{rowGap: 'var(--base-size-8)'}}
+  >
     <FormControl>
       <FormControl.Label>Choice one</FormControl.Label>
       <Checkbox value="one" />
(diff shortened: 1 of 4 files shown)


apps/docs/content/components/CheckboxGroup/react.mdx
```diff
@@ -133,7 +133,13 @@ When space is limited, checkboxes can be arranged horizontally using the [Stack]
   <CheckboxGroup.Caption>
     Some inline checkboxes with a visually hidden label
   </CheckboxGroup.Caption>
-  <Stack direction="horizontal" gap="normal" padding="none">
+  <Stack
+    direction="horizontal"
+    gap="normal"
+    padding="none"
+    flexWrap="wrap"
+    style={{rowGap: 'var(--base-size-8)'}}
```
rezrah, line 141: I think this is fine, but also quite ugly for our public docs. Could we remove it for now to keep it clean? We rarely have folks viewing docs on mobile, and there's bigger issues with that viewport generally with our docs.

rezrah:
```suggestion
```

joshfarrant: No prob, I guessed you might say that so I added that change in an easily-revertible commit 😄

That's been removed now

PR #912, opened 2025-01-29
ActionMenu falsy value bugfix

### LINKED ISSUE #906: 🐛 [BUG] - ActionMenu ignores items whose value is falsy
### Describe the bug

In the `ActionMenu` component, selection is ignored if the item's value is falsy (including here where it's intentionally `0`). I understand we might have wanted [this check](https://github.com/primer/brand/blob/main/packages/react/src/ActionMenu/ActionMenu.tsx#L169) to filter out invalid items so, if kept, maybe checking `newValue` against `undefined` instead would do? We might intentionally want to use `null` or `false` as values.

This is currently non-blocking for my use case since I'm working with numeric values (see ugly temp workaround below), but it might be different for a user who needs to use `null` or `false` for example. Thanks.

```
onToggle(String(+newValue - 1))

...

<ActionMenu.Item value={item.value + 1} ... />
```

### Reproduction steps

```bash
1. Setup an `ActionMenu` component with two items: one whose value is falsy (including but not limited to `undefined`), the other not.
2. Open the browser and attempt to select the first item in the menu
3. Observe that it cannot be selected
```

### Expected behavior

```bash
`ActionMenu` items should be selectable even if they have a falsy value, unless it is specifically `undefined`.
```

### Screenshots

https://github.com/user-attachments/assets/468ea74b-143f-45f4-9369-e3840dbe222b

### Browsers

Chrome

### OS

Mac

joshfarrant:
## Summary

Removes check for truthiness in `handleItemSelection` in `ActionMenu.tsx` to allow falsy values to trigger item selection callback.

## List of notable changes:

- `ActionMenu.tsx`: Removed check for truthiness in `handleItemSelection` to allow falsy values to trigger item selection callback.
- `ActionMenu.test.tsx`: Added a test to assert that falsy items can be selected

## Steps to test:

1. Open [the ActionMenu Docs](https://primer-f49b2ea768-26139705.drafts.github.io/brand/components/ActionMenu)
2. Edit the first example to set one of the values to `""`
3. Open the ActionMenu and click on the item associated with that value
4. Observe that the `onSelect` callback is called and triggers an alert.

- 🔗 [Link to docs](https://primer-f49b2ea768-26139705.drafts.github.io/brand/components/ActionMenu)
- Closes #906 

## Contributor checklist:

## Reviewer checklist:

### DIFF
--- a/packages/react/src/ActionMenu/ActionMenu.tsx
+++ b/packages/react/src/ActionMenu/ActionMenu.tsx
@@ -166,11 +166,9 @@ const _ActionMenuRoot = memo(
 
     const handleItemSelection = useCallback(
       (newValue: string) => {
-        if (newValue) {
-          handleOnSelect(newValue)
-          toggleMenu()
-          anchorElementRef.current?.focus()
-        }
+        handleOnSelect(newValue)
+        toggleMenu()
+        anchorElementRef.current?.focus()
       },
       [handleOnSelect, toggleMenu, anchorElementRef],
     )
--- a/.changeset/kind-years-repair.md
+++ b/.changeset/kind-years-repair.md
@@ -1,0 +1,5 @@
+---
+'@primer/react-brand': patch
+---
+
+Fixed a bug in the `ActionMenu` component where items with falsy values (eg `""`) would not trigger the `onSelect` callback when selected.
--- a/packages/react/src/ActionMenu/ActionMenu.test.tsx
+++ b/packages/react/src/ActionMenu/ActionMenu.test.tsx
@@ -349,4 +349,23 @@ describe('ActionMenu', () => {
       {timeout: 100},
     )
   })
+
+  it('should allow falsey items to be selected', () => {
+    const mockOnSelect = jest.fn()
+
+    const {getByRole} = render(
+      <ActionMenu onSelect={mockOnSelect} selectionVariant="single">
+        <ActionMenu.Button>Open menu</ActionMenu.Button>
+        <ActionMenu.Overlay aria-label="Actions">
+          <ActionMenu.Item value="test">Test string</ActionMenu.Item>
+          <ActionMenu.Item value="">Empty string</ActionMenu.Item>
+        </ActionMenu.Overlay>
+      </ActionMenu>,
+    )
+
+    fireEvent.click(getByRole('button', {name: 'Open menu'}))
+    fireEvent.click(getByRole('menuitemradio', {name: 'Empty string'}))
+
+    expect(mockOnSelect).toHaveBeenCalledWith('')
+  })
 })


packages/react/src/ActionMenu/ActionMenu.tsx
```diff
@@ -166,11 +166,9 @@ const _ActionMenuRoot = memo(
...
     const handleItemSelection = useCallback(
       (newValue: string) => {
-        if (newValue) {
-          handleOnSelect(newValue)
-          toggleMenu()
-          anchorElementRef.current?.focus()
-        }
+        handleOnSelect(newValue)
+        toggleMenu()
+        anchorElementRef.current?.focus()
```
rezrah, line 171: Do we need to remove the guard entirely if we want to only allow "" values to pass through? Could we continue to guard against other falsy values?

Something like:

```suggestion
        if (newValue || newValue === "") {
          handleOnSelect(newValue)
          toggleMenu()
          anchorElementRef.current?.focus()
        }
```

joshfarrant: It's a good spot, and I did have that at first, however I noticed that it was a bit redundant as `newValue`'s type is a (non-optional) string, so the only scenario in which it can ever be falsy is when it equals `""`. I didn't want to add redundant checks in that TypeScript already covered for us, so I ended up with this.

joshfarrant: > Curious what the use-cases are for allowing all falsy values through otherwise.

In reality I'm not sure there's ever a situation where a falsy value is the only way to accomplish something, but the current behaviour does feel a bit counterintuitive.

More context on the specific use-case that prompted this here https://github.com/primer/brand/issues/906

rezrah: Thanks, I've just looked and the original issue and wondering if this is even feasible.

We can't allow falsy values because the selected text is needed for the button label. This is why the guard is likely there to begin with. 

https://github.com/user-attachments/assets/34e67ec3-7eef-46d8-b20c-ff07ca8b1876

joshfarrant: From what I can tell, the `ActionMenu` doesn't manage the Button label. The label gets passed in as a child to `<ActionMenu.Button>` so it's up to the implementor to define. In many of our Storybook and Docs examples we have added some state to the rendered example to implement the behaviour shown in your video above, but this functionality is not actually a part of the component itself.

Assuming `ActionMenu` is intended to function in the same way as an HTML `<select>` element, then we should treat a value of `""` in the same way as any other string since `<select>` doesn't discriminate between `""` and any other string. In the case of `<select>`, using an empty string as a value doesn't seem to be an uncommon pattern.

rezrah: Ah yes, just saw that thanks. That's fine then. I guess it's okay to make this change, but I still can't really see the use-case for allowing null or undefined values through like this. I guess it means that the user has to mitigate those scenario's where data is from unreliable sources themselves?

joshfarrant: Yea potentially.

Equally we could approach the question from the other side and ask if there's a good reason that we _should_ impose restrictions on the strings that are passed. Maybe there's no situation where passing an empty value is the _only_ way to make something work, but it seems reasonable that we offer that flexibility to consumers in case that's the approach they want to take.

Just to confirm too; null or undefined values will be flagged by TypeScript, the only behaviour that is modified by this PR is specifically the `value=""` scenario.

rezrah: I see 🤔 PB does need to work in non-TS contexts though that don't have type inference. Want to make sure we're not making assumptions about that. But regardless, I agree with:

> it seems reasonable that we offer that flexibility to consumers in case that's the approach they want to take

👍

rezrah (approved):

PR #915, opened 2025-01-30
Add interface guidelines for Breadcrumbs component

danielguillan:
## Summary

Adds new interface guidelines for the Breadcrumbs component.

- 📚[Docs preview](https://primer-af43a6f01c-26139705.drafts.github.io/brand/components/Breadcrumbs)


apps/docs/content/components/Breadcrumbs/react.mdx
joshfarrant, line 1: Just spotted that the tab bar disappears when on the React tab

image

Can't remember why this happens, maybe because of the missing `export default ComponentLayout`?

joshfarrant (approved):
LGTM!

apps/docs/content/components/Breadcrumbs/index.mdx
```diff
@@ -1,46 +1,40 @@
...
-  <Breadcrumbs.Item href="/copilot">GitHub Copilot</Breadcrumbs.Item>
-  <Breadcrumbs.Item href="/copilot/chat" selected>
-    Chat
-  </Breadcrumbs.Item>
-</Breadcrumbs>
-```
+Breadcrumbs are most appropriate on pages that:
+
+- Do not have a section-level navigation
+- May need the ability to quickly go back to the previous (parent) page
```
rezrah, line 29: Could we give an example of section level navigation? Guessing you mean SubNav?

```suggestion
- Are missing section-level navigation elements E.g. [SubNav](/components/SubNav)
- Need to surface contextually-relevant content through its parent page(s)
```

apps/docs/content/components/Breadcrumbs/index.mdx
```diff
@@ -63,17 +57,19 @@ Ensure that the text of each individual Breadcrumb item fits entirely within a 3
   </Dont>
 </DoDontContainer>
 
-## Component props
+## Options
+
+### Variants
 
-### Breadcrumbs <Label>Required</Label>
+The breadcrumbs component is available in two variants: a default variant, where links use the standard text color, positioning the navigation as a secondary element on the page, and an accent variant, which emphasizes parent links with an accent color. The default variant is suitable for most scenarios, while the accent variant should be reserved for pages where the breadcrumbs are essential for navigation.
```
rezrah, line 64: Suggestion to break this up for easier reading.

```suggestion
The breadcrumbs component is available in two variants:

- The default variant uses the standard text color, positioning the navigation as a secondary element on the page. This is suitable for most scenarios.  
- The accent variant emphasizes parent links with an accent color. This variant should should be reserved for pages where the breadcrumbs are essential for navigation.
```

apps/docs/content/components/Breadcrumbs/react.mdx
```diff
@@ -0,0 +1,55 @@
+---
+title: Breadcrumbs
+source: https://github.com/primer/brand/tree/main/packages/react/src/Breadcrumbs/Breadcrumbs.tsx
+storybook: '/brand/storybook/?path=/story/components-breadcrumbs--playground'
+description: Use breadcrumbs to enable wayfinding through navigational links
```
rezrah, line 5: The overview page has a different description. Can we make them consistent?

Suggestion to go with your version...

```suggestion
description: Breadcrumbs display the current page within the site's hierarchy, allowing users to easily navigate to a different level.
```

Only reservation is using the term 'site' as the use-case for this isn't just limited to websites alone

apps/docs/content/components/Breadcrumbs/react.mdx
danielguillan: I forget about this every single time. Thanks @joshfarrant!

rezrah (approved):
Looks great, thanks @danielguillan.

PR #926, opened 2025-02-18
Make Pagination behaviour consistent across viewports

joshfarrant:
## Summary

Display a condensed Pagination on narrow (<40 rem) viewports. Update Pagination styles.

## List of notable changes:

- ⚠️ Deprecated the responsive object behaviour of the `showPages` prop, eg `showPages={{narrow: false, regular: true, wide: true}}` as it actively promoted inaccessible behaviour.
  - Hiding the page numbers pagination on some viewports is a violation of [WCAG SC 1.4.10 Reflow](https://www.w3.org/WAI/WCAG22/Understanding/reflow) as content must be presented without loss of information or functionality when changing the viewport to an equivalent of 320 x 256 pixels. 
  - The `showPages` prop now only accepts a boolean. For backwards compatibility, a small amount of logic has been implemented to ease this transition which allows consumers to still pass an object. If any value in the object is true, then we set `showPages` to `true`.
- Aligned the Pagination component with the Figma component. Our previous implementation rendered larger than our Figma component.

- Are you happy with the logic which handles the deprecated `showPages` object format? Would you prefer I remove it altogether and only support a `boolean` going forward?
- Are you happy with the updated styles?

## Steps to test:

1. Take a look at the component in Storybook
1. Resize the viewport to 320px wide
1. Check that it looks good and doesn't reflow or cause horizontal scroll

- Towards https://github.com/github/accessibility-audits/issues/10393

## Contributor checklist:

## Reviewer checklist:

## Screenshots:

<table>
<tr>
</tr>
<tr>
<td valign="top">

image
image

 </td>
<td valign="top">

image

</td>
</tr>
</table>


packages/react/src/Pagination/model.ts
```diff
@@ -158,7 +158,7 @@ export function buildComponentData(
       key = 'page-prev'
       content = 'Previous'
       if (page.disabled) {
-        Object.assign(props, {as: 'button', 'aria-disabled': 'true'})
+        Object.assign(props, {rel: 'prev', as: 'button', 'aria-disabled': 'true'})
```
joshfarrant, line 161: I've added these in to make the hiding/showing of the Previous/Next text simpler. I believe these attributes should be present regardless as the buttons are still `prev`/`next` buttons irrespective of whether they are disabled or not.

packages/react/src/Pagination/Pagination.module.css
```diff
@@ -7,57 +7,27 @@
...
-  .Pagination__item--visible-wide {
-    display: block;
-  }
-}
-
-.Pagination__item:last-child {
-  margin-right: 0;
+.Pagination__item * {
+  font-size: 14px !important;
+  line-height: 20px !important;
```
danielguillan, line 30:
```suggestion
  font-size: var(--brand-text-size-100);
```

I believe we can set the font size using a design token without needing the `!important` rule, as it seems we don't need to override anything. We might also be able to remove the line-height, since the default `16px` should work here.

danielguillan (approved):
Looking great! I just left a minor comment on the CSS.

packages/react/src/Pagination/Pagination.visual.spec.ts-snapshots/Visual-Comparison-Pagination-Mobile-viewport--b3b55-ation-Narrow-Page-Numbers-Hidden-By-Default-1-linux.png
rezrah, line 1: Nit..The arrow is off-center when focussed or hovered. Can this be optically aligned somehow?

Screenshot 2025-03-05 at 13 11 44

packages/react/src/Pagination/Pagination.tsx
```diff
@@ -56,13 +56,27 @@ export const Pagination = memo(
...
   }: PaginationProps) => {
-    const {isMedium} = useWindowSize()
+    if (typeof showPages === 'object') {
+      showPages = Object.values(showPages).some(x => x)
+
+      if (process.env.NODE_ENV === 'development' || process.env.NODE_ENV === 'test') {
+        // eslint-disable-next-line no-console
+        console.warn('The responsive object value for `showPages` is deprecated. Please use a boolean value instead.')
+      }
+    }
```
rezrah, line 72: I can see why this is useful, but do we need this? Guessing we did this in another component, and in hindsight think we should rely on putting a more comprehensive migration guide into the change log entry. That avoids us having to clean up logs like this down the line.

packages/react/src/Pagination/Pagination.module.css
joshfarrant: Great suggestions, thank you.

The line-height is required in this case as, without it, we have a very small vertical alignment issue within the buttons

image

.changeset/ninety-ducks-worry.md
```diff
@@ -0,0 +1,8 @@
+---
+'@primer/react-brand': minor
+---
+
+- Updated the `Pagination` component to ensure consistent behaviour across all viewports by displaying a condensed pagination on narrow viewports.
+- ⚠️ Deprecated responsive object support in the `showPages` prop.
+  - Eg, `showPages={{narrow: false, regular: true, wide: true}}` is no longer supported.
+  - The `showPages` prop now only accepts a boolean value to hide/show the page numbers on all viewports.
```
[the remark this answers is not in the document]
joshfarrant: Works for me! Take a look c3e9f039

packages/react/src/Pagination/Pagination.visual.spec.ts-snapshots/Visual-Comparison-Pagination-Mobile-viewport--b3b55-ation-Narrow-Page-Numbers-Hidden-By-Default-1-linux.png
joshfarrant: Good spot, fixed in e98d2f7393b730719af9e2f7b86be0d84ea42a1a

packages/react/src/Pagination/Pagination.module.css
```diff
@@ -7,57 +7,27 @@
...
-  .Pagination__item--visible-wide {
-    display: block;
-  }
-}
-
-.Pagination__item:last-child {
-  margin-right: 0;
+.Pagination__item * {
+  font-size: var(--brand-text-size-100);
+  line-height: 20px;
```
rezrah, line 30: As we try not to use px values directly...

```suggestion
  line-height: var(--base-size-20);
```

packages/react/src/Pagination/Pagination.module.css
```diff
@@ -107,9 +77,40 @@
...
+}
+
+.Pagination__item[rel='prev'] svg,
+.Pagination__item[rel='next'] svg {
+  margin-inline: 0;
+  position: relative;
+}
+
+.Pagination__item[rel='prev'] svg {
+  left: 2px;
```
rezrah, line 92:
```suggestion
  left: var(--base-size-2);
```

packages/react/src/Pagination/Pagination.module.css
```diff
@@ -107,9 +77,40 @@
...
+}
+
+.Pagination__item[rel='next'] svg {
+  right: 2px;
```
rezrah, line 96:
```suggestion
  right: var(--base-size-2);
```

packages/react/src/Pagination/Pagination.tsx
```diff
@@ -98,7 +97,7 @@ type UsePaginationPagesParameters = {
   hrefBuilder: (n: number) => string
   pageAttributesBuilder?: (n: number) => {[key: string]: string}
   marginPageCount: number
-  showPages?: PaginationProps['showPages']
+  showPages: boolean
```
rezrah, line 100: On line 37 the same value can also be undefined, should that be the case here too?

joshfarrant: No, there's no reason for this hook to accept `showPages` as being `undefined` as L53 sets a default value of `showPages` to be `true`.

So `showPages` can be `undefined` when passed to the `Pagination` component, however the `Pagination` will then make sure it's definitely a boolean before passing it to the hook.

packages/react/src/Pagination/Pagination.module.css
joshfarrant: Good spot, thanks!

rezrah (approved):
🚀

packages/react/src/Pagination/Pagination.module.css
```diff
@@ -7,57 +7,27 @@
...
-  }
-
-  .Pagination__item--visible-wide {
-    display: block;
-  }
-}
-
-.Pagination__item:last-child {
-  margin-right: 0;
+.Pagination__item * {
```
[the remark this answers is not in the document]
joshfarrant: Good suggestion, thanks 🙌

PR #927, opened 2025-02-18
Next.js + Nextocat documentation

rezrah:
## Summary

🤫 [_Next_ docs](https://primer-8643da8bad-26139705.drafts.github.io/brand/next/)

## List of notable changes:

- Dark ship next.js docs
  - Moved over two components and some general-purpose pages
- Fixed some typescript issues stemming from Next.js dependencies in the monorepo
- Upgrade Storybook to latest due to versioning conflicts in monorepo
- Upgrade Playwright due to versioning conflicts
- Regenerated snapshots because of the above (no change)

-
-

## Steps to test:

1.
1.
1.

-
-

## Contributor checklist:

## Reviewer checklist:

## Screenshots:

<table>
<tr>
</tr>
<tr>
<td valign="top">

 </td>
<td valign="top">

</td>
</tr>
</table>


danielguillan (approved):
This is so cool! Let's ship this one and build from here. 🚀

packages/react/src/forms/Radio/Radio.tsx
```diff
@@ -50,8 +50,8 @@ const _Radio = (
         value={value}
         {...rest}
       />
-      <label htmlFor={uniqueId} className={clsx(styles.Radio, className)}>
-        <span className={styles['Radio-dot']} />
+      <label htmlFor={uniqueId} className={clsx(styles.Radio, className)} aria-label={value}>
+        <span className={styles['Radio-dot']}></span>
```
joshfarrant, line 54: What was the reason for making this change? This label is only used to render the radio button and it's dot, and not as the descriptive label for the component.

Adding the label here degrades the screen reader experience a bit as NVDA will announce the label twice.

Before the change, NVDA announces:

```
Mona  radio button  not checked  
Hubot  radio button  not checked  
```

After the change, it announces:
```
Mona mona  radio button  not checked  
Hubot hubot  radio button  not checked  
```

Happy to discuss this further if there's a reason for the change

packages/react/src/Card/Card.tsx
```diff
@@ -86,11 +86,11 @@ const CardRoot = forwardRef<HTMLDivElement, CardProps>(
...
-          child.type === CardHeading ||
-          child.type === CardDescription
+          (child as React.ReactElement).type === CardImage ||
+          (child as React.ReactElement).type === CardIcon ||
+          (child as React.ReactElement).type === CardLabel ||
+          (child as React.ReactElement).type === CardHeading ||
+          (child as React.ReactElement).type === CardDescription
         ) {
           return true
         }
```
joshfarrant, line 96: Not one for this PR, but since we have this kind of check all over the codebase we might benefit from a small centralised [type predicate](https://www.typescriptlang.org/docs/handbook/2/narrowing.html#using-type-predicates) which will remove the need for dozens of these type assertion throughout the codebase.

```tsx
// utils.ts
export const isValidReactElement = (element: unknown): element is React.ReactElement => {
  return React.isValidElement(child) && typeof child.type !== 'string');
};
```

```suggestion
     // Use the type predicate here
     if (isValidReactElement(child)) {
        if (
          isFragment(child) ||
          // So that TypeScript knows that `child` is a ReactElement here
          child.type === CardImage ||
          child.type === CardIcon ||
          child.type === CardLabel ||
          child.type === CardHeading ||
          child.type === CardDescription
        ) {
          return true
        }
```

apps/next-docs/scripts/build
```diff
@@ -0,0 +1,8 @@
+rm -rf ../../apps/docs/public/next
+mkdir -p ../../apps/docs/public/next
+npm run clean
+mkdir -p ../../apps/docs/public/next
+npm run clean
```
joshfarrant, line 5: Do we need to do this twice?

packages/react/src/forms/Radio/Radio.tsx
joshfarrant: Sounds good, thanks for fixing!

apps/next-docs/scripts/build
rezrah: 👍 thanks, removed

Issue #928, opened 2025-02-24
🐛 [BUG] - Icon in Card component with cut off edges


stefankp:
### Describe the bug

As seen at the bottom of https://github.com/features/copilot, the edges of the icons are cut off by the border-radius.

### Reproduction steps

```bash
1. Go to https://github.com/features/copilot
2. Scroll down to Cards section
3. See error
```

### Expected behavior

```bash
1. Go to https://github.com/features/copilot
2. Scroll down to Cards section
3. See icons without edges cut off
```

### Screenshots

<img width="1354" alt="Card components with cut off icons" src="https://github.com/user-attachments/assets/8534522f-a4ab-43f7-832f-be0c07cd8aed" />

### Browsers

Safari 18.3

### OS

Mac OS 15.3

PR #931, opened 2025-02-27
Wrap Icon component's SVG in a div to fix Safari rendering issue

joshfarrant:
## Summary

Fixes an issue where the edges of an `Icon` component's rendered SVG would be cut off in Safari.

## List of notable changes:

- Wraps the rendered SVG in a div
- Applies styling to wrapper component, rather than directly to the SVG

- Take a look at the docs/storybook to make sure I've not broken any styling

- Closes https://github.com/primer/brand/issues/928

## Contributor checklist:

## Reviewer checklist:

## Screenshots:

<table>
<tr>
</tr>
<tr>
<td valign="top">

image

 </td>
<td valign="top">

image

</td>
</tr>
</table>


danielguillan (approved):

.changeset/grumpy-countries-fail.md
danielguillan, line 1: Is this bug only happening in Safari? The old snapshot, which isn't rendered in Safari, shows a cut-off icon.

joshfarrant: Sorry Dani, I missed this comment until now.

That's a great spot! I don't have a good answer for why that was happening before, apart from the fact that the one story/snapshot where this occurred was where we pass in a custom SVG into the card, rather than relying on an Octicon.

I don't know the specifics of why it was happening there, but it looks like it was a different (but related) issue which has been resolved along with this fix.

packages/react/src/Icon/Icon.tsx
```diff
@@ -1,4 +1,4 @@
-import React, {type SVGAttributes} from 'react'
+import React, {cloneElement, isValidElement, type ReactElement, type SVGAttributes} from 'react'
```
rezrah, line 1: Nit, but non-blocking. I prefer to not destructure React methods to help code readability on long component files. Makes it obvious it's not a local function. Think it's fine for types though.

packages/react/src/Icon/Icon.module.css
```diff
@@ -3,6 +3,41 @@
   padding: 0;
 }
 
+.Icon--size-20 {
+  width: var(--base-size-20);
+  height: var(--base-size-20);
+}
+
+.Icon--size-24 {
```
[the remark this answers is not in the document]
joshfarrant: This was all working before, but these styles didn't need to be explicitly declared in this component as the size was blindly passed through to the underlying icon.

Now that we're wrapping the icon in a div we need to size the div instead, so we have to handle the sizing ourselves here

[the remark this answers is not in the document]
rezrah: Gotcha, thanks 👌

rezrah (approved):

Issue #765, opened 2024-09-23
[Feature request] riverAccordion


nsolerieu:
## Context

We often stack our river to create a product features et narrative. This leads to long pages that spread small amount of incremental information over lots of scroll and many demos. By condensing the content of multiple river into an accordion we could create a stronger condensed narrative.

Here is an example of both approach side by side:

![Image](https://github.com/user-attachments/assets/4fd76a98-0dc8-444f-8b19-74053a3f4864)

@simmonsjenna has been exploring similar UX for the homepage section

![Image](https://github.com/user-attachments/assets/db6d020f-e83d-4d6c-9e26-88e1cf35935b)

## Design suggestion

The component relies structurally on up 4 text+image internal units (with a minimum of 2, 1 being a regular river) and an active state rotation setting focus on one unit at the time. It would not be proper accordion that the extent that it would have a single active element at any time.

- Similarly to the `riverStoryScroll` we could combine multiple rivers
- Create a variant of the `river` with multiple image+content internal units
- Create a net new component

#### Component props:
Each props would apply equally to all units:

- Autoplay (default: true - 5s)
- Link (default: false)
- Status icon (default: true) = + or - depending on active item
- Description (default: false)
- leadingIcon (default: false)

## Timing/urgency

Homepage is likely going to need a quick build with very specific styling led by brand refactoring thus will be custom built. I suggest that this should inform the definition phase in parallel.

rezrah:
Triaged at planning on 28th October.

- @simmonsjenna will present the pattern at a11y office hours
   - Some ambiguity around motion and auto-scrolling behaviours  
- @jesussandreas will review the existing designs for accuracy and provide a design file diffing any changes
- Primer Brand able to take on this request in Q2

simmonsjenna:
@rezrah Following up with my notes from the A11y Design pairing session with @ericwbailey and @janmaarten-a11y:

- Advice: simplify by getting rid of the definition lists and let the buttons themselves be the enumeration
- Suppress tab focus on child links when in collapsed state (e.g. when tabbing through and the item is not expanded, the child links like "Download GitHub Mobile" are still getting focused)

We also discussed autoplay and how that might be approached. I think we should leave autoplay out of this component and revisit it in the future if needed due to the complexities it introduces.

jesussandreas:
Hellooo, here's the reviewed design for us to apply on other pages.

It follows patterns that are close to the river container in order to ensure it feels part of the component ecosystem.

[...]

PR #939, opened 2025-03-10
Add RiverAccordion component

joshfarrant:
## Summary

Adds RiverAccordion component.

## List of notable changes:

- Added RiverAccordion component.
  - Includes `RiverAccordion.Item`, `RiverAccordion.Content`, `RiverAccordion.Heading`, and `RiverAccordion.Visual`.
- Modified Heading component to remove usage of `createElement`.
  - Calling `createElement` each time that the Heading was rendered caused the previously rendered heading to be torn down and a new one created in its place. When the Heading was being used as a child of a button — as I'm doing here — this caused focus to be lost whenever the button and Heading were re-rendered. The result of this was that when an accordion item was expanded using the keyboard then focus would be lost from the button which is disorienting for assistive tech users.

- I added a bit of animation to the expanding/collapsing panels. To achieve the effect of the content sliding out from behind the grey line between items I slide the content up slightly as the accordion item expands. The below video shows the animation slowed down so you can see the effect more clearly. I'm quite happy with how it turned out, but if you have any feedback please let me know.

https://github.com/user-attachments/assets/291d8c1a-5b9b-4641-9003-4f027219f932

- The implementation for this component was based heavily on the [ARIA Authoring Practises Guide (APG) for the Accordion pattern](https://www.w3.org/WAI/ARIA/apg/patterns/accordion/). All of the required functionality should be implemented, however if there's anything I've missed please do let me know.
- This implementation has been tested with a screen reader; specifically NVDA. You can see a video of the screen reader in action below (turn your sound on to hear the announcements). The slightly odd _unavailable expanded_ announcement is due to the `aria-disabled="true"` which is added to the button when a panel is expanded, as per the APG here:
  > _If the accordion panel associated with an accordion header is visible, and if the accordion does not permit the panel to be collapsed, the header button element has aria-disabled set to true._

https://github.com/user-attachments/assets/62d942f7-5e24-47f2-81d8-e0c5420caf5a

## Steps to test:

- 🔗 [Link to Storybook](https://primer-dffe8bfd01-26139705.drafts.github.io/brand/storybook/?path=/story/components-riveraccordion--default)
- 🔗 [Link to Docs](https://primer-dffe8bfd01-26139705.drafts.github.io/brand/components/RiverAccordion)

- Closes https://github.com/primer/brand/issues/765

## Contributor checklist:

## Reviewer checklist:


apps/docs/content/components/RiverAccordion.mdx
```diff
@@ -0,0 +1,193 @@
+---
+title: River Accordion
```
rezrah, line 2:
```suggestion
title: River accordion
```

joshfarrant: fd3b66ccb25a31d1faf39a18135932b2b1170c6f

apps/docs/content/components/RiverAccordion.mdx
```diff
@@ -0,0 +1,193 @@
...
+description: Use the river accordion to create collapsible content panels with associated visuals.
+source: https://github.com/primer/brand/blob/main/packages/react/src/river/RiverAccordion/RiverAccordion.tsx
+figma: 'https://www.figma.com/design/BJ95AjraesmRCWsKA013GS/Primer-Brand?node-id=16649-1497&p=f&t=lzotdAbtGDWpDEej-11'
```
rezrah, line 5: I think this should link to the frame containing the examples?

```suggestion
figma: 'https://www.figma.com/design/BJ95AjraesmRCWsKA013GS/Primer-Brand?node-id=16649-1564&t=hUxfzn1hD1LwE1aQ-4'
```

apps/docs/content/components/RiverAccordion.mdx
```diff
@@ -0,0 +1,193 @@
...
+storybook: '/brand/storybook/?path=/story/components-riveraccordion--playground'
```
rezrah, line 6: Could we create a default example and link to that instead of playground please?

apps/docs/content/components/RiverAccordion.mdx
```diff
@@ -0,0 +1,193 @@
...
+<RiverAccordion align="start">
+  <RiverAccordion.Item>
+    <RiverAccordion.Heading>Heading 1</RiverAccordion.Heading>
+    <RiverAccordion.Content>
+      <Text as="p" variant="muted" weight="medium" size="200">
+        Lorem ipsum dolor sit amet, consectetur adipiscing elit. In sapien sit
+        ullamcorper id. Aliquam luctus sed turpis felis nam pulvinar risus
+        elementum.
+      </Text>
+      <Link href="#" variant="accent">
```
rezrah, line 30: Is the use of accent links required in this component and something we need enforce @danielguillan?

rezrah: Might just be something we can capture int he interface guidelines, but want to make sure the design can be used with the normal variant too, otherwise we should consider enforcing it.

apps/docs/content/components/RiverAccordion.mdx
```diff
@@ -0,0 +1,193 @@
...
+        Call to action
+      </Link>
+    </RiverAccordion.Content>
+    <RiverAccordion.Visual>
+      <img
+        src="https://placehold.co/600x400/FF5733/ffffff?text=1"
```
rezrah, line 36: Could we make this a local fixture please?

apps/docs/content/components/RiverAccordion.mdx
```diff
@@ -0,0 +1,193 @@
...
+## Examples
+
+### Default
+
+```jsx live
+<RiverAccordion align="start">
+  <RiverAccordion.Item>
+    <RiverAccordion.Heading>Heading 1</RiverAccordion.Heading>
+    <RiverAccordion.Content>
+      <Text as="p" variant="muted" weight="medium" size="200">
```
rezrah, line 25: Similar to the link below, this is quite optioned and I'm wondering if the pattern falls apart if someone forgets to add of most of this (particularly weight). I think we should set these as the default styles if we can.

apps/docs/content/components/RiverAccordion.mdx
```diff
@@ -0,0 +1,193 @@
...
+        alt="Visual content for item 3"
+      />
+    </RiverAccordion.Visual>
+  </RiverAccordion.Item>
+</RiverAccordion>
+```
+
+### Visual alignment
+
+Use the `align` prop to control the position of the visuals. Setting `align="end"` will position the visuals on the right side of the content.
```
rezrah, line 86:
```suggestion
Use the `align` prop to control the position of the visuals. Setting `align="end"` will invert the default position of the visuals and content.
```

apps/docs/content/components/RiverAccordion.mdx
```diff
@@ -0,0 +1,193 @@
...
+
+`RiverAccordion.Item` extends the HTML `div` element and supports all `div` props.
+
+### RiverAccordion.Heading <Label>Required</Label>
+
+| Name       | Type     | Default | Description  |
+| :--------- | :------- | :-----: | :----------- |
+| `children` | `string` |         | Heading text |
+
+`RiverAccordion.Heading` extends the [`Heading`](/components/Stack) component and supports all `Heading` props.
```
rezrah, line 177:
```suggestion
`RiverAccordion.Heading` extends the [`Heading`](/components/heading) component and supports all `Heading` props.
```

apps/docs/content/components/RiverAccordion.mdx
```diff
@@ -0,0 +1,193 @@
...
+| `children` | `React.ReactElement[]`                                      |           | RiverAccordion.Item components                                         |
+| `align`    | <PropTableValues values={['start', 'end']} addLineBreaks /> | `'start'` | Controls the position of the visuals relative to the accordion content |
+
+`RiverAccordion` extends the HTML `div` element and supports all `div` props.
+
+### RiverAccordion.Item <Label>Required</Label>
+
+| Name       | Type                   | Default | Description                                          |
+| :--------- | :--------------------- | :-----: | :--------------------------------------------------- |
+| `children` | `React.ReactElement[]` |         | Must include Heading, Content, and Visual components |
```
rezrah, line 167: Could we add some links to the description? So they scroll to the relevant sections below like FAQ does?

Heading in particular could be easily confused with the Heading component instead of RiverAccordion.Heading?

.changeset/empty-garlics-smoke.md
```diff
@@ -0,0 +1,37 @@
...
+      <RiverAccordion.Heading>One More Heading</RiverAccordion.Heading>
+      <RiverAccordion.Content>Some content</RiverAccordion.Content>
+      <RiverAccordion.Visual>
+        <img src="placeholder.png" alt="placeholder" />
+      </RiverAccordion.Visual>
+    </RiverAccordion.Item>
+  </RiverAccordion>
+  ```
+
+- Fixed a bug in the `Heading` component where the component would be re-mounted each render, causing it to lose focus when used inside a focusable element.
```
rezrah, line 37: It doesn't seem like the user needs to make any changes here but double checking that's the case @joshfarrant? There's no breaking changes here?

joshfarrant: Yes correct, no changes needed for consumers, just a edge-case bug fix

packages/react/src/river/RiverAccordion/RiverAccordion.module.css
```diff
@@ -0,0 +1,161 @@
+.RiverAccordion {
+  display: grid;
+  grid-template-columns: 1fr;
+  grid-template-rows: auto auto;
+  grid-template-areas:
+    'visual'
+    'accordion';
+  gap: 0;
+  position: relative;
+  interpolate-size: allow-keywords;
```
rezrah, line 10: This isn't supported in Safari or baseline, can you do it another way?

apps/docs/content/components/RiverAccordion.mdx
joshfarrant: Thank you, I swear I spotted and fixed this once already, but clearly not! 😅

packages/react/src/river/RiverAccordion/RiverAccordion.module.css
```diff
@@ -0,0 +1,161 @@
...
+.RiverAccordion__accordionContainer .RiverAccordion__visual {
+  position: absolute;
+  width: 1px;
+  height: 1px;
+  padding: 0;
+  overflow: hidden;
+  clip: rect(0, 0, 0, 0);
+  word-wrap: normal;
+  border: 0;
+}
```
rezrah, line 90: Could you replace this by using our `visually-hidden` class instead? It's part of the main stylesheet.

packages/react/src/river/RiverAccordion/RiverAccordion.module.css
```diff
@@ -0,0 +1,161 @@
...
+.RiverAccordion__panel {
+  height: 0;
+  transition: all var(--brand-animation-duration-default) var(--brand-animation-easing-default);
+  margin-block-start: 32px;
+  overflow-y: clip;
+}
+
+.RiverAccordion__item--open .RiverAccordion__panel {
+  margin-block-start: 12px;
+  padding-block-end: 32px;
```
rezrah, line 140:
```suggestion
  margin-block-start: var(--base-size-12);
  padding-block-end: var(--base-size-32);
```

packages/react/src/river/RiverAccordion/RiverAccordion.module.css
```diff
@@ -0,0 +1,161 @@
...
+  position: absolute;
+  display: block;
+  width: 100%;
+  height: 100%;
+}
+
+.RiverAccordion__panel {
+  height: 0;
+  transition: all var(--brand-animation-duration-default) var(--brand-animation-easing-default);
+  margin-block-start: 32px;
```
rezrah, line 134:
```suggestion
  margin-block-start: var(--base-size-32);
```

packages/react/src/river/RiverAccordion/RiverAccordion.module.css
```diff
@@ -0,0 +1,161 @@
...
+  content: '';
+  position: absolute;
+  display: block;
+  width: 100%;
+  height: 100%;
+}
+
+.RiverAccordion__panel {
+  height: 0;
+  transition: all var(--brand-animation-duration-default) var(--brand-animation-easing-default);
```
rezrah, line 133: Can you narrow the scope from all to just height (and whatever else you need)?

packages/react/src/river/RiverAccordion/RiverAccordion.module.css
joshfarrant: Not easily unfortunately.

The `.RiverAccordion__visual` element is used in two places; once for the large image that is visible, and again for the visually-hidden image contained within the accordion content. To make this work using `visually-hidden` I'd have to add the `visually-hidden` class to all visuals by default, then selectively remove that class when rendering the large (non-hidden) visual. It's possible but cumbersome, which is why I took this approach.

I can do that if you'd like, but this felt like the simplest approach

packages/react/src/river/RiverAccordion/RiverAccordion.module.css
joshfarrant: Let me look into it 👍

packages/react/src/river/RiverAccordion/RiverAccordion.tsx
```diff
@@ -0,0 +1,227 @@
...
+
+const RiverAccordionContent = ({className, ...props}: RiverAccordionContentProps) => {
+  return <div className={clsx(styles.RiverAccordion__content, className)} {...props} />
+}
+
+const RiverAccordionVisual = ({className, ...props}: RiverAccordionVisualProps) => {
+  return <div className={clsx(styles.RiverAccordion__visual, className)} {...props} />
+}
+
+const createComponentTypeGuard =
```
rezrah, line 215: Nice!

packages/react/src/river/RiverAccordion/RiverAccordion.tsx
```diff
@@ -0,0 +1,227 @@
...
+  <T,>(componentType: React.ComponentType<T>) =>
+  (element: unknown): element is React.ReactElement<T> =>
+    React.isValidElement(element) && element.type === componentType
+
+const isRiverAccordionItem = createComponentTypeGuard(RiverAccordionItem)
+const isRiverAccordionContent = createComponentTypeGuard(RiverAccordionContent)
+const isRiverAccordionHeading = createComponentTypeGuard(RiverAccordionHeading)
+const isRiverAccordionVisual = createComponentTypeGuard(RiverAccordionVisual)
+
+export const RiverAccordion = Object.assign(RiverAccordionRoot, {
```
rezrah, line 225:
```suggestion
/**
 * Use the RiverAccordion to create collapsible content panels with associated visuals.
 * @see https://primer.style/brand/components/RiverAccordion
 */
 export const RiverAccordion = Object.assign(RiverAccordionRoot, {
```

apps/docs/content/components/RiverAccordion.mdx
```diff
@@ -0,0 +1,193 @@
+---
+title: River Accordion
+description: Use the river accordion to create collapsible content panels with associated visuals.
```
rezrah, line 3: Being pedantic but should this be `to create expandable content panels with associated visuals.` instead of `collapsible` because you a) can't collapse an open panel and if you're presenting a visual you're expanding not collapsing.

packages/react/src/river/RiverAccordion/RiverAccordion.tsx
```diff
@@ -0,0 +1,227 @@
...
+      <button
+        type="button"
+        className={styles.RiverAccordion__trigger}
+        onClick={onClick}
+        aria-disabled={isOpen}
+        aria-controls={id}
+        aria-expanded={isOpen}
+      >
+        {children}
+        <PlusIcon size={24} className={styles.RiverAccordion__icon} />
```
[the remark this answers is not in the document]
joshfarrant: I wondered this as well.

I ended up not using our own Accordion component as a details/summary approach (which Accordion uses under the hood) really didn't want to work in this particular way (exactly always 1 item open).

The [Authoring Practises Guidelines for the Accordion pattern](https://www.w3.org/WAI/ARIA/apg/patterns/accordion/) do include guidelines on what to do for our particular pattern, so by their definition at least it is an Accordion.

> If the accordion panel associated with an accordion header is visible, and if the accordion does not permit the panel to be collapsed, the header button element has [aria-disabled](https://w3c.github.io/aria/#aria-disabled) set to true.

.changeset/empty-garlics-smoke.md
```diff
@@ -0,0 +1,37 @@
...
+    <RiverAccordion.Item>
+      <RiverAccordion.Heading>One More Heading</RiverAccordion.Heading>
+      <RiverAccordion.Content>Some content</RiverAccordion.Content>
+      <RiverAccordion.Visual>
+        <img src="placeholder.png" alt="placeholder" />
+      </RiverAccordion.Visual>
+    </RiverAccordion.Item>
+  </RiverAccordion>
+  ```
+
```
rezrah, line 36: Adding the usual external link for more info...

```suggestion

🔗 [See the documentation for more usage examples](https://primer.style/brand/components/RiverAccordion)

```

packages/react/src/river/RiverAccordion/RiverAccordion.module.css
rezrah: No worries! Just calling it out in case you weren't aware of that utility class being available. Thanks for clarifying.

packages/react/src/river/RiverAccordion/RiverAccordion.tsx
rezrah: I think we have a helpers function, but you can move it there after. It's only worthwhile if it's a direct swap out in other components, so could do in a follow up.

apps/docs/content/components/RiverAccordion.mdx
joshfarrant: Would you prefer I update the implementation of `RiverAccordion.Content` to wrap all children in `<Text as="p" variant="muted" weight="medium" size="200">`, or apply all of the styles that `Text` would apply to the `.RiverAccordion__content` class?

rezrah: If you use a css class with those defaults, are they still overrideable using the Text props? I think that's the easier approach, but not sure if specificity would make overriding harder (which we should still allow). The children approach should still allow that.

packages/react/src/river/RiverAccordion/RiverAccordion.module.css
joshfarrant: Replaced `interpolate-size: allow-keywords;` with a `grid-template-rows` transition, as we do on the homepage.

apps/docs/content/components/RiverAccordion.mdx
joshfarrant: Used a class in the end. Tested and it can still be overridden by `<Text>` ✅

packages/react/src/river/RiverAccordion/RiverAccordion.tsx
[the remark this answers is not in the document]
nsolerieu: Good catch, I'd follow the styling and interaction of the FAQ component where we have `+` and `-` states with a nice rotate transition. That fits the expected, traditional `accordion` interaction.

I don't have full context on how this component but I'd expect chevrons up/down instead of plus/minus like this

Screenshot 2025-04-07 at 9 17 13 AM

.changeset/empty-garlics-smoke.md
```diff
@@ -0,0 +1,39 @@
...
+- Fixed a bug in the `Heading` component where the component would be re-mounted each render, causing it to lose focus when used inside a focusable element.
+
+🔗 [See the documentation for more usage examples](https://primer.style/brand/components/RiverAccordion)
```
rezrah, line 39: Just noticed that this is in the wrong place, my bad for previous suggestion...


```suggestion

```

joshfarrant: Yes of course, my bad for dropping it into the wrong place. Thanks for spotting 👍

apps/docs/content/components/RiverAccordion.mdx
```diff
@@ -0,0 +1,181 @@
...
+
+  <RiverAccordion.Item>
+    <RiverAccordion.Heading>Heading 3</RiverAccordion.Heading>
+    <RiverAccordion.Content>
+      Lorem ipsum dolor sit amet, consectetur adipiscing elit. In sapien sit
+      ullamcorper id. Aliquam luctus sed turpis felis nam pulvinar risus
+      elementum.
+      <Link href="#" variant="accent">
+        Call to action
+      </Link>
```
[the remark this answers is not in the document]
rezrah: Also heads up that the rendered HTML in your current approach no longer wraps the description in a `<p>`, which it should do.

Screenshot 2025-04-08 at 11 11 02

packages/react/src/river/RiverAccordion/RiverAccordion.tsx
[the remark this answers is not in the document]
rezrah: @nsolerieu thanks for sharing that. I wonder if the apple version you shared suffers from the same intuitiveness issues as ours, but even more pronounced because they added the visual cue to close a panel, but it doesn't follow-through and close it when pressed 😄 



https://github.com/user-attachments/assets/47250b85-84e4-4093-920d-4c8e124d5557


If we do add the cue, I think we should allow closing or keep as-is. If former, natural question is what happens. I think you could make the first one open again, but again might feel more off. 

cc. @danielguillan ☝️

apps/docs/content/components/RiverAccordion.mdx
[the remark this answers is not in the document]
joshfarrant: Just updated it, take a look and lmk what you think @rezrah

packages/react/src/river/RiverAccordion/RiverAccordion.tsx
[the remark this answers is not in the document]
danielguillan: > even more pronounced because they added the visual cue to close a panel, but it doesn't follow-through and close it when pressed

Agree. Let's keep as-is and test further. AFAIK we haven't heard of any issues with the current design in the Copilot page.

apps/docs/content/components/RiverAccordion.mdx
danielguillan: Yes, I believe we should be consistent with other instances such as the River component and enforce the accent variant by default.

packages/react/src/river/RiverAccordion/RiverAccordion.module.css
```diff
@@ -0,0 +1,161 @@
...
+  .RiverAccordion__align-start {
+    grid-template-columns: 4fr 7fr; /* 12 columns, minus 1 for the gap */
+    grid-template-areas: 'accordion visual';
+  }
+
+  .RiverAccordion__align-end {
+    grid-template-columns: 7fr 4fr; /* 12 columns, minus 1 for the gap */
+    grid-template-areas: 'visual accordion';
+  }
+}
```
danielguillan, line 27: The math for the grid does not seem to align properly with the columns of the `Grid`. We assume there are 12 fractions of the available space, but we are not considering the gaps between the columns.

The gap between the visual and the accordion should equal 1 Grid column plus 1 gap. We can use `var(--brand-Grid-spacing-column-gap)` to accurately calculate the final spacing.

joshfarrant: Great spot @danielguillan, thank you! 🙏

I've corrected that now and double-checked the maths and the rendered sizes, but take a look and let me know what you think.

packages/react/src/river/RiverAccordion/RiverAccordion.module.css
```diff
@@ -0,0 +1,161 @@
...
+.RiverAccordion__accordionContainer .RiverAccordion__visual {
+  position: absolute;
+  width: 1px;
+  height: 1px;
+  padding: 0;
+  overflow: hidden;
+  clip: rect(0, 0, 0, 0);
+  word-wrap: normal;
+  border: 0;
+}
```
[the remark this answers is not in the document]
danielguillan: Oh! I missed that comment. Thanks for the context! 😃

danielguillan (approved):
Great work @joshfarrant!

apps/docs/content/components/RiverAccordion.mdx
rezrah: cc. @joshfarrant can we make the accent link the default please ☝️  (but continue to allow overrides if neccessary)

packages/react/src/river/RiverAccordion/RiverAccordion.tsx
```diff
@@ -0,0 +1,254 @@
...
+  const {id, index, isOpen} = useRiverAccordionItemContext()
+
+  const onClick = useCallback(() => {
+    if (!isOpen) {
+      setOpenIndex(index)
+    }
+  }, [index, isOpen, setOpenIndex])
+
+  return (
+    <Heading size="6" as="h3" className={clsx(styles.RiverAccordion__heading, className)} {...props}>
```
rezrah, line 187: Stylistic nit but the string values for `size` and `as` prop values were throwing me off a little. To make it clearer (and consistent with how we do it in other components) can we please destructure these values in the function arguments and set the default values there? Having the users overrides spread in with `...props` was a less explicit IMO.

packages/react/src/river/RiverAccordion/RiverAccordion.tsx
```diff
@@ -0,0 +1,254 @@
...
+    return React.Children.toArray(children).map(child => {
+      if (!isText(child)) {
+        return child
+      }
+
+      return React.cloneElement(child, {
+        as: 'p',
+        variant: 'muted',
+        weight: 'medium',
+        size: '200',
```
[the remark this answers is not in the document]
joshfarrant: Thank you — just spotted the same thing after applying the same change for the link 😅

Fixed now 👍

apps/docs/content/components/RiverAccordion.mdx
joshfarrant: Updated 👍

rezrah (approved):
Looks great @joshfarrant 🚀 :shipit:

PR #940, opened 2025-03-10
Allow Accordion component to be either controlled or uncontrolled

joshfarrant:
## Summary

Updates and refactors the Accordion component to allow it to be used in a controlled manner, as well as an uncontrolled manner.

Much of the previous code in the Accordion could be removed as it was handling things that the browser already handled for us as we're using the details/summary HTML elements; eg the space

## List of notable changes:

- Reduced the number of props needed to be passed down to the AccordionHeading and AccordionContent components by:
  - Removing manual tracking of `open`/`isOpen` from the component, instead using the `<details>` element's built-in `open` state.
  - Removing the passing of refs between components
  - Hoisting the focus management and <kbd>Escape</kbd> key tracking to the top-level Accordion component
- This left only one prop to be passed down — `variant` — which is now passed via context to simplify the component and remove the need to iterate through — and subsequently clone — children.
- Leaned more heavily on the `<details>` element's native `toggle` event via the `onToggle` prop. This allows us to maintain the existing `handleOpen` prop and its previous behaviour purely for backwards compatibility. The prop is now slightly redundant as the same thing can be achieved using the `<details>` element's built-in `onToggle` prop, eg
  ```tsx
  <Accordion
    handleOpen={open => {
      // Do something with `open` boolean
    }}
    onToggle={e => {
      const {open} = e.currentTarget
      // Do something with `open` boolean
    }}
  ```
- Unrelated to these changes; updated the _Color Mode_ plugin to allow per-story color mode overrides via the `parameters` property.

- Ensure functionality hasn't changed, including keyboard interactivity
- I'd be in favour of removing `handleOpen`, but keeping it isn't a big deal either

## Steps to test:

1. This component is only available in Storybook

## Contributor checklist:

## Reviewer checklist:


apps/storybook/.storybook/preview.js
```diff
@@ -36,16 +36,17 @@ export const globalTypes = {
 }
 
 const ThemeProviderDecorator = (Story, context) => {
+  const colorMode = context.parameters?.colorMode || context.globals.colorMode
```
rezrah, line 39: Was this needed for this change, or just a random optimisation?

packages/react/src/Accordion/Accordion.visual.spec.ts-snapshots/Visual-Comparison-Accordion-Accordion-Default-1-linux.png
rezrah, line 1: We've lost the default toggle color here, any ideas why @joshfarrant?

apps/storybook/.storybook/preview.js
joshfarrant: This was part of this change

> Unrelated to these changes; updated the Color Mode plugin to allow per-story color mode overrides via the parameters property.

It came about as I was updating the Accordion's stories and spotted one where we were manually simulating our 'all' color mode using grids and a theme provider. I've replaced that with just using our pre-existing color mode ([here](https://github.com/primer/brand/pull/940/files#diff-83a3f0c1c28d8a58ed806a53d5a2febb3ba84069e5747c35bbe1acbaf7bd4245R75))

Happy to remove that if you'd prefer it be pulled out.

packages/react/src/Accordion/Accordion.tsx
```diff
@@ -1,146 +1,105 @@
...
+          details.querySelector('summary')?.focus()
         }
       },
-      [handleOpen, open],
+      [_onKeyDown, ref],
     )
 
+    return (
+      <AccordionContext.Provider value={accordionContextValue}>
+        {/* eslint-disable-next-line jsx-a11y/no-noninteractive-element-interactions */}
```
rezrah, line 77: We try to avoid using eslint disable rules starting with `jsx-a11y/*` because they are tracked and reported on externally. Can we remove this?

packages/react/src/Accordion/Accordion.features.stories.tsx
```diff
@@ -0,0 +1,207 @@
...
+        </Accordion>
+      ))}
+    </>
+  ),
+}
+
+export const ExclusiveWithoutUsingNameAttribute: StoryObj = {
+  render: () => {
+    // eslint-disable-next-line react-hooks/rules-of-hooks
+    const containerRef = useRef<HTMLDivElement>(null)
```
rezrah, line 175: Is there another way to call this hook outside of a render method to avoid the rule of hook violation?

packages/react/src/Accordion/Accordion.tsx
```diff
@@ -1,146 +1,105 @@
...
+import {Heading, type HeadingProps} from '../'
 import {ChevronDownIcon, ChevronUpIcon} from '@primer/octicons-react'
 import {Colors, BiColorGradients as Gradients} from '../constants'
 
-/**
- * Design tokens
- */
 import '@primer/brand-primitives/lib/design-tokens/css/tokens/functional/components/accordion/colors-with-modes.css'
 
-/** * Main Stylesheet (as a CSS Module) */
```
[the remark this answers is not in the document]
joshfarrant: 👍 b1a43ff10d8711e4484ae268fd6ff911d2fd0def

packages/react/src/Accordion/Accordion.visual.spec.ts-snapshots/Visual-Comparison-Accordion-Accordion-Default-1-linux.png
joshfarrant: Good spot though, thank you 🙌

packages/react/src/Accordion/Accordion.tsx
joshfarrant: 👍 5b3a8e1e8b8c12819b7273266d37b9f15f64823a

rezrah:
Could you please add a changeset to this PR @joshfarrant?

rezrah:
We should add them whenever there's a material change to the package. Technically without one, you can't prepare a release, although in our case one is already open). Agree that's it's not immediately obvious what we should add there so as a suggestion:

```
Updated the `Accordion` component to support additional state handling methods. All changes are backwards compatible. Refer to Storybook for examples of these additional `Accordion` features.
```

packages/react/src/Accordion/Accordion.visual.spec.ts-snapshots/Visual-Comparison-Accordion-Accordion-Toggle-Colors-1-linux.png
rezrah, line 1: As the story is now full bleed we're seeing an issue with the background not applying to the full page height. Probably because 100dvh is set, when it should be 100% in the storybook theme provider, can we update that please and regen @joshfarrant?

PR #943, opened 2025-03-12
Set Card content max-width

danielguillan:
## Summary

Sets a maximum width for the heading and description of the `Card` component to prevent very long lines on wider cards.

- Related issue: https://github.com/github/marketing-engineering/issues/4193#issuecomment-2707285560

## Contributor checklist:

## Reviewer checklist:

## Screenshots:

<table>
<tr>
</tr>
<tr>
<td valign="top">

image

 </td>
<td valign="top">

image

</td>
</tr>
</table>


packages/react/src/Card/Card.visual.spec.ts-snapshots/Visual-Comparison-Card-Card-Icon-Colors-1-linux.png
rezrah, line 1: @danielguillan it looks like the overall width of the card has been affected here somehow? Can't immediately see how that would happen though, any ideas? I wouldn't expect this diff here.

rezrah (approved):
LGTM 

One question for you @danielguillan, which looks like it could either be a regression or flakey test? Hopefully latter, as otherwise this could cause significant downstream regression too 🤔

packages/react/src/Card/Card.visual.spec.ts-snapshots/Visual-Comparison-Card-Card-Icon-Colors-1-linux.png
danielguillan: @rezrah  Good catch! With this update, the card's maximum total width is now limited to `400px` (based on the descriptions' `max-width`) instead of the `420px` `max-width` set at the container level.

To avoid regressions, we could match the description's `max-width` to the cards' current `max-width` (i.e., `420px`). Wdyt? We could set `400px` on the full-width variant only, but I don't believe it's worth it.

rezrah: That sounds like a good plan 👍. Yeah let's do that please. If we can limit scope of these changes to inner content only like the PR indicates, that would help reduce the testing burden and risks of regressions outside of the card.

danielguillan:
@rezrah I limited these changes to the full-width variant to avoid any potential regressions in card instances with an intrinsic size.

rezrah (approved):
🚀

PR #945, opened 2025-03-19
Remove global CSS rules for reduced motion

rezrah:
## Summary

Resolves https://github.com/github/primer/issues/4904

Removes the global css rules for `reduced-motion` in `reset.css`. 

Moves those rules to individual components that require them instead. 

## List of notable changes:

- Updated all standalone components that feature either transitions or animation rules

- See if any components were missed.

## Steps to test:

1. Enable reduced-motion in your OS setting
1. Go through Storybook and verify that motion is still disabled

## Contributor checklist:

## Reviewer checklist:

## Screenshots:
N/A should work same as before


packages/react/src/Card/Card.module.css
```diff
@@ -197,3 +197,15 @@
 .Card:has(.Card__link:hover) .Card--expandableArrow path:nth-of-type(2) {
   stroke-dashoffset: 20;
 }
+
+/* Remove all animations, transitions and smooth scroll for people that prefer not to see them */
+@media (prefers-reduced-motion: reduce) {
```
joshfarrant, line 202: Looking through the docs with prefers-motion enabled, Card feels a bit jarring as it jumps to a slightly bigger size.

Do you think we should remove the scale effect altogether? If we wanted to keep a hover style then maybe we could apply a focus ring on hover instead?

joshfarrant (commented):
Looking good!

Only other thing I spotted was that EyebrowBanner still has a transition on the arrow when hovered

packages/react/src/Card/Card.module.css
rezrah: > Card feels a bit jarring as it jumps to a slightly bigger size
> Do you think we should remove the scale effect altogether? 

Yeah, agree but would prefer we don't tackle that here. The scale effect is existing behaviour and unrelated to the global rule change here. The goal of this update is to have parity and not regress. 

I do remember various conversations about this before though 🤔, and a11y design confirmed this was okay. @joshfarrant did this come up in recent audits? I also think focus ring could be misleading because that relates to a different type of interaction; focus vs hover.

rezrah:
> Looking good!
> 
> Only other thing I spotted was that EyebrowBanner still has a transition on the arrow when hovered

Thanks Josh, didn't realise that component had a transition. Fixed ✔️

packages/react/src/Card/Card.module.css
joshfarrant: No prob, that works for me 👍

No this hasn't come up at all. Purely my opinion that the jump in size feels a bit off to me

joshfarrant (approved):
LGTM!

danielguillan (commented):
I think we're missing a couple of components:
- [Prose](https://github.com/primer/brand/blob/main/packages/react/src/Prose/Prose.module.css#L199)
- [Timeline](https://github.com/primer/brand/blob/main/packages/react/src/Timeline/Timeline.module.css#L65)

Also, should we add the rules to remove animations and transitions from the `Animation` classes?

PR #947, opened 2025-03-20
improve VideoPlayer appearance on narrow viewports

joshfarrant:
## Summary

Improved the appearance of the `VideoPlayer` component on narrow viewports

## List of notable changes:

- Moved the controls to below the video, rather than overlaying the video
- Inserted closed captions between the controls and the video, rather than overlaying the video
- Reduced the size of the video title on narrow viewports
- Removed the `showControlsWhilePlaying` prop as hiding/showing the controls causes the height of the VideoPlayer component to change, which isn't a great UX
- Removed `Controls` component as it's no longer uses (and hasn't been for a while)
- Fixed a bug where a control's tooltip wouldn't close when it was clicked.

- Any feedback on the design would be appreciated.
- I use `#000000` in a couple of places; is there a token that would be preferable in this case?

## Steps to test:

- 🔗 [Link to Storybook](https://primer-fa46115b76-26139705.drafts.github.io/brand/storybook/?path=/story/components-videoplayer--default)
- 🔗 [Link to docs](https://primer-fa46115b76-26139705.drafts.github.io/brand/components/VideoPlayer)

- Towards https://github.com/github/accessibility-audits/issues/10406

## Contributor checklist:

## Reviewer checklist:

## Screenshots:

<table>
<tr>
<th> Before</th> <th> After (cc hidden) </th>  <th> After (cc visible)</th>
</tr>
<tr>
<td valign="top">

image

 </td>
<td valign="top">

image

 </td>
<td valign="top">

image

</td>
</tr>
</table>


packages/react/src/VideoPlayer/components/Captions/Captions.tsx
```diff
@@ -39,13 +39,13 @@ export const Captions = ({className, ...props}: HTMLAttributes<HTMLDivElement>)
     return () => videoRef.removeEventListener('timeupdate', compareAndSetCaption)
   }, [videoRef, trackInformation])
 
-  if (!caption) {
+  if (!trackInformation) {
```
joshfarrant, line 42: Ensure the captions bar is always visible if captions are enabled and available. This prevents the height of the VideoPlayer reducing when no captions are visible.

packages/react/src/VideoPlayer/VideoPlayer.module.css
```diff
@@ -96,50 +118,49 @@
...
-  transition: opacity var(--brand-animation-duration-default) var(--brand-animation-easing-default);
-  opacity: 0;
+.VideoPlayer__controlsBar__row1,
+.VideoPlayer__controlsBar__row2 {
+  display: flex;
+  align-items: center;
+  gap: var(--base-size-16);
+}
+
+@media (max-width: 30rem) {
```
joshfarrant, line 144: 30rem isn't a breakpoint that we use anywhere else, however it makes sense for this particular component.

This is the point at which the single row of controls breaks to two rows

rezrah: Our `small` breakpoint is `34rem`, could that work?

packages/react/src/VideoPlayer/VideoPlayer.module.css
```diff
@@ -18,20 +18,37 @@
 
 .VideoPlayer__container {
   width: 100%;
-  position: relative;
   border-radius: var(--brand-borderRadius-medium);
+  overflow: hidden;
+  background: #000000;
```
rezrah, line 23: Is this supposed to be black in all color modes? or the usual background color?

If former you want:

```suggestion
  background-color: var(--base-color-scale-black-0);
```

packages/react/src/VideoPlayer/components/VideoTooltip/VideoTooltip.tsx
```diff
@@ -29,17 +29,24 @@ export const VideoTooltip = ({children, className, ...rest}: VideoTooltipProps)
       setHasFocus(isFocused)
     }
 
+    const onClick = () => {
+      // Close tooltip when element is clicked
+      setHasFocus(false)
+    }
```
rezrah, line 35: Can you pull these functions out of the useEffect and wrap in a useCallback?

packages/react/src/VideoPlayer/hooks/useIsElementFullScreen.ts
```diff
@@ -9,6 +9,12 @@ export const useIsElementFullScreen = (element?: HTMLElement | null): [boolean,
     const handleFullScreenChange = () => {
       const isCurrentlyFullScreen = document.fullscreenElement === element
 
+      if (isCurrentlyFullScreen) {
+        element.setAttribute('data-is-fullscreen', 'true')
```
rezrah, line 13: It seems like this is only used for styling purposes, is that right? If so, could you do this with a css class instead of a data attribute?

packages/react/src/VideoPlayer/VideoPlayer.tsx
```diff
@@ -57,91 +55,66 @@ const Root = ({
...
-
-      hideControlsTimeout = setTimeout(() => {
-        setIsInteracting(false)
-      }, inactivityTimeout)
-    }
-
-    const hideControls = () => {
-      setIsInteracting(false)
-    }
+  const {ccEnabled, fullscreenRef, isPlaying, ref, togglePlaying} = useVideoContext
```
rezrah, line 59: Just noticed this. Can we rename `useVideoContext` to something else? It reads like a hook (function) rather than a value. Could we rename this to something else like videoContext?

packages/react/src/VideoPlayer/VideoPlayer.tsx
```diff
@@ -57,91 +55,66 @@ const Root = ({
...
-          {showSeekControl && <SeekControl />}
-          {showCCButton && <CCButton />}
-          {showMuteButton && <MuteButton />}
-          {showVolumeControl && !isSmall && <VolumeControl />}
-          {showFullScreenButton && <FullScreenButton />}
-        </ControlsBar>
+          onClick={togglePlaying}
+          aria-label={isPlaying ? 'Pause' : 'Play'}
+        >
+          {!isPlaying && <PlayIcon />}
```
rezrah, line 95: Should this be ternary instead with a pause icon (or null) as the fallback?

.changeset/odd-paws-reply.md
```diff
@@ -0,0 +1,6 @@
+---
+'@primer/react-brand': patch
```
rezrah, line 2: Should be a minor as you're removing a prop

```suggestion
'@primer/react-brand': minor
```

.changeset/odd-paws-reply.md
```diff
@@ -0,0 +1,6 @@
...
+---
+
+- Improved `VideoPlayer` appearance on narrow viewports.
+- Removed `VideoPlayer` `showControlsWhenPaused` prop.
```
rezrah, line 6: Just curious, did we need to remove this as part of the accessibility update. Described as a UX improvement in PR description, but would like to understand how we comm this to users.

packages/react/src/VideoPlayer/VideoPlayer.module.css
joshfarrant: Thanks! Wasn't sure what to set these to, as per my comment in the description

packages/react/src/VideoPlayer/VideoPlayer.tsx
joshfarrant: No, this is the overlay that's used on paused videos. Adding a ternary with a pause icon here would result in a pause button being overlayed on top of playing videos.

packages/react/src/VideoPlayer/VideoPlayer.tsx
joshfarrant: Good suggestion, thanks 👍

.changeset/odd-paws-reply.md
joshfarrant: Great catch, thanks!

danielguillan (approved):
Looks great! Approving from a design perspective.

packages/react/src/VideoPlayer/hooks/useIsElementFullScreen.ts
joshfarrant: Yes no problem. I did consider that but didn't know if the data attribute would be preferable to having to use a `:global` CSS selector, but I'll switch it out now

.changeset/odd-paws-reply.md
joshfarrant: I didn't strictly need to remove it, but I think the component is better for not having it. In general I think limiting the modification of video controls (closed captions, volume, audio tracks when we implement them down the line) to only when the video is playing could be detrimental to some users.

How about

```md

- Improved `VideoPlayer` appearance on narrow viewports.
  - This necessitated the removal of the `showControlsWhenPaused` prop to prevent the height of the `VideoPlayer` from changing when the video is played/paused.
```

packages/react/src/VideoPlayer/components/VideoTooltip/VideoTooltip.tsx
joshfarrant: I can do, but I'd be curious to know what the benefit of that would be?

.changeset/odd-paws-reply.md
rezrah: 👍 LGTM, thanks

packages/react/src/VideoPlayer/components/VideoTooltip/VideoTooltip.tsx
rezrah: They don't need to be set in this useEffect, and will be recreated on each render if someone adds something to the dependency array in future

joshfarrant: I think that would actually be a performance _hit_ rather than optimisation. Regardless of whether the function is wrapped in a useCallback or not, it will be recreated on every render if it's in the top-level render method. 

> ["Note that useCallback does not prevent creating the function."](https://react.dev/reference/react/useCallback#should-you-add-usecallback-everywhere) — React docs

Having it in the useEffect means it will only be created when a dependency changes, which will always be less than or equal to the number of times the component renders.

I'm happy to make the change as I don't want to block this PR, but I'll add our usage of `useCallback` as a topic to the next maintainer sync as I think it would be worth clarifying the reasons we're using it.

packages/react/src/VideoPlayer/VideoPlayer.module.css
joshfarrant: @rezrah I've just spotted that this token isn't pure black (#1f2328).

My personal preference would be to use pure black here. Would you like me to stick with the token, or can I use #000000?

packages/react/src/VideoPlayer/components/VideoTooltip/VideoTooltip.tsx
rezrah: > Having it in the useEffect means it will only be created when a dependency changes

Yeah, so adding a dependency that _does_ change frequently will trigger the recreation of this function? That's what I'm alluding to . I appreciate it's an edge case, but pulling it outside and memoizing it should help there, no?

The React docs you shared says...

> Caching a function with useCallback  is only valuable in a few cases:

> The function you’re passing is later used as a dependency of some Hook. For example, another function wrapped in useCallback depends on it, _**or you depend on this function from [useEffect.](https://react.dev/reference/react/useEffect)**_

Unless I'm misunderstanding here, you're doing the latter ☝️

joshfarrant: We may be better off talking about this in Maintainer Sync as we'll be going into the weeds a bit here, but I think the crux of this is that `useCallback` doesn't prevent functions from being recreated every render.

[Kent C Dodds' article When to useMemo and useCallback](https://kentcdodds.com/blog/usememo-and-usecallback#why-is-usecallback-worse) does a good job of highlighting this in an intuitive way.

---

> Yeah, so adding a dependency that does change frequently will trigger the recreation of this function? That's what I'm alluding to . I appreciate it's an edge case, but pulling it outside and memoizing it should help there, no?

If we had a dependency which changed frequently then it would trigger the recreation of this function _regardless of whether the function is declared inside a useEffect or inside a useCallback_. I think that's the fundamental (albeit couterintuitive) thing to understand. Whatever we do, the function is going to be recreated every render, the only difference between the two is that the one wrapped in useCallback is also running whatever code runs inside the internals of useCallback every render too.

---

> The function you’re passing is later used as a dependency of some Hook. For example, another function wrapped in useCallback depends on it, or you depend on this function from useEffect.

In this scenario the function isn't _a dependency_ of useEffect, it's _created inside_ useEffect. That's why this doesn't apply here.

If we're considering functions which are a dependency of useEffect (ie functions which are explicitly listed in the dependencies array) then yes there is definite benefit to memoizing those with useCallback, however that's not what we have here. useEffect doesn't depend on this function, it creates it.

I think I'm also a bit confused here because we don't seem to do what you're suggesting here anywhere else in the codebase. I pulled up SubNav as an example and there are quite a few examples in there ([1](https://github.com/primer/brand/blob/af1b89ddaca8fab94c3e415ae70d45e85831223c/packages/react/src/SubNav/SubNav.tsx#L94-L96), [2](https://github.com/primer/brand/blob/af1b89ddaca8fab94c3e415ae70d45e85831223c/packages/react/src/SubNav/SubNav.tsx#L173-L180), [3](https://github.com/primer/brand/blob/af1b89ddaca8fab94c3e415ae70d45e85831223c/packages/react/src/SubNav/SubNav.tsx#L441-L444), [4](https://github.com/primer/brand/blob/af1b89ddaca8fab94c3e415ae70d45e85831223c/packages/react/src/SubNav/SubNav.tsx#L515-L523)) where we're declaring functions inside of a useEffect. I may well have missed examples of this, but from my quick look I've not found any cases where we pull functions which are only used inside a useEffect out into the top level of the component.

packages/react/src/VideoPlayer/VideoPlayer.module.css
joshfarrant: No it looks ok, if sticking to the token is preferable then we can go with that 👍

Both options below for transparency:

| Black | Off-black (design token) |
| --- | --- |
| Visual-Comparison-VideoPlayer-VideoPlayer-Without-Branding-1-linux | Visual-Comparison-VideoPlayer-VideoPlayer-Without-Branding-1-linux |

joshfarrant: Ah sorry, that's not _exactly_ right. The title overlay will need a pure black background to guarantee contrast with the content. The rest of it can be off-black though.

rezrah: 🤔 okay, sounds like you need a fixed `#000` value if it's a contrast issue. For context, at some point we need to introduce our primer css lint rules, and hexes will be flagged, so trying to reduce ones we introduce if we can avoid it. On this occasion, let's stick to hex with an annotation for future reference? Separately, we can look at making `black-0` an actual black.

packages/react/src/VideoPlayer/components/VideoTooltip/VideoTooltip.tsx
joshfarrant: Thanks for taking a look. Let's see if we can hop on a call before maintainer sync as we can probably avoid adding to the agenda and just chat it through between ourselves 👍

packages/react/src/VideoPlayer/hooks/useIsElementFullScreen.ts
```diff
@@ -9,6 +9,12 @@ export const useIsElementFullScreen = (element?: HTMLElement | null): [boolean,
     const handleFullScreenChange = () => {
       const isCurrentlyFullScreen = document.fullscreenElement === element
 
+      if (isCurrentlyFullScreen) {
+        element.classList.add('is-fullscreen')
```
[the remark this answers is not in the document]
joshfarrant: Ah, I forgot that it could be done that way, thanks 👍

PR #967, opened 2025-04-08
Add `isExternal` prop to Link component

danielguillan:
## Summary

Adds an `isExternal` prop to the `Link` component to display an external link icon. 

Solves https://github.com/github/primer/issues/4795

## List of notable changes:

- Added an `isExternal` prop to the Link component.
- Setting `isExternal` disables the `start` and `end` arrows.
- Refactored CSS slightly to manage arrows and external icons spacing together and ensure proper underline animation. 
  - As a result, this fixes an issue in `Pagination` where there was no spacing between the arrows and the labels on the next and previous buttons.

- [Link in Storybook](https://primer-e929cd9e4f-26139705.drafts.github.io/brand/storybook/?path=/story/components-link-features--external-link)
- [Pagination in Storybook](https://primer-e929cd9e4f-26139705.drafts.github.io/brand/storybook/?path=/story/components-pagination--default)
- No regressions on components that leverage Link styles

## Contributor checklist:

## Reviewer checklist:

## Screenshots:

image


packages/react/src/Card/Card.module.css
```diff
@@ -221,6 +221,10 @@
   z-index: -1;
 }
 
+.Card--expandableArrow {
+  margin-left: var(--base-size-4);
```
danielguillan, line 225: Since the gap between the label and the arrow plus external icon in `Link` is now managed at the parent level (i.e., `.Link`), and since `Card` links do not use those parent styles, we need to reintroduce the spacing here.

rezrah: Sounds good @danielguillan. Nit but could we use a logical property here instead?

```suggestion
  margin-inline-start: var(--base-size-4);
```

packages/react/src/Link/Link.module.css
```diff
@@ -1,13 +1,28 @@
 .Link {
-  --arrow-width: var(--base-size-20);
+  --visual-width: var(--external-icon-size, var(--arrow-size, 0));
```
rezrah, line 2: is the default supposed to be 0 here or var(--base-size-20)?

danielguillan: The default value is `0`. This allows us to set the underline width in one single place. So, when there is no visual (no arrow or external link icon), nothing is deducted from the total width (the result is `100%`).

PR #976, opened 2025-04-14
Update default Hero description text size

danielguillan:
## Summary

Bumps the Hero's description default text size to a new `350` step. Solves https://github.com/github/primer/issues/4944

## List of notable changes:

- Adds a new `350` token to the typographic scale.
- Adds a new `350` `size` option to the `Text` component.
- Updates the default text size for the `Hero` description to `350` and set its default font weight to "normal".
- Adds a new story for custom heading and description sizes

- Storybook

- Closes https://github.com/github/primer/issues/4944

## Contributor checklist:

## Reviewer checklist:

## Screenshots:

<table>
<tr>
<th> 
Before

</th> 

 <th> 
After
 </th>
</tr>
<tr>
<td valign="top">

image

 </td>
<td valign="top">

image

</td>
</tr>
</table>


packages/react/src/Hero/Hero.module.css
```diff
@@ -63,6 +63,11 @@
   max-width: 600px;
 }
 
+.Hero-description:not(.Hero-description--custom-size) {
```
danielguillan, line 66: I'm not a big fan of this. This selector increases specificity to override the default size set by the underlying `Text` component. Any other ideas for setting a default size _only when no `size` is provided_?

packages/react/src/Hero/Hero.module.css
```diff
@@ -63,6 +63,11 @@
...
+  font-size: var(--brand-text-subhead-size-large);
+  line-height: var(--brand-text-subhead-lineHeight-large);
```
danielguillan, line 68: Setting the size here, as `subhead` sizes are only available in `Heading`, not in `Text`.

rezrah: @danielguillan I think you should be able to add a breakpoint override for `text-300` so it matches subhead-large. It'll mean that you can use `size="300"` as the new default (instead of 200).

To clarify...

The base size for 300 and subhead-large area already the same:

Screenshot 2025-04-14 at 13 16 02

We're just missing the larger responsive value for 300, which seems like it would fit nicely at 1.25rem?

Screenshot 2025-04-14 at 13 16 08

I know this is wider change to the typescale, but I feel like that's where we _should_ be making this change. The current approach bypasses our type scale and sets a precedent for individually overriding text sizes in certain components. It also conflates the heading and text semantics like you pointed out.

joshfarrant (approved):

rezrah (approved):
Looks great, thanks @danielguillan

PR #993, opened 2025-05-02
Fix potentially confusing dropdown toggle accessible names

joshfarrant:
## Summary

Updating the both the aria-label and aria-expanded attributes of a component when it is expanded can potentially be confusing to screen reader users. This PR removes the aria-label updating from the `SubNav` and `ActionButton` components.

See the attached issue for more information.

- Closes https://github.com/github/primer/issues/4679
- Closes https://github.com/github/accessibility-audits/issues/10399

## Contributor checklist:

## Reviewer checklist:

 </td>
<td valign="top">

</td>
</tr>
</table>


packages/react/src/ActionMenu/ActionMenu.tsx
```diff
@@ -392,9 +392,9 @@ const ActionMenuButton = forwardRef<HTMLButtonElement, ActionMenuButtonProps>(
             variant={variant}
             hasArrow={false}
             aria-haspopup="true"
-            aria-label={`${menuOpen ? 'Close' : 'Open'} menu`}
+            aria-label="Action menu"
```
rezrah, line 395: While "Action menu" is our internal UI component name, it shouldn't be announced on our marketing pages this way... Can we simply call this a "Menu"?

packages/react/src/ActionMenu/ActionMenu.tsx
```diff
@@ -476,8 +476,8 @@ const ActionMenuItem = ({
         className,
       )}
       role={roleTypeMap[type || 'single']}
-      aria-checked={as === 'a' || type === 'none' ? undefined : selected ? 'true' : 'false'}
-      aria-disabled={disabled ? 'true' : 'false'}
+      aria-checked={as === 'a' || type === 'none' ? undefined : selected}
+      aria-disabled={disabled}
```
rezrah, line 480: Was this refactoring relevant to resolving the aria-checked / aria-expanded issue? It seems to be minor and potentially noop, but can we still capture these in changesets. You can call it out as general internal refactoring to other component props like `aria-disabled`. Helps us all keep an eye out for regressions and prevent indirection because current changeset implies the update only affects aria-checked and aria-expanded.

joshfarrant: Just minor tidying to, imo, help with readability.

Updated the changeset 👍

packages/react/src/SubNav/SubNav.tsx
```diff
@@ -455,7 +455,7 @@ const SubNavLinkWithSubmenu = forwardRef<HTMLDivElement, SubNavLinkProps>(
             onClick={toggleExpanded}
             aria-expanded={isExpanded ? 'true' : 'false'}
             aria-controls={submenuId}
-            aria-label={`${isExpanded ? 'Close' : 'Open'} submenu`}
+            aria-label={`${label} submenu`}
```
rezrah: Could you add a trim() to the label to remove unintended space at the end. Currently there's a double space in the interpolated string value. E.g.

Screenshot 2025-06-05 at 11 33 34

packages/react/src/ActionMenu/ActionMenu.tsx
```diff
@@ -392,7 +392,7 @@ const ActionMenuButton = forwardRef<HTMLButtonElement, ActionMenuButtonProps>(
             variant={variant}
             hasArrow={false}
             aria-haspopup="true"
-            aria-label={`${menuOpen ? 'Close' : 'Open'} menu`}
+            aria-label="Menu"
```
rezrah, line 396: You'll need to add this to line 414 too, as this change will only apply to split button mode?

joshfarrant: I may be misunderstanding your comment, but I believe the non-split button below doesn't need an `aria-label` because it's already labelled by the `children` passed in, e.g. "Copilot".

packages/react/src/SubNav/SubNav.tsx
joshfarrant: Added 👍

packages/react/src/ActionMenu/ActionMenu.tsx
rezrah: Ah sorry, yes you're right I was looking at a different aria prop, nevermind. This looks good 👍

PR #1001, opened 2025-05-08
Add subheading to SubNav component

rezrah:
## Summary

Closes https://github.com/github/primer/issues/5147

Adds a new `SubNav.SubHeading` child to the `SubNav` component for deeper IA support.

🔗 [Storybook preview link](https://primer-cce331d353-26139705.drafts.github.io/brand/storybook/?path=/story/components-subnav-features--sub-heading&globals=colorMode:dark)
🔗 [Docs preview](https://primer-cce331d353-26139705.drafts.github.io/brand/components/SubNav#subheading)

Before: Parent => Child, Child, Child
Now: Parent => Child => Child, Child, Child

No breaking changes to API, existing functionality or visual presentation. 

## List of notable changes:

- Adds support for  `SubNav.SubHeading` 
- Adds Storybook examples
- Adds visual test coverage

- Please verify that the new feature appear as intended in BOTH light and dark color modes
- There is no design available for when SubNav.SubHeading is active on mobile, so I've retained the button appearance as before. [Example in Storybook covering this path.
](https://primer-cce331d353-26139705.drafts.github.io/brand/storybook/?path=/story/components-subnav-features--active-sub-heading-narrow&globals=colorMode:dark;viewport:iphonex)

## Steps to test:

1. [Storybook examples](https://primer-cce331d353-26139705.drafts.github.io/brand/storybook/?path=/story/components-subnav-features--sub-heading&globals=colorMode:dark)

## Contributor checklist:

## Reviewer checklist:

## Screenshots:

Screenshot 2025-05-08 at 12 40 44


simmonsjenna (approved):
@rezrah This looks good to me in both dark and light mode and across device sizes. The only question I had is: does the lack of side padding in the Storybook examples affect this? I noticed that the examples before and after the new Subheading examples have visible side padding so I was curious if that was a bug or not.

rezrah:
> @rezrah This looks good to me in both dark and light mode and across device sizes. The only question I had is: does the lack of side padding in the Storybook examples affect this, or is that provided by its container? I noticed that the examples before and after the new Subheading examples have visible side padding so I was curious if that was a bug or not.

Thanks for spotting that @simmonsjenna. I mistakenly left the `fullWidth` option enabled. After removing it, the default padding has been restored. Quick before/afters:

Screenshot 2025-05-08 at 16 17 21

packages/react/src/SubNav/SubNav.tsx
```diff
@@ -162,6 +191,13 @@ const _SubNavRoot = memo(({id, children, className, 'data-testid': testId, fullW
     }
   }, [isOpenAtNarrow, isLarge])
 
+  // On narrow viewports the focus should be retained on the button when menu is open
+  useEffect(() => {
+    if (!isLarge && narrowButtonRef.current && isOpenAtNarrow) {
+      narrowButtonRef.current.focus()
+    }
+  }, [isOpenAtNarrow, isLarge])
```
joshfarrant, line 199: I think this may not be working as expected as, when the dropdown is collapsed, focus is lost and so returns to the start of the document.

To reproduce, open the narrow-viewport menu using the keyboard, hit <kbd>Escape</kbd> to close it, press tab, then observe that focus now returns to the start of the document.

One way to resolve this is as follows.

```suggestion
  useEffect(() => {
    if (!isLarge && narrowButtonRef.current) {
      narrowButtonRef.current.focus()
    }
  }, [isOpenAtNarrow, isLarge])
```

However I think the fact that we need this code block in the first place is actually indicative of another issue, because removing it entirely causes the button to lose focus every time the menu expands or collapses, which shouldn't naturally happen.

I believe the issue is being caused by the `useCallback` on L264. The callback is recomputed every render (probably due to `activeLinkLabel` changing) and so the button is being recreated each render, meaning that when the `_SubNavRoot` component re-renders when the dropdown expands/collapses, the button is destroyed and recreated, causing it to lose focus.

The easiest way to fix this is just to not `useCallback` the button and instead just store the JSX in a variable, eg:

```tsx
const narrowButton = (
  <button
    ref={narrowButtonRef}
    className={clsx(styles['SubNav__overlay-toggle'], isOpenAtNarrow && styles['SubNav__overlay-toggle--open'])}
    {/* etc */}
  >
    {/* ...*/}
  </button>
)
```

Which you can then use below as

```tsx
{!isLarge && (!SubHeadingChild || subHeadingIsActive) && narrowButton}

// and

{!isLarge && SubHeadingChild && !subHeadingIsActive && narrowButton}
```

This will stop the button being torn down and recreated every render, and allow you to remove the `useEffect` in favour of adding `narrowButtonRef.current?.focus()` on L176

packages/react/src/SubNav/SubNav.test.tsx
```diff
@@ -200,4 +203,21 @@ describe('SubNav', () => {
...
+    const expectedLink = '#subheading'
+    const {getByText} = render(
+      <SubNav>
+        <SubNav.Heading href={headingLink}>{heading}</SubNav.Heading>
+        <SubNav.SubHeading href={expectedLink}>{expectedText}</SubNav.SubHeading>
+        <SubNav.Link href="#">Link</SubNav.Link>
+      </SubNav>,
+    )
+
+    const el = getByText(expectedText)
```
joshfarrant, line 218: Very minor, but the following will also check that the element is a link


```suggestion
    const {getByRole} = render(
      <SubNav>
        <SubNav.Heading href={headingLink}>{heading}</SubNav.Heading>
        <SubNav.SubHeading href={expectedLink}>{expectedText}</SubNav.SubHeading>
        <SubNav.Link href="#">Link</SubNav.Link>
      </SubNav>,
    )

    const el = getByRole('link', {name: expectedText})
```

packages/react/src/SubNav/SubNav.test.tsx
```diff
@@ -200,4 +203,21 @@ describe('SubNav', () => {
...
+
+    expect(el).toBeInTheDocument()
+    expect(el).toHaveAttribute('href', expectedLink)
+  })
 })
```
joshfarrant, line 223: If you get chance, it would be great to have a test which checks that focus is returned to the button on narrow submenu close too

packages/react/src/SubNav/SubNav.features.stories.tsx
```diff
@@ -92,6 +92,98 @@ WithShadow.parameters = {
...
+export const SubHeadingNarrowOpen = args => <SubHeading {...args} />
+SubHeadingNarrowOpen.parameters = {
+  layout: 'fullscreen',
+  viewport: {
+    defaultViewport: 'iphonex',
+  },
+}
+SubHeadingNarrowOpen.play = async ({canvasElement}) => {
+  const canvas = within(canvasElement)
+  await userEvent.click(canvas.getByTestId('SubNav-root-button'))
```
joshfarrant, line 138: Really minor and appreciate this is more of a demo than a test, but it would be nice to use a `getByRole` in these cases, rather than adding more `getByTestId` uses to the codebase.

rezrah: Good suggestion. Will defer that for now given the urgency for this to ship.

packages/react/src/SubNav/SubNav.test.tsx
rezrah: 👍 Added

packages/react/src/SubNav/SubNav.tsx
rezrah: Good suggestion. I was scratching my head about this one, and went with the latter option with added memoization. Works as before now.

danielguillan (approved):
Looks good!

PR #1009, opened 2025-05-14
Fix incorrect role and focus order in Pagination component

joshfarrant:
## Summary

Fixed two reported bugs in the Pagination component, as well as an additional bug uncovered during the course of this work.

## List of notable changes:

This may look like more of a refactor than it is. Check individual commits for more info, but the key changes are: 

- Renamed the `usePaginationPages` hook to `PaginationItem` as it was behaving more like a component than a hook; it accepted a props object and returned JSX
- Unified the attribute calculation and component rendering logic by removing `buildComponentData` from `model.ts` and instead put that logic inside the `PaginationItem` component. 
- Updated stories to allow controls to be used across all stories
  - While not necessary for this work, this change makes it easier for the component to be manually tested in Storybook.
- Added a load of new tests to better assert the expected behaviour of the component. 

The majority of the key logic hasn't changed, it's just been moved around and simplified slightly.

<details>
  <summary>Bonus HTML!</summary>
  
I also grabbed the rendered HTML from the Pagination component before and after these changes and dropped them into a couple of quick diffs so you can more easily see the changes to the rendered markup 

- [Pagination — Page 4 of 5](https://www.diffchecker.com/SlNudCxr/)
- [Pagination — Page 100 of 100](https://www.diffchecker.com/V46WMCpp/)

Note the following changes:

- Previous & Next buttons both now have `role="button"`
- `tabindex="0"` has been removed from all elements as it's not necessary since we're using the browser's default tab order. This also resolves https://github.com/github/primer/issues/3734 by removing `tabindex` from the ellipsis
- `as="span"` has been removed from the ellipsis element. This was leaking through to the DOM due to `as="span"` not being a valid prop on our `Link` component. The previously described refactor and improved typing will prevent this category of issues from happening in the future.
- When the Previous & Next buttons were disabled, their `aria-label` was removed. This PR also resolves that bug.

</details>

- Test the component and make sure it works as expected.
- I've added tests to cover this, but check all of the expected attributes are still in the right place. 

## Steps to test:

I've added controls to all [Pagination stories](https://primer-30a8808ad5-26139705.drafts.github.io/brand/storybook/?path=/story/components-pagination--playground), so poke around in Storybook to check everything.

- Closes https://github.com/github/primer/issues/3734
- Closes https://github.com/github/primer/issues/3735
- Fixes an additional bug discovered during the course of this work where the Previous/Next buttons would lose their `aria-label` when disabled.

## Contributor checklist:

## Reviewer checklist:


packages/react/src/Pagination/model.ts
```diff
@@ -135,75 +129,7 @@ export function buildPaginationModel(
...
-        'aria-label': `Page ${page.num}${page.precedesBreak ? '...' : ''}`,
-        onClick,
-        'aria-current': page.selected ? 'page' : undefined,
-      })
-      break
-    }
-    case 'BREAK': {
-      key = `page-${page.num}-break`
-      content = '…'
-      Object.assign(props, {as: 'span', role: 'presentation'})
```
joshfarrant, line 204: In the previous implementation, this `as` wasn't working as expected. The element was still rendered as an `<a>` despite the `as` prop here.

Note the rendered HTML:

image

This contributed (along with the tabIndex set [here](https://github.com/primer/brand/pull/1009/files?diff=split&w=0#diff-4242d1a3dcad5607a6d5c3858cbc7b71b628f7962e2ac0d269ecc5d040eaf483L159)) to them still being focusable, and causing https://github.com/github/primer/issues/3734

packages/react/src/Pagination/Pagination.test.tsx
```diff
@@ -107,4 +135,132 @@ describe('Pagination', () => {
...
+    expect(getByRole('button', {name: 'Page 3'})).not.toHaveAttribute('aria-disabled')
+    expect(getByRole('button', {name: 'Page 4'})).not.toHaveAttribute('aria-disabled')
+    expect(getByRole('button', {name: 'Page 5...'})).not.toHaveAttribute('aria-disabled')
+    expect(getByRole('button', {name: 'Next Page'})).not.toHaveAttribute('aria-disabled')
+  })
+
+  it('sets "aria-disabled" to true for just the "Next" button when on the last page', () => {
+    const {getByRole} = render(<Pagination pageCount={5} currentPage={5} />)
+
+    expect(getByRole('button', {name: 'Previous Page'})).not.toHaveAttribute('aria-disable')
```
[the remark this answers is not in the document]
joshfarrant: :copilot: 💖

packages/react/src/Pagination/model.ts
```diff
@@ -1,11 +1,5 @@
-import {LinkProps} from '../Link'
-
-type PaginationModel =
-  | {type: string; num: number}
-  | {type: string; num: number; selected: boolean; precedesBreak: boolean}
-
-type PageType = {
-  type: string
+export type PageType = {
```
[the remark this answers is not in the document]
joshfarrant: https://github.com/primer/brand/blob/31fe3e8f5a41917bf8b4702406fc77e7a80aaae0/packages/react/src/Pagination/model.ts#L6

Come on now :copilot:, keep up

packages/react/src/Pagination/Pagination.stories.tsx
```diff
@@ -1,38 +1,33 @@
...
-        const target = e.target as HTMLElement
-        // eslint-disable-next-line no-console
-        console.log('pressed: ', target.textContent)
-      }}
-      {...args}
-    />
-  </main>
-)
-
-export const Default = Template.bind({})
```
rezrah, line 27: The default story seems to be have been removed. Was that intended @joshfarrant?

packages/react/src/Pagination/Pagination.test.tsx
```diff
@@ -53,6 +54,20 @@ describe('Pagination', () => {
     expect(linksAsVerbatimText).toEqual(['Previous', '1', '…', '3', '4', '5', '6', '7', '…', '10', 'Next'])
   })
 
+  it('applies the correct attributes to the ellipsis button', () => {
+    const {container} = render(<Pagination pageCount={10} currentPage={1} />)
+
+    // Testing library can't find this element as it's presentation-only, so we need to use querySelector
```
rezrah: Could you elaborate on this part please? Wondering why Testing Library can't see it, but a query selector can. They should be capable of doing similar things. 

Did `getByRole('presentation')` or `queryByRole('presentation')` not work?

packages/react/src/Pagination/Pagination.stories.tsx
joshfarrant: I may have removed this as it was identical to the Playground story. I can add it back in through if you prefer?

rezrah: Yes please, FYI going forward it's a cross-Primer convention to have a default story, so they shouldn't be removed. It should be similar to our default (baseline) example in public docs: https://primer.style/brand/components/Pagination#default

packages/react/src/Pagination/Pagination.test.tsx
joshfarrant: It was a while back so I'm not 100% sure why I did this 😅

I've just done some testing and I think that I'd (incorrectly) assumed that testing library would ignore elements with a role of "presentation" in the same way it ignores hidden elements.

That doesn't seem to be the case though, so I'll swap this out for a `getByRole`, thanks!

joshfarrant: I may have tried selecting it using `getByRole('presentation', {name: '…'})` which _doesn't_ work, so possibly that's where my incorrect assumption came from

packages/react/src/Pagination/Pagination.stories.tsx
joshfarrant: That's added back in 👍

joshfarrant: Actually, would you prefer that I align it with the example in our docs as it currently doesn't match

```tsx
export const Default: Story = {
  args: {
    pageCount: 10,
    currentPage: 5,
  },
}
```

rezrah: 👍 sounds good. Default examples are just intended to show the minimum set of arguments required to use the pattern. Playground can be more opinionated.

rezrah:
I can see a change to the visual snapshot but wonder if that should have changed here? 🤔 Can't leave an inline comment so FYI @joshfarrant

PR #1029, opened 2025-05-22
Intentionally invalidate a snapshot (demo PR, don't merge)

joshfarrant:
Here I've intentionally invalidated a snapshot by manually replacing it with a different image.

Note that [the VRT comment below](https://github.com/primer/brand/pull/1029#issuecomment-2901520389) incorrectly states:

> 🟢 No visual differences found

The snapshot I manually replaced _should_ be flagged as invalid, and so we should be getting the following message:

> ⚠️ Visual differences found


joshfarrant:
Closing as this is a demo PR and associated fix has been merged.

Issue #936, opened 2025-03-04
[Feature Request] Hero to support VideoPlayer and iframe embed sources


joshvogel17:
We would like to enhance the Hero to be able to support having videos, either self hosted content like an mp4 or from a 3rd party source to render in an iframe. Ex. Youtube.

Currently in SWP we support embedding Youtube links in an iframe, but our player is outside of the Hero component and styled to look like it is in the Hero. As a result, we can only render the video player at the bottom of the Hero and not inline with the Hero Content. An example of this is the https://github.com/features/copilot/copilot-business page.

We would like to include the video player or iframe embed (depending on source) into the Primer Brand Hero component for variant control options as well as better brand and style fidelity.

PR #1042, opened 2025-06-05
Add video support to Hero component

rezrah:
## Summary

Resolves https://github.com/primer/brand/issues/936

Adds a new slot for inserting custom videos / Youtube embeds

🔗 Docs preview
🔗 Youtube example
🔗 VideoPlayer example
🔗 `<video>` element example

## List of notable changes:

- Adds a new slot for inserting video elements in a similar manner to images

- Reviewing the layout and usage examples for correctness

## Steps to test:

1. Review links at the top of the description
1. Check tests are passing

## Contributor checklist:

## Reviewer checklist:

## Screenshots:

image


packages/react/src/Hero/Hero.tsx
```diff
@@ -35,58 +49,68 @@ const Root = forwardRef<HTMLElement, PropsWithChildren<HeroProps>>(
...
     const heroLayoutClass = HeroImageChild ? styles['Hero--layout-image'] : styles['Hero--layout-default']
     return (
       <section
         className={clsx(
           styles.Hero,
-          imagePosition !== 'inline-end' && styles[`Hero--align-${align}`],
+          mediaPosition !== 'inline-end' && styles[`Hero--align-${align}`],
           heroLayoutClass,
-          HeroImageChild && styles[`Hero--image-pos-${imagePosition}`],
+          (HeroImageChild || HeroVideoChild) && styles[`Hero--image-pos-${mediaPosition}`],
```
joshfarrant, line 98: I might be missing something here, but is the class `Hero--image-pos-...` defined anywhere?

packages/react/src/Hero/Hero.tsx
```diff
@@ -179,6 +212,28 @@ const HeroImage = forwardRef<HTMLImageElement, HeroImageProps>(
   },
 )
 
+type HeroVideoProps = {
+  position?: 'inline-end' | 'block-end'
+  'data-testid'?: string
+} & PropsWithChildren<BaseProps<HTMLDivElement>>
+
+const HeroVideo = forwardRef<HTMLDivElement, HeroVideoProps>(
```
joshfarrant, line 220: Appreciate that we probably want to avoid an API change here, but having both `HeroImage` and `HeroVideo` feels a but duplicative.

From what I can tell, they look to be treated in exactly the same way, and we even have to have additional code to handle cases where both are provided.

What would you think to instead removing `Hero.Image` and renaming `Hero.Video` to `Hero.Media`?

I believe consumers would then be able to replace usage of `Hero.Image` in the following way:

```tsx
<Hero.Image position="inline-end" src={placeholderImage} alt="placeholder image, blank with gray solid fill" />
```

with

```tsx
<Hero.Media position="inline-end">
  <Image src={placeholderImage} alt="placeholder image, blank with gray solid fill" />
</Hero.Media>
```

And videos can be provided in the same way you've already described in this PR, just replacing `Hero.Video` with `Hero.Media`

```tsx
<Hero.Video>
  <VideoPlayer title="GitHub media player" poster={posterImage}>
    <VideoPlayer.Source src="./example.mp4" type="video/mp4" />
    <VideoPlayer.Track src="./example.vtt" default />
  </VideoPlayer>
</Hero.Video>
```

Again, appreciate that you probably want to avoid API changes, however I think this approach simplifies the implementation, keeps things flexible, and gives us one fewer components to maintain going forward.

wdyt?

apps/docs/scripts/components-with-animation.js
```diff
@@ -10,8 +10,8 @@ export const supportedComponents = [
   'Label',
   'Pillar',
   'SectionIntro',
-  'Stack',
   'Statistic',
+  'Stack',
```
danielguillan, line 14: No clue why this happens from time to time 🤷

danielguillan (approved):
Looking great!

apps/docs/scripts/components-with-animation.js
rezrah: Yeah, I kind of know why but haven't got around to fixing it yet in the build script 😬... leave it with me, and thanks for the reminder. It's a noop file change FWIW.

packages/react/src/Hero/Hero.tsx
joshfarrant: Appreciate that the class wasn't added by your change, but this might be a good opportunity to remove it?

packages/react/src/Hero/Hero.tsx
joshfarrant: I wanted to test this quickly so I span up a fork of your branch with this change applied.

https://github.com/primer/brand/pull/1043

It's a fairly minor change that consumers will need to make (rename `Hero.Image` to `Hero.Media`, add in `Image` as a child), and when compared to the simplification and consistency we get in return within `Hero` then, imo, it's a good trade-off

rezrah: Thanks for the suggestion Josh. I'm opting against making a breaking change here. I intended for `Hero.Image` and `Hero.Video` to behave differently by design (latter being a slot) to give us more flexibility and isolation. I'm not a huge fan of catch-all slots like `X.Media`, as they increase risk of becoming a downstream escape-hatch. They are different content types and while they have visual parity now, this is coincidental and may not always be the case in future.

packages/react/src/Hero/Hero.tsx
rezrah: Good spot, but going to punt this as I want to unblock this and avoid going out-of-scope in this PR. Will look into it in a follow-up.

PR #1051, opened 2025-06-16
Allow VideoPlayer tooltips to be hovered

joshfarrant:
## Summary

`VideoPlayer` tooltips no longer dismiss immediately and instead allow a small window of time in which the cursor can be moved from the trigger to the tooltip, at which point the tooltip will remain visible until the cursor leaves the tooltip.

- Check that the hovering works as expected

- Closes https://github.com/github/primer/issues/3802

## Contributor checklist:

## Reviewer checklist:

### DIFF
--- a/.changeset/heavy-turtles-accept.md
+++ b/.changeset/heavy-turtles-accept.md
@@ -2,4 +2,4 @@
 '@primer/react-brand': patch
 ---
 
-Tooltips in the `VideoPlayer` component are now hoverable.
+Tooltips in the `VideoPlayer` component now remain briefly visible after the pointer is moved away from the toggle. This small delay improves general usability and helps meet the [WCAG 1.4.13 criterion](https://www.w3.org/TR/WCAG22/#content-on-hover-or-focus) for accessible hover interactions with tooltips.
--- a/.changeset/heavy-turtles-accept.md
+++ b/.changeset/heavy-turtles-accept.md
@@ -1,0 +1,5 @@
+---
+'@primer/react-brand': patch
+---
+
+Tooltips in the `VideoPlayer` component are now hoverable.
--- a/packages/react/src/VideoPlayer/components/VideoTooltip/VideoTooltip.tsx
+++ b/packages/react/src/VideoPlayer/components/VideoTooltip/VideoTooltip.tsx
@@ -9,6 +9,7 @@ type VideoTooltipProps = HTMLAttributes<HTMLDivElement>
 export const VideoTooltip = ({children, className, ...rest}: VideoTooltipProps) => {
   const tooltipRef = useRef<HTMLDivElement>(null)
   const [hasFocus, setHasFocus] = useState(false)
+  const mouseenterTimeoutRef = useRef<NodeJS.Timeout | null>(null)
 
   useEffect(() => {
     const tooltip = tooltipRef.current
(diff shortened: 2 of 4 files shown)


.changeset/heavy-turtles-accept.md
```diff
@@ -0,0 +1,5 @@
+---
+'@primer/react-brand': patch
+---
+
+Tooltips in the `VideoPlayer` component are now hoverable.
```
rezrah: Let's tweak for clarity, as tooltips were previously hoverable but just not meeting specific usability/accessibility requirements. 

```suggestion
Tooltips in the `VideoPlayer` component now remain briefly visible after the pointer is moved away from the toggle. This small delay improves general usability and helps meet the [WCAG 1.4.13 criterion](https://www.w3.org/TR/WCAG22/#content-on-hover-or-focus) for accessible hover interactions with tooltips.
```

packages/react/src/VideoPlayer/components/VideoTooltip/VideoTooltip.tsx
```diff
@@ -34,11 +35,30 @@ export const VideoTooltip = ({children, className, ...rest}: VideoTooltipProps)
       setHasFocus(false)
     }
 
+    const onMouseEnterTooltip = () => {
+      if (mouseenterTimeoutRef.current) {
+        clearTimeout(mouseenterTimeoutRef.current)
```
rezrah: This is probably sufficient, but should we also do a clearTimeout in the useEffect cleanup too just in case?

packages/react/src/VideoPlayer/components/VideoTooltip/VideoTooltip.tsx
```diff
@@ -34,11 +35,30 @@ export const VideoTooltip = ({children, className, ...rest}: VideoTooltipProps)
...
+    }
+
+    const onMouseLeaveTooltip = () => {
+      if (mouseenterTimeoutRef.current) {
+        clearTimeout(mouseenterTimeoutRef.current)
+      }
+
+      mouseenterTimeoutRef.current = setTimeout(() => {
+        setHasFocus(false)
+      }, 100)
```
rezrah, line 55: `100` feels nice to me, as it's not too noticeable but I am wondering if this is sufficient. WCAG don't specify this, but did a11y team indicate how long the tooltip should linger before it disappears? It's currently so subtle I wouldn't notice it unless someone pointed it out to me. 

https://github.com/user-attachments/assets/279976e1-d446-472d-8cfc-6d359aed7ba2

rezrah (approved):
LGTM. Some test coverage would be nice if that's possible @joshfarrant? WDYT?

.changeset/heavy-turtles-accept.md
joshfarrant: I've updated the changeset 👍 Just to clarify though, the tooltips themselves weren't previously hoverable. They would show when the associated control was hovered, but the tooltips couldn't be hovered as they would be dismissed before you could get the mouse over to them.

packages/react/src/VideoPlayer/components/VideoTooltip/VideoTooltip.tsx
joshfarrant: Yes good suggestion, thank you.

I've also made sure that I set `mouseenterTimeoutRef.current = null` when clearing the timeout too, just to be thorough in the cleanup

packages/react/src/VideoPlayer/components/VideoTooltip/VideoTooltip.tsx
joshfarrant: > It's currently so subtle I wouldn't notice it unless someone pointed it out to me.

I think that's a feature more than an issue tbh 

The ultimate goal of this PR isn't to make the tooltips stick around a bit longer after having moved the mouse away — instead the goal is to enable users to move their cursor from the control to the associated tooltip.

If you try this on the current `VideoPlayer` you'll notice that the tooltip is dismissed before you can get the cursor over to it. The 100ms timer here is to give users a brief window of time to have their cursor cross the gap of a few pixels between the control and the tooltip.

There's no guidance on how long this window should be, it's just down to us to find a time which is reasonable.

PR #1053, opened 2025-06-17
Update heading levels in the `FAQ` component

joshfarrant:
## Summary

Updated heading levels in the `FAQ` component.

- Updated default `FAQ.Subheading` heading level to `h4`.
- `FAQ.Question` now renders as a `h5` if there is a `FAQ.Subheading` present.

This change also necessitated updates to the `FAQGroup` component, which has been updated to respect all provided `FAQ.Heading` props, not just children.

This change also required updating the `FeaturePreviewLevelTwo` playground to correct the now-invalid `FAQ.Question` heading levels. Consumers using the `FAQ` component with default settings won't need to modify their code — only those who have customised the default heading levels will need to make adjustments.

- Check that the changes work as expected and that the updated heading structure is logical.

- Closes https://github.com/github/primer/issues/3827

## Contributor checklist:

## Reviewer checklist:

## Screenshots:

<table>
<tr>
</tr>
<tr>
<td valign="top">

image

 </td>
<td valign="top">

image

</td>
</tr> 

<tr>
<td valign="top">

image

 </td>
<td valign="top">

image

</td>
</tr>

</table>

### DIFF
--- a/packages/react/src/FAQ/FAQ.test.tsx
+++ b/packages/react/src/FAQ/FAQ.test.tsx
@@ -248,4 +248,28 @@ describe('FAQ', () => {
     expect(subheadingEl).toBeInTheDocument()
     expect(questionheadingEl).toBeInTheDocument()
   })
+
+  it('renders the heading as a h3, the subheading as a h4, and the question as a h5 when there is a subheading present, FAQ.Items are wrapped in a Fragment, and all FAQ.Items are wrapped in an array', () => {
+    const {getByRole} = render(
+      <FAQ>
+        <FAQ.Heading>{mockHeading}</FAQ.Heading>
+        <FAQ.Subheading>{mockSubheading}</FAQ.Subheading>
+        <>
+          {[1, 2].map(item => {
+            return (
+              <FAQ.Item key={`item-${item}`}>
+                <FAQ.Question>Question {item}</FAQ.Question>
+                <FAQ.Answer>Answer {item}</FAQ.Answer>
+              </FAQ.Item>
+            )
+          })}
+        </>
+      </FAQ>,
+    )
+
+    expect(getByRole('heading', {level: 3, name: mockHeading})).toBeInTheDocument()
+    expect(getByRole('heading', {level: 4, name: mockSubheading})).toBeInTheDocument()
+    expect(getByRole('heading', {level: 5, name: 'Question 1'})).toBeInTheDocument()
+    expect(getByRole('heading', {level: 5, name: 'Question 2'})).toBeInTheDocument()
+  })
 })
--- a/packages/react/src/FAQ/FAQ.tsx
+++ b/packages/react/src/FAQ/FAQ.tsx
@@ -22,19 +22,30 @@ export type FAQRootProps = PropsWithChildren<BaseProps<HTMLElement>> & React.HTM
 const FAQRoot = forwardRef<HTMLElement, FAQRootProps>(({children, style, animate, className, ...rest}, ref) => {
   const {classes: animationClasses, styles: animationInlineStyles} = useAnimation(animate)
 
-  const filteredChildren = React.Children.toArray(children).filter(child => {
-    if (React.isValidElement(child) && typeof child.type !== 'string') {
-      if (
-        isFragment(child) ||
-        (child as React.ReactElement).type === FAQHeading ||
-        (child as React.ReactElement).type === FAQSubheading ||
-        (child as React.ReactElement).type === AccordionRoot
-      ) {
-        return true
+  const filteredChildren = React.Children.toArray(children)
+    .filter(child => {
+      if (React.isValidElement(child) && typeof child.type !== 'string') {
+        if (
+          isFragment(child) ||
+          (child as React.ReactElement).type === FAQHeading ||
+          (child as React.ReactElement).type === FAQSubheading ||
+          (child as React.ReactElement).type === AccordionRoot
+        ) {
+          return true
+        }
       }
-    }
-    return false
-  })
+      return false
+    })
+    .map(childMaybeFragment => {
+      /**
+       * If children are wrapped in a Fragment, extract the children.
+       * Children may also be an array of FAQ.Item components, so we call
+       * `toArray` to ensure we always have an array.
+       */
+      return React.Children.toArray(
+        isFragment(childMaybeFragment) ? childMaybeFragment.props.children : childMaybeFragment,
+      )
+    })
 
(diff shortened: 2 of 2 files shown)


rezrah:
> This change also necessitated updates to the FAQGroup component

@joshfarrant As you're changing FAQGroup here, could we double check the levels are correct in grouped mode please? 

E.g. I'm looking at this: https://primer-875bfd87e2-26139705.drafts.github.io/brand/components/FAQ/react#grouped-faqs

These don't have subheadings, but the questions should still follow the same guidance as...

> FAQ.Question now renders as a h5 if there is a FAQ.Subheading present.

image

joshfarrant:
Thanks @rezrah 👍

Added in https://github.com/primer/brand/pull/1053/commits/1417dbf1e9953fb3a31849462bf6aa0efd8354fb

rezrah:
Could you update the default tags in the prop tables for each docs site too please? Currently showing outdated values (and in some cases nothing at all - let's backfill). 

Screenshot 2025-06-24 at 14 20 52

.changeset/purple-plants-search.md
```diff
@@ -0,0 +1,9 @@
+---
+'@primer/react-brand': patch
+---
+
+Improved heading level structure in the `FAQ` component.
```
rezrah: To offer immediate clarity that this won't affect user overrides

```suggestion
Improved default heading levels in the `FAQ` component.
```

.changeset/purple-plants-search.md
```diff
@@ -0,0 +1,9 @@
...
+
+- Updated default `FAQ.Subheading` heading level from `h3` to `h4`.
```
rezrah:
```suggestion
- Updated `FAQ.Subheading` heading level from `h3` to `h4`.
```

.changeset/purple-plants-search.md
```diff
@@ -0,0 +1,9 @@
...
+- `FAQ.Question` now renders as a `h5` if there is a `FAQ.Subheading` present, or if rendered inside a `FAQGroup`.
+- `FAQGroup` now respects all provided `FAQ.Heading` props, not just children.
```
rezrah, line 9: @joshfarrant could there be a possible regression when users upgrade? Whereby they added workarounds for our incorrect leveling, which will no longer work now that we've fixed it on our end? If that's a possibility, we should issue an update here advising them to review usage of FAQ for correct leveling, and that prior workarounds should be removed.

joshfarrant:
@rezrah Thank you, sorry for missing that.

I've updated the `FAQ.Subheading` default `as` in the docs here: https://github.com/primer/brand/pull/1053/commits/059a6d0166c78d04884a628bcb0e8b8ef99646ab

I'm not sure there is a value which makes sense to add as the default value for `FAQ.Question` as it depends on whether there is a subheading present. If there is then it's `h5`, if there's not then it's `h4`. Are you happy to just leave the default value for `FAQ.Question` blank in the docs for now?

.changeset/purple-plants-search.md
joshfarrant: That's a good question

- `FAQ.Subheading` and `FAQ.Question` — If anyone has explicitly set an `as` on either of these then they won't be affected, only those who haven't specified `as` will see a change. We've already called this out in the changelog so I think that's a reasonable level of warning from us.
- `FAQ.Heading` inside `FAQGroup` — I'm not sure that anyone _could_ have worked around this before, as previously an `FAQ.Heading` inside an `FAQGroup` would ignore any `as` passed in.

Based on this I think our changelog covers everything. If consumers have added workarounds then those will continue to work. If they haven't then they just need to be aware that defaults are changing.

.changeset/purple-plants-search.md
```diff
@@ -0,0 +1,9 @@
+---
+'@primer/react-brand': patch
+---
+
+Improved default heading levels in the `FAQ` component.
+
+- Updated `FAQ.Subheading` heading level from `h3` to `h4`.
+- `FAQ.Question` now renders as a `h5` if there is a `FAQ.Subheading` present, or if rendered inside a `FAQGroup`.
```
rezrah, line 8: >`FAQ.Question` now renders as a `h5` if there is a `FAQ.Subheading` present

Doesn't seem be to be working in [this story](https://primer-875bfd87e2-26139705.drafts.github.io/brand/storybook/?path=/story/components-faq-features--with-subheadings), where the question should be a h5 not h4? Feel like this was working before, so maybe something regressed between commits?

Screenshot 2025-06-24 at 16 43 09

joshfarrant: Thanks Rez, this was an edge case I'd missed. When the `FAQ.Items` are wrapped in an array and a Fragment we weren't traversing the children correctly. You can see the logic I added to handle this [here](https://github.com/primer/brand/pull/1053/files/341f120af1ffb6f50a014691250327e03b62f243..03133530eb6c8e14190a9a7c5677007981d31bd9).

I've checked the heading levels of all of the stories, and added tests for these edge cases, and I think we're good now.

packages/react/src/FAQ/FAQ.test.tsx
```diff
@@ -181,4 +180,96 @@ describe('FAQ', () => {
...
+    const mainheadingEl = getByRole('heading', {level: 3, name: mockHeading})
+    const subheadingEl = getByRole('heading', {level: 4, name: mockSubheading})
+    const questionheadingEl = getByRole('heading', {level: 5, name: mockQuestion})
+
+    expect(mainheadingEl).toBeInTheDocument()
+    expect(subheadingEl).toBeInTheDocument()
+    expect(questionheadingEl).toBeInTheDocument()
+  })
+
+  it('renders the heading as a h3, the subheading as a h4, and the question as a h5 when there is a subheading present, and the FAQ.Item is wrapped in a Fragment', () => {
```
rezrah, line 227: 🥇 for longest test case name I've seen, very thorough... love it 😄

joshfarrant: If you like that, you'll love:

> [renders the heading as a h3, the subheading as a h4, and the question as a h5 when there is a subheading present, FAQ.Items are wrapped in a Fragment, and all FAQ.Items are wrapped in an array](https://github.com/primer/brand/pull/1053/files#diff-ac6cfaccc6ab9f035fb680e4596da44deb9cca96c1653d0a7b5111e0d21aa4d6R252)

I don't want to have to think about that edge case ever again — it's Jest's problem now 😂

Issue #1080, opened 2025-07-02
🐛 [BUG] - Clicking a Checkbox which uses a callback ref doesn't change its checked state


joshfarrant:
### Describe the bug

Rendering a Checkbox with a callback ref stops the checkbox graphic from being clickable

### Reproduction steps

```bash
1. Navigate to a simple checkbox example in the production docs.
1. Replace <Checkbox /> with <Checkbox ref={() => {}} />
1. Observe the clicking the checkbox no longer toggles the checkbox state.
```

### Expected behavior

```bash
The checked state changes on click
```

### Screenshots

```bash

```

### Browsers

Chrome

### OS

Mac

PR #1081, opened 2025-07-02
update useProvidedRefOrCreate hook to support function refs

rezrah:
## Summary

Resolves https://github.com/primer/brand/issues/1080

## List of notable changes:

-
-

-
-

## Steps to test:

1.
1.
1.

-
-

## Contributor checklist:

## Reviewer checklist:

## Screenshots:

<table>
<tr>
</tr>
<tr>
<td valign="top">

 </td>
<td valign="top">

</td>
</tr>
</table>

### DIFF
--- a/packages/react/src/hooks/useRef.ts
+++ b/packages/react/src/hooks/useRef.ts
@@ -1,4 +1,4 @@
-import React, {useRef} from 'react'
+import React, {useRef, useMemo} from 'react'
 
 /**
  * There are some situations where we only want to create a new ref if one is not provided to a component
@@ -8,9 +8,55 @@ import React, {useRef} from 'react'
  * @param providedRef The ref to use - if undefined, will use the ref from a call to React.useRef
  * @type TRef The type of the RefObject which should be created.
  */
-
-export function useProvidedRefOrCreate<TRef>(providedRef?: React.RefObject<TRef>): React.RefObject<TRef> {
+export function useProvidedRefOrCreate<TRef>(
+  providedRef?: React.RefObject<TRef> | React.RefCallback<TRef>,
+): React.RefObject<TRef> {
   const createdRef = useRef<TRef>(null)
+  const callbackRef = useRef<React.RefCallback<TRef> | null>(null)
+  const wrapperRef = useRef<React.MutableRefObject<TRef | null> | null>(null)
 
-  return providedRef ?? createdRef
+  // needed to keep the callback ref up-to-date between renders
+  callbackRef.current = typeof providedRef === 'function' ? providedRef : null
+
+  const finalRef = useMemo(() => {
+    // Return a new ref if no ref was provided
+    if (!providedRef) {
+      wrapperRef.current = null
+      return createdRef
+    }
+
+    // Return the provided ref if it's a RefObject
+    if (typeof providedRef === 'object' && providedRef.current !== undefined) {
+      wrapperRef.current = null
+      return providedRef
+    }
+
+    // Handle function refs
+    // We wrap it in a MutableRefObject to ensure we can set the current value
+    // and also call the callback with the latest value
+    if (typeof providedRef === 'function') {
+      if (!wrapperRef.current) {
+        let currentVal: TRef | null = null
+
+        wrapperRef.current = {
+          get current() {
+            return currentVal
+          },
+          set current(value: TRef | null) {
+            currentVal = value
+
+            if (callbackRef.current) {
+              callbackRef.current(value)
+            }
+          },
+        }
+      }
+      return wrapperRef.current
+    }
+
+    wrapperRef.current = null
+    return createdRef
+  }, [providedRef, createdRef])
+
+  return finalRef
 }
--- a/packages/react/src/forms/Checkbox/Checkbox.test.tsx
+++ b/packages/react/src/forms/Checkbox/Checkbox.test.tsx
@@ -1,0 +1,106 @@
+import React from 'react'
+import {render, cleanup} from '@testing-library/react'
+import userEvent from '@testing-library/user-event'
+import '@testing-library/jest-dom'
+import {toHaveNoViolations} from 'jest-axe'
+
+import {Checkbox} from '../Checkbox'
+
+expect.extend(toHaveNoViolations)
+
+describe('Checkbox', () => {
+  afterEach(cleanup)
+
+  it('renders a checkbox correctly into the document', () => {
+    const {getByRole} = render(<Checkbox value="one" defaultChecked />)
+
+    const checkbox = getByRole('checkbox')
+    expect(checkbox).toBeInTheDocument()
+    expect(checkbox).toBeChecked()
+    expect(checkbox).toHaveAttribute('value', 'one')
+  })
+
+  it('toggles the checkbox when clicked', async () => {
+    const user = userEvent.setup()
+
+    const {getByRole} = render(<Checkbox value="one" defaultChecked />)
+
+    const checkbox = getByRole('checkbox')
+
+    expect(checkbox).toBeChecked()
+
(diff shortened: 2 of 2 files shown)


packages/react/src/hooks/useRef.ts
```diff
@@ -1,16 +1,63 @@
...
+            }
+          },
+        }
+      }
+      return wrapperRef.current
+    }
+
+    wrapperRef.current = null
+    return createdRef
+  }, [providedRef, createdRef])
```
joshfarrant: Looks good! Only thing I found was that I was having trouble following some of the logic and understanding the `wrapperRef.current = null` that we're doing in a few places. What do you think about reordering the logic slightly?

(Maybe try copy/pasting the below suggestion into VSCode as the diff view here is showing it as a complete change, when all I've done is reorder the checks from `falsey > object > function > default` to `function > object > default`, which allows us to only set `wrapperRef.current = null` once, which makes it a little clearer imo. wdyt?)


```suggestion
  const finalRef = useMemo(() => {
    // Handle function refs
    // We wrap it in a MutableRefObject to ensure we can set the current value
    // and also call the callback with the latest value
    if (typeof providedRef === 'function') {
      if (!wrapperRef.current) {
        let currentVal: TRef | null = null

        wrapperRef.current = {
          get current() {
            return currentVal
          },
          set current(value: TRef | null) {
            currentVal = value

            if (callbackRef.current) {
              callbackRef.current(value)
            }
          },
        }
      }
      return wrapperRef.current
    }

    wrapperRef.current = null

    // Return the provided ref if it's a RefObject
    if (typeof providedRef === 'object' && providedRef?.current !== undefined) {
      return providedRef
    }

    // Return a new ref if it's neither a function nor a RefObject
    return createdRef
  }, [providedRef, createdRef])
```

joshfarrant: Ah sorry, til `typeof null` is `"object"`, so I've updated my suggestion slightly.

packages/react/src/hooks/useRef.ts
```diff
@@ -1,16 +1,63 @@
...
  * This hook aims to encapsulate that logic, so the consumer doesn't need to be concerned with violating `rules-of-hooks`.
- * @param providedRef The ref to use - if undefined, will use the ref from a call to React.useRef
+ * @param providedRef The ref to use - can be a RefObject, RefCallback (functional ref), or undefined.
  * @type TRef The type of the RefObject which should be created.
+ * @returns A RefObject that either uses the provided ref, wraps a functional ref, or creates a new one
  */
-
-export function useProvidedRefOrCreate<TRef>(providedRef?: React.RefObject<TRef>): React.RefObject<TRef> {
+export function useProvidedRefOrCreate<TRef>(
+  providedRef?: React.RefObject<TRef> | React.RefCallback<TRef>,
```
joshfarrant: I think this type needs updating


```suggestion
  providedRef?: React.RefObject<TRef> | React.RefCallback<TRef> | null,
```

If you add something like this to the hook

```tsx
if (providedRef === null) {
  console.log('providedRef is null')
}
```

Then TS will complain at you for an unnecessary conditional, but if you check the console you'll see that the console.log is printed. I think that's due to our type assertions. Eg in SubNav we have

```tsx
const ref = useProvidedRefOrCreate(forwardedRef as RefObject<HTMLDivElement>)
```

If you remove `as RefObject<HTMLDivElement>` then you'll see a type error as `forwardedRef` may be null.

wdyt to adding `| null` as suggested above and removing the type assertions we have around the usage of this hook? Having correct typing throughout our components will give us more confidence in the bigger change to the hook

.changeset/sixty-walls-pull.md
```diff
@@ -0,0 +1,5 @@
+---
+'@primer/react-brand': patch
+---
+
+Update `useProvidedRefOrCreate` to support functional refs
```
joshfarrant: As this hook isn't exposed to consumers, do you think it would be more relevant to say something like


```suggestion
Updated `Accordion`, `Checkbox`, `Radio`, `SubNav`, `Tooltip` and `VideoPlayer` components to support functional refs.
```

wdyt?

(I've omitted `Card`, `RiverAccordion` as, although they use the hook, they don't actually interact with the ref and instead just forward it on, so they've always supported functional refs. The hook could safely be removed from those two components, but I think that's outside the scope of this PR.)

joshfarrant (commented):
Looks good! I've tested across the components that consume this hook and everything's working as expected 👍

Just a couple of small questions/comments

packages/react/src/hooks/useRef.ts
```diff
@@ -1,16 +1,63 @@
...
+        wrapperRef.current = {
+          get current() {
+            return currentVal
+          },
+          set current(value: TRef | null) {
+            currentVal = value
+
+            if (callbackRef.current) {
+              callbackRef.current(value)
+            }
```
joshfarrant: I could definitely be missing something here, but could you help me understand this `callbackRef` please?

Since we're already inside the `typeof providedRef === 'function'` block, TS knows that `providedRef` is a function, so could we just call `providedRef(value)` directly here? Do we need the callbackRef at all?


```suggestion
            providedRef(value)
```

packages/react/src/hooks/useRef.ts
rezrah: 👍 yep don't mind inverting that as it removes a few extra lines and should be a noop change.

packages/react/src/hooks/useRef.ts
rezrah: >I think this type needs updating

🤔 Yeah, can see that issue and have added null now. I admittedly haven't looked too deeply into implications of adding `null` as a valid type, but assume it's safe enough. Tests are still passing everywhere. I don't plan to remediate those other components like SubNav castings right now though.

packages/react/src/hooks/useRef.ts
rezrah: It's to ensure that the providedRef is updated across rerenders, so I reset it on line 20 and store the latest version of it there. Calling providedRef directly could mean something changes upstream and we would be out of sync in the hook.

.changeset/sixty-walls-pull.md
rezrah: I like it 👍. I want to be clear where the main changed happened though, so went with a slight hybrid. that's inclusive of your suggestion.

packages/react/src/hooks/useRef.ts
joshfarrant: Great, thanks for adding that 👍

I don't think there are any implications of adding `null` as a type because `null` is already being passed to the function, e.g. on first render of a component using this hook. It's already supporting it, so by adding it we're just being explicit about it and letting TS know that it's a possibility.

packages/react/src/hooks/useRef.ts
joshfarrant: Ah ok, I see what you're trying to defend against.

Just to check though - since the `useMemo` depends on `providedRef`, won't the `useMemo` re-run if `providedRef` ever changes? Won't that mean the following can never happen?

> Calling providedRef directly could mean something changes upstream and we would be out of sync in the hook.

Just trying to make sure I'm not missing something here 🙂

rezrah: Ah I see, yes that's correct. Removing providedRef fixes it, but becomes an exhaustive dependency issue so calling providedRef directly works. I'll go ahead and apply that.

joshfarrant: Awesome, thanks!

We can get rid of `callbackRef` completely now then, right?

rezrah: 👍 yes that's been cleaned up

PR #451, opened 2023-10-05
SubdomainNavBar overflow fix

joseph-lozano:
## Summary

This PR fixes 2 issues:
1. If the sandwich menu was open on narrow screen sizes, and then the viewport was widened, nav items would be duplicated.
2. If there were too many nav items for the height of the viewport, they would overflow into the CTA actions. (#405)

Fixes #405 
## List of notable changes:

The if the Nav menu is open, the background will no longer be scrollable, but the Nav menu will be.

## What should reviewers focus on?

Verify that the SubdomainNavBar is working as intended, and there were no unintended side-effect changes.

## Steps to test:

1. Go to `MobileMenuOpenManyItems` in Storybook and verify that it is working as intended.
1. Verify that it is working as intended.
1. Ensure all other stories are unchanged

## Contributor checklist:

- [ ] All new and existing CI checks pass
- [ ] Tests prove that the feature works and covers both happy and unhappy paths
- [ ] Any drop in coverage, breaking changes or regressions have been documented above
- [ ] New visual snapshots have been generated / updated for any UI changes
- [ ] All developer debugging and non-functional logging has been removed
- [ ] Related issues have been referenced in the PR description

## Reviewer checklist:

- [ ] Check that pull request and proposed changes adhere to our [contribution guidelines](../../CONTRIBUTING.md) and [code of conduct](../../CODE_OF_CONDUCT.md)
- [ ] Check that tests prove the feature works and covers both happy and unhappy paths
- [ ] Check that there aren't other open Pull Requests for the same update/change

## Screenshots:

> Please try to provide before and after screenshots or videos

<table>
<tr>
<th> Before</th> <th> After </th>
</tr>
<tr>
<td valign="top">

CleanShot 2023-10-05 at 12 50 22@2x

 </td>
<td valign="top">

CleanShot 2023-10-05 at 12 50 42@2x

</td>
</tr>
</table>

PR #1082, opened 2025-07-03
Fix leaking props in SubdomainNavBar to remove warning from unit tests

joshfarrant:
## Summary

Fixes props which were leaking into the DOM, both in Storybook and in the `SubdomainNavBar` component itself.

## List of notable changes:

- Explicitly destructured props in `SubdomainNavBar.stories.tsx` to prevent props leaking into the DOM of rendered story.
- Removed unused `showOnlyOnNarrow` prop from `SubdomainNavBar.tsx` to prevent it leaking into the DOM and to fix a related warning in the unit tests.

- I can't see the `showOnlyOnNarrow` prop being used anywhere, and I'm not sure it was ever implemented since being added in #451. Another set of eyes on this would be appreciated though.
- Check that the SubdomainNavBar still works in the expected way.

## Contributor checklist:

## Reviewer checklist:

## Screenshots:

<table>
<tr>
</tr>
<tr>
<td valign="top">

image

 </td>
<td valign="top">

image

</td>
</tr>
</table>

### DIFF
--- a/packages/react/src/SubdomainNavBar/SubdomainNavBar.tsx
+++ b/packages/react/src/SubdomainNavBar/SubdomainNavBar.tsx
@@ -322,7 +322,6 @@ function Root({
                     )}
                     {hasLinks && !menuHidden && (
                       <NavigationVisbilityObserver
-                        showOnlyOnNarrow
                         className={clsx(styles['SubdomainNavBar-primary-nav-list--visible'])}
                       >
                         {menuItems}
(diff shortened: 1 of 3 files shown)


packages/react/src/SubdomainNavBar/SubdomainNavBar.tsx
```diff
@@ -322,7 +322,6 @@ function Root({
                     )}
                     {hasLinks && !menuHidden && (
                       <NavigationVisbilityObserver
-                        showOnlyOnNarrow
```
rezrah, line 325: Oh wow, so this wasn't doing anything at all? 😬

joshfarrant: No I don't believe so 😅

PR #1085, opened 2025-07-03
Fix VideoPlayer in Next docs

joshfarrant:
## Summary

Fixes `VideoPlayer` in Next docs by ensuring `VideoPlayer` `ref` isn't overriden by `null` ref passed by Next docs. I'm not entirely sure why, but the Next docs pass a ref of `null` to rendered components.

The `VideoPlayer` intentionally doesn't accept a `ref` prop as it instead exposes a `useVideo()` context to facilitate interaction with the video player. A `ref` is available through that context if required.

Because of this, we can explicitly override any refs provided directly to `VideoPlayer` and resolve the issue with the Next docs.

- Check that the VideoPlayer functions in the Next docs

## Contributor checklist

## Reviewer checklist


packages/react/src/VideoPlayer/VideoPlayer.tsx
```diff
@@ -70,7 +70,7 @@ const Root = ({
   return (
     <div className={styles.VideoPlayer__container} ref={fullscreenRef}>
       <div className={styles.VideoPlayer__overlayContainer}>
-        <video ref={ref} title={title} controls={false} className={clsx(styles.VideoPlayer, className)} {...rest}>
+        <video title={title} controls={false} className={clsx(styles.VideoPlayer, className)} {...rest} ref={ref}>
```
rezrah, line 73: @joshfarrant can you add a test for this to ensure the ref doesn't get overridden?

.changeset/chilled-bottles-hug.md
```diff
@@ -0,0 +1,5 @@
+---
+'@primer/react-brand': patch
+---
+
+Minor internal refactoring to `VideoPlayer` component
```
rezrah, line 5: Could there be a scenario where someone was passing a ref as a prop, which now won't work because we're no longer spreading it in?

joshfarrant: I don't believe that's possible no. The `VideoPlayer` doesn't implement `forwardRef`, so `ref` is dropped as a prop before it ever reaches the component. That means it's not been possible for someone to pass in a ref before this point.

Given that, I'm not 100% sure why `props.ref` is ending up as `null` (rather than `undefined`) when rendered in the docs.

The only thing that I think could be causing it is that `react-live` doesn't implement that same React behaviour of dropping the `ref` prop, which results in it getting through to the component.

In that case, the only environment this change should affect is in our docs.

packages/react/src/VideoPlayer/VideoPlayer.tsx
joshfarrant: I don't think I can unfortunately.

We'd be testing the internals of React at this point and I'm not sure there's a way to actually assert that.

Might be worth discussing this a bit further as — if I understand the underlying issue correctly — we're working around a react-live bug/functionality here.

joshfarrant: I've just tested in [react-live's demo](https://nearform.com/open-source/react-live/#demo) and the `ref` prop isn't being set to `null` there, so maybe there's a bug in our implementation. I'll put this PR on hold for a while and look into it a bit further

rezrah:
@joshfarrant FYI this change, combined with the recent Doctocat updates seems to have fixed the VideoPlayer previews 🙌 https://primer-e8d998b27d-26139705.drafts.github.io/brand/components/VideoPlayer/

Once you're confident this change won't lead to a downstream regression in the library, let's ship it.

PR #1086, opened 2025-07-07
Update Footnotes component docs to increase InlineLink touch area size

joshfarrant:
## Summary

Increase `InlineLink` touch area size in `Footnotes` documentation.

There's no minimum required size for these elements as they're inline links, but making them larger is a best practice for accessibility.

🔗 [Link to documentation](https://primer-56784b469e-26139705.drafts.github.io/brand/components/Footnotes/react/#citations-with-return-links)

## List of notable changes

- Increase touch target size of `InlineLink` in `Footnotes` documentation
- Add a recommendation to `Footnote` documentation to increase touch target size

- Are you happy with the docs change and the new touch target size?

- Closes https://github.com/github/primer/issues/4671

## Contributor checklist

## Reviewer checklist

## Screenshots

<table>
<tr>
</tr>
<tr>
<td valign="top">

image

 </td>
<td valign="top">

image

</td>
</tr>
</table>

### DIFF
--- a/apps/next-docs/content/components/Footnotes/react.mdx
+++ b/apps/next-docs/content/components/Footnotes/react.mdx
@@ -118,7 +118,7 @@ When linking to `Footnotes` via a citation, ensure that the touch target is larg
         Unlocking innovation at scale with AI-driven software development.
       </Heading>
       <Text as="p" size="200" weight="light" variant="muted">
-        The world&apos;s most widely adopted AI developer tool.{' '}
+        The most widely adopted AI developer tool.{' '}
         <Text size="100" weight="light">
           <sup>
             <InlineLink


apps/next-docs/content/components/Footnotes/react.mdx
```diff
@@ -89,7 +91,11 @@ import {Footnotes} from '@primer/react-brand'
             Third-party tools support your favorite languages and frameworks{' '}
             <Text size="100" weight="light">
               <sup>
-                <InlineLink href="#economic-impact" id="economic-impact-ref">
+                <InlineLink
```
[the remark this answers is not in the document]
joshfarrant: This is Prettier's choice. The first usage goes over the character limit, the second doesn't

packages/react/src/Footnotes/Footnotes.examples.stories.tsx
```diff
@@ -34,7 +34,11 @@ export const RiversWithCitations = () => (
                       Third-party tools support your favorite languages and frameworks{' '}
                       <Text size="100" weight="light">
                         <sup>
-                          <InlineLink href="#economic-impact" id="economic-impact-ref">
+                          <InlineLink
+                            href="#economic-impact"
+                            id="economic-impact-ref"
+                            style={{display: 'inline-block', minWidth: 'var(--base-size-24)'}}
```
rezrah: The space is a bit too wide IMO, and doesn't work when you have text following the reference link. 

Screenshot 2025-07-10 at 11 52 14

Could we knock this back to `--base-size-12`?

Also, we aren't adding any space vertically or to the left of the link, is that okay? I assumed increasing click target meant adding space in all directions?

joshfarrant: There's not any specific guidance on how/where we should apply the space in this particular case. If we were to add additional vertical space then we'd either be increasing the line height, or at risk of having our touch area overlap with nearby links.

We don't actually _have_ to increase the touch area here as it's an inline link and exempt from that accessibility criterion. This is purely a recommendation to improve UX

joshfarrant: We could go to a min-height of 24px by the looks of it, as that's the line height we use in this scenario

apps/next-docs/content/components/Footnotes/react.mdx
```diff
@@ -115,7 +121,11 @@ import {Footnotes} from '@primer/react-brand'
         The world&apos;s most widely adopted AI developer tool.{' '}
         <Text size="100" weight="light">
           <sup>
-            <InlineLink href="#ai" id="ai-ref">
+            <InlineLink
```
rezrah, line 124: The reference link falls to new line, as a workaround can we just drop world's so they appear on same line?

rezrah: Screenshot 2025-07-10 at 14 32 41

PR #1094, opened 2025-07-09
Add i18n locale / language switcher to Storybook

rezrah:
## Summary

Towards https://github.com/github/primer/issues/2667

Adds a localization / language switcher to Storybook, and paves path for multi-lingual stories.

🔗 [Storybook preview](https://primer-adb63896be-26139705.drafts.github.io/brand/storybook/?path=/story/components-accordion-features--always-exactly-one-panel-open)

Screenshot 2025-07-10 at 10 10 49

## List of notable changes:

- **adds** support for react-i18n to Storybook, using manual translation files 
- **adds** content localizations for fr, en, ja, de, pt-BR and es to Accordion component only
- **adds** visual snapshot blobs for all localized stories (~3000), but only really working for Accordion
  - VRT timings have increased from 7m => 12m as a result of the additional snapshots 
- **adds** a translation file validator, to ensure that all languages share the same keys and aren't missing files
  
  Screenshot 2025-07-11 at 16 15 33

## Steps to test:

1. Review code changes using [this filtered file diff ](https://github.com/primer/brand/pull/1094/files?file-filters%5B%5D=.js&file-filters%5B%5D=.json&file-filters%5B%5D=.ts&file-filters%5B%5D=.tsx&show-viewed-files=true)(excludes images)

## Contributor checklist:

## Reviewer checklist:

## Screenshots:

JA:
image

DE: 
image


apps/storybook/.storybook/preview.js
```diff
@@ -34,6 +37,38 @@ export const globalTypes = {
       title: 'Color mode',
     },
   },
+  locale: {
+    name: 'Locale',
+    description: 'Internationalization locale',
+    defaultValue: 'en',
+    toolbar: {
+      icon: 'globe',
```
joshfarrant, line 45: This is super unnecessary, but wdyt about changing the icon we use for "Color mode" to `'paintbrush'`, as I agree that `"globe"` makes sense for "Locale".

Completely unnecessary, but having different icons for the two different controls makes sense to me 😄

[List of supported icons](https://main--64b56e737c0aeefed9d5e675.chromatic.com/?path=/docs/introduction--docs)

apps/storybook/static/locales/en/translation.json
```diff
@@ -0,0 +1,21 @@
...
+  "here": "here",
+  "Who is eligible to apply?": "Who is eligible to apply?",
+  "Startups who meet the following criteria are eligible to apply for the program:": "Startups who meet the following criteria are eligible to apply for the program:",
+  "Must be associated with a current GitHub for Startups partner.": "Must be associated with a current GitHub for Startups partner.",
+  "Self-funded or funded (Seed-Series A)": "Self-funded or funded (Seed-Series A)",
+  "Not a current GitHub Enterprise customer": "Not a current GitHub Enterprise customer",
+  "Must not have previously received credits for GitHub Enterprise": "Must not have previously received credits for GitHub Enterprise",
+  "What if my startup is not eligible? Are there other resources for me?": "What if my startup is not eligible? Are there other resources for me?",
+  "If you're not currently eligible for the GitHub for Startups but would like to try GitHub Enterprise, please feel to sign up for a trial": "If you're not currently eligible for the GitHub for Startups but would like to try GitHub Enterprise, please feel to sign up for a trial",
```
joshfarrant: This existed before your change, but I've just spotted it 🙂

```suggestion
  "If you're not currently eligible for the GitHub for Startups but would like to try GitHub Enterprise, please feel to sign up for a trial": "If you're not currently eligible for the GitHub for Startups but would like to try GitHub Enterprise, please feel free to sign up for a trial",
```

packages/react/src/Accordion/Accordion.features.stories.tsx
```diff
@@ -10,79 +11,95 @@ const meta: Meta = {
...
+        <Accordion>
+          <Accordion.Heading>
+            {t('What if my startup is not eligible? Are there other resources for me?')}
+          </Accordion.Heading>
+          <Accordion.Content>
+            <p>
+              If you’re not currently eligible for the GitHub for Startups but would like to try GitHub Enterprise,
+              please feel to sign up for a trial{' '}
+              <a href="https://copilot.github.com/" target="_blank" rel="noreferrer">
+                here
```
joshfarrant: Should this use the translation?

apps/storybook/static/locales/en/translation.json
```diff
@@ -0,0 +1,21 @@
+{
```
joshfarrant: Is having full sentences as keys here the right way to go? I'm wondering if it's a bit redundant and if it will be harder to maintain than simpler, structured keys, eg

```json
{
  "Accordion": {
    "whatsIncluded": {
      "question": "What's included in the GitHub for Startups offer?",
      "answer": "All GitHub for Startups companies receive up to {{seats}} seats of GitHub Enterprise for free for year one and {{discount}}% off year two. Learn more about the features and capabilities of GitHub Enterprise",
      "link": "here"
    },
    "eligibility": {
      "question": "Who is eligible to apply?",
      "intro": "Startups who meet the following criteria are eligible to apply for the program:",
      "criteria": {
        "partner": "Must be associated with a current GitHub for Startups partner.",
        "funding": "Self-funded or funded (Seed-Series A)",
        "notCustomer": "Not a current GitHub Enterprise customer",
        "noCredits": "Must not have previously received credits for GitHub Enterprise"
      }
    },
    "notEligible": {
      "question": "What if my startup is not eligible? Are there other resources for me?",
      "answer": "If you're not currently eligible for the GitHub for Startups but would like to try GitHub Enterprise, please feel free to sign up for a trial",
      "link": "here"
    },
    "becomePartner": {
      "question": "How can my organization become a GitHub for Startups partner?",
      "answer": "Any investor, accelerator, or startup support organization is eligible to apply for the GitHub for Startups program.",
      "link": "Apply here"
    }
  }
}
```

My concern is that the current structure will quickly become difficult to maintain as we add more translations

apps/storybook/static/locales/en/translation.json
```diff
@@ -0,0 +1,21 @@
...
+  "If you're not currently eligible for the GitHub for Startups but would like to try GitHub Enterprise, please feel to sign up for a trial": "If you're not currently eligible for the GitHub for Startups but would like to try GitHub Enterprise, please feel to sign up for a trial",
+  "How can my organization become a GitHub for Startups partner?": "How can my organization become a GitHub for Startups partner?",
+  "Any investor, accelerator, or startup support organization is eligible to apply for the GitHub for Startups program.": "Any investor, accelerator, or startup support organization is eligible to apply for the GitHub for Startups program.",
+  "Apply here": "Apply here",
+  "Toggle color": "Toggle color",
+  "Description": "Description",
+  "Lorem ipsum dolor, sit amet consectetur adipisicing elit. Blanditiis quidem veniam vero omnis consequuntur cum quae libero dolor dicta odio in, corporis perspiciatis nesciunt facere. Eius vero culpa quae itaque?": "Lorem ipsum dolor, sit amet consectetur adipisicing elit. Blanditiis quidem veniam vero omnis consequuntur cum quae libero dolor dicta odio in, corporis perspiciatis nesciunt facere. Eius vero culpa quae itaque?"
```
joshfarrant: A few of these appear to be unused. Is that intentional or can they be removed?

As an aside, is there any way to detect unused translations?

packages/react/src/Accordion/Accordion.features.stories.tsx
```diff
@@ -10,79 +11,95 @@ const meta: Meta = {
...
-              Apply here
-            </a>
-            .
-          </p>
-        </Accordion.Content>
-      </Accordion>
-    </>
-  ),
+  render: function CompositionComponent() {
+    const {t} = useTranslation()
```
joshfarrant: To remove a bit of duplication and prevent us from having to call this hook every time, wdyt to wrapping it in a decorator?

```tsx
const WithTranslationDecorator: Decorator = (Story, context) => {
  const {t} = useTranslation()
  context.t = t
  return <Story {...context} />
}
```

```tsx
const meta: Meta = {
  title: 'Components/Accordion/Features',
  component: Accordion,
  decorators: [WithTranslationDecorator],
}

type Story = StoryObj<typeof meta>
```

```tsx
export const Composition: Story = {
  render: function CompositionComponent(args, {t}) {
    return (
      <div>{t('here')}</div>
      {/* ... */}
```

Components can then opt-in to translations in a centralised way.

This could potentially be integrated into your `withI18next` decorator too

apps/storybook/.storybook/preview.js
rezrah: Yeah maybe in future 👍 But if we change it, I'd rather use the same one as product primer for consistency:

Screenshot 2025-07-10 at 15 56 01

apps/storybook/static/locales/en/translation.json
rezrah: I see the concern. I prefer the arbitrary-ness of having non semantic keys. It means we don't need to worry about nesting, and/or naming challenges. FYI I've also added a validator now, which will drop nested keys.

apps/storybook/static/locales/en/translation.json
rezrah: I'll go through the list to double-check, but no there isn't a way to cross-check their usage. I have however, added a validator to ensure that keys are sync'd between the different translation files 😌

apps/storybook/scripts/validate-translation-keys.ts
```diff
@@ -0,0 +1,271 @@
...
+    process.exit(1)
+  }
+})()
+
+function loadJsonFile(filePath: string): TranslationContent | null {
+  try {
+    const content = fs.readFileSync(filePath, 'utf8')
+    return JSON.parse(content)
+  } catch (error) {
+    console.error(`Error loading ${filePath}: ${(error as Error).message}`)
```
joshfarrant: Nit, but you can do something like this to make sure you still see the log on the off chance that the `error` that's thrown isn't an instance of `Error`. It removes an unnecessary type assertion too

```suggestion
    console.error(`Error loading ${filePath}: ${${error instanceof Error ? error.message : error}}`)
```

packages/react/src/Accordion/Accordion.features.stories.tsx
rezrah: 🤔 yeah, I can certainly see the value in doing that. I also wonder how confusing this could be, a `t` is so ambiguous unless you know two things: a) this is for translations and b) there's a magical decorator handling this. For this reason, I prefer being really explicit about it, even if that leads to a bit more duplication.

joshfarrant (approved):

danielguillan (approved):
LGTM! 🚀

PR #1101, opened 2025-07-14
Add play/pause button to IDE component

joshfarrant:
## Summary

Added a play/pause button to the IDE component and removed the `showReplayButton` prop as it's no longer needed.

## List of notable changes

- Added a play/pause button to the IDE component to allow animations to be paused and resumed by the user.
  - Pressing the pause button pauses both the chat and editor animations, if present
  - When both animations complete, pressing the button will replay the animation
  - Animations run separately, as they did before, so you can switch between tabs while a chat animation is running, for example
- Removed the Replay button
- Added tests to assert that animations play and can be paused.

### Implementation

Both the `Chat` and `Editor` components now run their animations via `requestAnimationFrame`, rather than `setTimeout` as they did previously. This change was required as there was no practical way of pausing and resuming a group of timeouts; once they were started they would run until complete.

The root `IDE` component now interacts with `Chat` and `Editor` in 3 ways:

- `IDE` passes them both an `animationIsPaused` prop which they use to decide whether or not to play the animation.
- `IDE` also passes them both with a `setAnimationIsDone` callback so they can each tell report back when their animation is finished. Once both animations have finished the `IDE` sets the play/pause button to the paused state, allowing it to be clicked to replay the animation.
- `Chat` and `Editor` expose a `resetAnimation` method via `useImperativeHandle` which `IDE` can call when the user presses the play button to replay the animation.

### Limitations

- The [Perspective examples](https://primer-fdb5def8e5-26139705.drafts.github.io/brand/storybook/?path=/story/components-ide-features--perspective-example) in Storybook hide the play/pause button, but I'm not sure that this is something we need to work around as it's just an example.
- When both the Editor and the Chat are used, we hide the chat completely on narrow viewports which results in the animation completing on the editor but the play/pause button continues to indicate that the animation is still in progress as it waits for the (not visible to the user) chat animation to complete. I don't think this is something we need to work around at the moment as the loss of information which results from hiding the chat completely on narrow viewports is an accessibility issue in and of itself.

This PR fundamentally changes how animations are run, however nothing else should have changed in the component.

- Check that animations work as expected and can be paused, resumed, and replayed across the chat and/or the editor. You should be able to change tabs as animations are playing too, as before.

- Closes https://github.com/github/primer/issues/3767

## Contributor checklist

## Reviewer checklist


packages/react/src/IDE/IDE.tsx
```diff
@@ -120,125 +224,253 @@ export type IDEChatMessage = {handle: string; message: string; codeSnippet?: str
...
+                // Remove faded class from all messages at the end
+                if (animation.isLast) {
+                  setTimeout(() => {
+                    for (const defaultMessage of defaultMessages) {
+                      defaultMessage.classList.remove(styles['IDE__Chat-message--faded'])
+                    }
+                    for (const newerMessage of newerMessages) {
+                      newerMessage.classList.remove(styles['IDE__Chat-message--faded'])
+                    }
+                  }, 100)
```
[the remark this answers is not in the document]
joshfarrant: Practically speaking, I don't think this is a concern

packages/react/src/IDE/IDE.module.css
```diff
@@ -377,17 +377,52 @@ pre.IDE__Chat-copilot-indicator {
...
+  right: var(--base-size-16);
+  display: flex;
+  align-items: center;
+  justify-content: center;
+  width: var(--base-size-48);
+  height: var(--base-size-48);
+  margin: 0;
+  padding: 0;
+  border: none;
+  border-radius: var(--base-size-48);
```
rezrah: I think you should be able to use this:

```suggestion
  border-radius: var(--brand-borderRadius-full);
```

packages/react/src/IDE/IDE.test.tsx
```diff
@@ -235,4 +244,211 @@ describe('IDE', () => {
...
+
+    await act(async () => {
+      await wait(1000)
+    })
+
+    expect(message2Parent).not.toHaveClass('IDE__Chat-message--visible')
+    expect(message3Parent).not.toHaveClass('IDE__Chat-message--visible')
+
+    const playButton = getByRole('button', {name: 'Play animation'})
+    expect(playButton).toBeInTheDocument()
```
rezrah, line 300: Can we extend this to test that the final two messages are now present after restarting the animation? Can't see that being tested elsewhere.. Something like:

```suggestion
    expect(playButton).toBeInTheDocument()
    
    await user.click(playButton)

    await act(async () => {
      await wait(1000)
    })

    expect(message2Parent).toHaveClass('IDE__Chat-message--visible')
    expect(message3Parent).toHaveClass('IDE__Chat-message--visible')

File: packages/react/src/IDE/fixtures/content.ts [hunk 4]
```diff
@@ -165,9 +165,8 @@ const drawScatterplot = (data, height, width) => {
     name: 'parse_expenses.py',
     alternativeText:
       'Python function parsing expense records from a string format. Copilot suggests implementing date parsing logic and expense record processing.',
-    suggestedLineStart: 2,
-    code: hljs.highlight(
-      `import datetime
+    suggestedLineStart: 3,
+    code: `import datetime
```
rezrah: Why did we need to change these values?

packages/react/src/recipes/FeaturePreviewLPs/FeaturePreviewLevelTwo/FeaturePreviewLevelTwo.visual.spec.ts-snapshots/Visual-Comparison-FeaturePreviewLevelTwo-FeaturePreviewLevelTwo-2-1-variant-es-1-linux.png
rezrah, line 1: Any ideas on why the background color is missing on this play/pause button? 

Screenshot 2025-07-17 at 17 47 13

packages/react/src/IDE/IDE.tsx
```diff
@@ -27,6 +27,50 @@ import '@primer/brand-primitives/lib/design-tokens/css/tokens/functional/compone
 /** * Main Stylesheet (as a CSS Module) */
 import animationStyles from '../animation/Animation.module.css'
 import styles from './IDE.module.css'
+import {useProvidedRefOrCreate} from '../hooks/useRef'
+
+const PauseIcon = () => (
```
rezrah: Could you please move the play/pause code to bottom of the file? This is an internal-only component, so front-loading is is pushing the API, types and component code for the actual IDE component down to line 91 compared to 47.

packages/react/src/IDE/IDE.tsx
```diff
@@ -27,6 +27,50 @@ import '@primer/brand-primitives/lib/design-tokens/css/tokens/functional/compone
...
+  <svg fill="none" role="presentation" aria-hidden="true" viewBox="3.17 2.28 13.7 15.42">
+    <path
+      d="M4.66148 2.3125C3.83593 2.3125 3.16669 2.98174 3.16669 3.80729V16.1927C3.16669 17.0183 3.83593 17.6875 4.66148 17.6875H7.65106C8.47662 17.6875 9.14586 17.0183 9.14586 16.1927V3.80729C9.14586 2.98174 8.47662 2.3125 7.65106 2.3125H4.66148ZM4.44794 3.80729C4.44794 3.68936 4.54355 3.59375 4.66148 3.59375H7.65106C7.769 3.59375 7.86461 3.68936 7.86461 3.80729V16.1927C7.86461 16.3106 7.769 16.4062 7.65106 16.4062H4.66148C4.54355 16.4062 4.44794 16.3106 4.44794 16.1927V3.80729ZM12.349 2.3125C11.5235 2.3125 10.8542 2.98174 10.8542 3.80729V16.1927C10.8542 17.0183 11.5235 17.6875 12.349 17.6875H15.3386C16.1642 17.6875 16.8334 17.0183 16.8334 16.1927V3.80729C16.8334 2.98174 16.1642 2.3125 15.3386 2.3125H12.349ZM12.1355 3.80729C12.1355 3.68936 12.2311 3.59375 12.349 3.59375H15.3386C15.4565 3.59375 15.5521 3.68936 15.5521 3.80729V16.1927C15.5521 16.3106 15.4565 16.4062 15.3386 16.4062H12.349C12.2311 16.4062 12.1355 16.3106 12.1355 16.1927V3.80729Z"
+      fill="currentColor"
+    ></path>
+  </svg>
```
rezrah: Now that we have this in a couple of components, is there an opportunity to centralise this somewhere?

packages/react/src/IDE/IDE.tsx
```diff
@@ -61,25 +105,77 @@ export type IDEProps = {
...
+              onPlayPause={isPlaying => {
+                if (allAnimationsAreDone) {
+                  setChatAnimationIsDone(false)
+                  setEditorAnimationIsDone(false)
+                  chatRef.current?.resetAnimation()
+                  editorRef.current?.resetAnimation()
+                } else {
+                  setAnimationIsPaused(!isPlaying)
+                }
+              }}
```
rezrah: could you extract this into a separate handler function please, to avoid recreating it as an inline function. E.g. handleOnPlayPause.

packages/react/src/IDE/IDE.test.tsx
joshfarrant: I struggled with the tests a bit when testing the animations in the recommended way (using fake timers) as I couldn't find a way to make it work nicely with `requestAnimationFrame`

If you're happy for some tests to have longer than normal timeouts then I can just have them wait the few seconds it takes for all of the animations to run. I'll take a look shortly

packages/react/src/IDE/fixtures/content.ts
joshfarrant: I thought it may not have been set up correctly as this tab didn't animate in line-by-line like the other two tabs ([see deployed version](https://stunning-chainsaw-j82glqz.pages.github.io/?path=/story/components-ide-features--editor-only)), however if that's intentional then I'll revert this change.

packages/react/src/recipes/FeaturePreviewLPs/FeaturePreviewLevelTwo/FeaturePreviewLevelTwo.visual.spec.ts-snapshots/Visual-Comparison-FeaturePreviewLevelTwo-FeaturePreviewLevelTwo-2-1-variant-es-1-linux.png
joshfarrant: I'm not sure, no. It shows correctly in Storybook and if I generate the snapshots locally then it shows the correct colour too 🤔

packages/react/src/IDE/IDE.tsx
joshfarrant: Yes absolutely, that would be my preference too. Just to check, are you referring to the whole `PlayPauseButton`, or just the SVGs?

I'd be happy to create a new (internal-only) `PlayPauseButton` component which we can then use here, and in `LogoSuite`.

packages/react/src/IDE/IDE.tsx
joshfarrant: No problem 👍

[See here for a question](https://github.com/primer/brand/pull/1101#discussion_r2215142309) about pulling this component entirely out of this file too.

packages/react/src/recipes/FeaturePreviewLPs/FeaturePreviewLevelTwo/FeaturePreviewLevelTwo.visual.spec.ts-snapshots/Visual-Comparison-FeaturePreviewLevelTwo-FeaturePreviewLevelTwo-2-1-variant-es-1-linux.png
joshfarrant: It's resolved now — still not sure what caused it though

packages/react/src/IDE/fixtures/content.ts
joshfarrant: Reverted here 8c80ed00a988d394be468f4c1ab934c4b8372791

packages/react/src/IDE/IDE.test.tsx
joshfarrant: Great suggestion, thank you 🚀

packages/react/src/IDE/IDE.tsx
joshfarrant: Regardless of the chosen approach, would you be happy with me addressing it in a follow-up PR to prevent me from having to modify LogoSuite in this IDE-focussed PR?

rezrah (approved):

packages/react/src/IDE/IDE.tsx
rezrah: Yeah sounds good, but also not urgent

PR #1102, opened 2025-07-14
remove aria-disabled prop from Button and set appropriate disabled at…

joshfarrant:
…tribute depending on rendered element

## List of notable changes:

-
-

-
-

## Steps to test:

1.
1.
1.

-
-

## Contributor checklist:

## Reviewer checklist:

## Screenshots:

<table>
<tr>
</tr>
<tr>
<td valign="top">

 </td>
<td valign="top">

</td>
</tr>
</table>


packages/react/src/Button/Button.test.tsx
```diff
@@ -48,16 +48,76 @@ describe('Button', () => {
...
+    expect(btnEl).not.toHaveAttribute('aria-disabled')
+  })
+
+  it('sets the aria-disabled attribute to true when disabled prop is true and component is rendered as a link', () => {
+    const {getByText} = render(
+      <Button variant="primary" size="medium" disabled as="a">
+        Primary Button
+      </Button>,
+    )
+    const linkEl = getByText('Primary Button').closest('a')
```
rezrah: Do we need to use `closest` here? Can we use `getByRole` instead like we do for the other stories?

Could the reason for the matcher not working be that you're missing the href here?

```suggestion
    const { getByRole } = render(
      <Button as="a" href="#" disabled>
        Primary Button
      </Button>,
    );
    const linkEl = getByRole('link');
```

Removed the other unrelated props to keep the tests clean. Can we do same to others too please.

packages/react/src/Button/Button.tsx
```diff
@@ -153,6 +153,12 @@ export const _Button = forwardRef(
       [isDisabled, onBlur],
     )
 
+    let disabledProp
+
+    if (isDisabled) {
+      disabledProp = Component === 'button' ? {disabled: true} : {'aria-disabled': true}
```
rezrah: Can we do this inside the return instead please? It helps to co-locate prop logic in the same place and makes it easier to quickly understand what's happening in each render

joshfarrant: Do you have a preference between these options, or something different entirely?

```tsx
<Component
  {...(isDisabled ? (Component === 'button' ? {disabled: true} : {'aria-disabled': true}) : {})}
/>
```

```tsx
<Component
  disabled={(isDisabled && Component === 'button') || undefined}
  aria-disabled={(isDisabled && Component !== 'button') || undefined}
/>
```

packages/react/src/Button/Button.test.tsx
joshfarrant: `getByRole` wasn't working for me previously; it wouldn't find the element, which is why I ended up using `closest`.

You were exactly right that the missing `href` was the cause of the missing role. Adding that in allowed me to use `getByRole` — Thanks!

packages/react/src/Button/Button.tsx
joshfarrant: I went with option 2 — let me know if you'd prefer it otherwise.

rezrah: Yeah option two works 👍 

could we use a strict equality check though? 

E.g.  

```
<Component
  disabled={(isDisabled && Component === 'button') || undefined}
  aria-disabled={(isDisabled && Component === 'a') || undefined}
/>
```

Bit more readable and reflective of the requirements IMO

joshfarrant: I did consider that, although the Button's `as` prop isn't limited to just `'a'` and technically allows users to render it as any HTML element.

I went with this approach just on the off chance that someone did find a reason to render it as something else. This approach is a bit more forgiving in those cases, and still results in valid HTML.

I don't know what a good example would be, but the Button technically supports this kind of thing.

```tsx
// E.g. this
<details>
  <Button as="summary" disabled>Example</Button>
</details>

// Results in
<details>
  <summary aria-disabled="true">Example</summary>
</details>
```

Probably not the best example, but hopefully illustrates what I mean. We don't advertise this in our docs, but neither TypeScript nor the component disallow it.

If your confident it won't be an issue then I can make the change you suggested though — let me know your preference.

rezrah: > the Button's as prop isn't limited to just 'a' and technically allows users to render it as any HTML element.

Ahhh yeah forgot about that. Thanks for bringing that up. Yeah let's keep as-is then.

rezrah: Not great though 😅, I remember there was a use-case once before to set as a 'span' or 'div', but in hindsight the accessibility issues this could cause are kinda scary. Let's not scope-creep here though, and keep your inequality check in place.

.changeset/warm-meals-learn.md
```diff
@@ -0,0 +1,5 @@
+---
+'@primer/react-brand': patch
+---
+
+The `Button` component now automatically applies the appropriate disabled attribute (`disabled` for button elements, `aria-disabled` for other elements) based on the rendered element type.
```
rezrah: Something I didn't spot earlier, is that we may want to clarify the downstream impact here. 

I think we should recommend that users review usage of Button, and any dependencies on attributes being set in the markup. E.g. if they were targetting a[disabled] before, they should now target a[aria-disabled]. Does this make sense @joshfarrant, or am I overthinking this?

rezrah:
```suggestion
The `Button` component now automatically applies semantically correct disabled attributes based on the rendered element type. E.g. `disabled` for button elements and `aria-disabled` for other elements.

⚠️ Please review usage of `Button` in your application code or tests, to ensure that you are not relying on the previously incorrect markup. For example, if you previously targeted `a[disabled]`, you should now target `a[aria-disabled]` instead. 
```

packages/react/src/Button/Button.tsx
joshfarrant: Sounds good! 👍

.changeset/warm-meals-learn.md
joshfarrant: That's a good warning to add, thanks @rezrah

PR #1107, opened 2025-07-15
Add tests for Testimonial component

joshfarrant:
## Summary

Backfills tests for Testimonial component to bring [coverage up to 100%](https://github.com/primer/brand/actions/runs/16293248338/job/46008099945?pr=1107#step:10:227).

## Contributor checklist

## Reviewer checklist

### DIFF
--- a/packages/react/src/Testimonial/Testimonial.test.tsx
+++ b/packages/react/src/Testimonial/Testimonial.test.tsx
@@ -357,7 +357,6 @@ describe('Testimonial', () => {
     const logo = getByRole('img', {name: logoAlt})
     expect(logo).toBeInTheDocument()
     expect(logo).toHaveAttribute('src', logoSrc)
-    expect(logo).toBeInTheDocument()
   })
 
   it('renders logo container', () => {
--- a/packages/react/src/Testimonial/Testimonial.test.tsx
+++ b/packages/react/src/Testimonial/Testimonial.test.tsx
@@ -6,9 +6,9 @@ import {Testimonial} from './Testimonial'
 
 expect.extend(toHaveNoViolations)
 
-const leftDoubleQuote = '“'
-
 describe('Testimonial', () => {
+  const leftDoubleQuote = '“'
+
   const mockQuote = 'GitHub helps us ensure that we have our security controls baked into our pipelines.'
   const mockName = 'David Ross'
   const mockPosition = 'Staff Security Engineer'
@@ -52,7 +52,7 @@ describe('Testimonial', () => {
     expect(quoteMark).toHaveClass('Testimonial__quoteMark--default')
   })
 
-  it('applies custom className', () => {
+  it('allows forwarding of custom classes', () => {
     const {getByRole} = render(
       <Testimonial className="custom-testimonial">
         <Testimonial.Quote>Quote text</Testimonial.Quote>
@@ -62,6 +62,7 @@ describe('Testimonial', () => {
 
     const figure = getByRole('figure')
     expect(figure).toHaveClass('custom-testimonial')
+    expect(figure).toHaveClass('Testimonial')
   })
 
   it('supports RefObject for Testimonial component', () => {
@@ -332,19 +333,6 @@ describe('Testimonial', () => {
     expect(avatar).toHaveAttribute('src', mockAvatarSrc)
   })
 
-  it('applies testimonial avatar class', () => {
-    const {getByAltText} = render(
-      <Testimonial>
-        <Testimonial.Quote>Quote text</Testimonial.Quote>
-        <Testimonial.Name>Name</Testimonial.Name>
-        <Testimonial.Avatar src={mockAvatarSrc} alt={mockAvatarAlt} />
-      </Testimonial>,
-    )
-
-    const avatarContainer = getByAltText(mockAvatarAlt)
-    expect(avatarContainer).toBeInTheDocument()
-  })
-
   it('uses fixed avatar size of 48px', () => {
     const {getByAltText} = render(
       <Testimonial>
@@ -438,30 +426,6 @@ describe('Testimonial', () => {
     expect(quoteIndex).toBeLessThan(nameIndex)
   })
 
-  it('renders avatar, logo, and name in media section', () => {
-    const {container, getByRole, getByText} = render(
-      <Testimonial>
-        <Testimonial.Quote>Quote text</Testimonial.Quote>
-        <Testimonial.Avatar src={mockAvatarSrc} alt={mockAvatarAlt} />
-        <Testimonial.Logo>
-          <img src="/logo.png" alt="Logo" />
-        </Testimonial.Logo>
-        <Testimonial.Name>{mockName}</Testimonial.Name>
-      </Testimonial>,
-    )
-
-    const mediaSection = container.querySelector('.Testimonial__media')
-    expect(mediaSection).toBeInTheDocument()
-
-    const avatar = getByRole('img', {name: mockAvatarAlt})
-    const logo = getByRole('img', {name: 'Logo'})
-    const name = getByText(mockName)
-
-    expect(mediaSection).toContainElement(avatar)
-    expect(mediaSection).toContainElement(logo)
-    expect(mediaSection).toContainElement(name)
-  })
-
   it('uses semantic HTML structure', () => {
     const {getByRole, getByText} = render(<DefaultTestimonial />)
 
@@ -500,7 +464,7 @@ describe('Testimonial', () => {
     expect(name).toBeInTheDocument()
   })
 
-  it('handles mixed children types', () => {
+  it("filters out children that aren't part of the Testimonial namespace", () => {
     const {getByText, queryByText} = render(
       <Testimonial>
         <div>Invalid child</div>
--- a/packages/react/src/Testimonial/Testimonial.test.tsx
+++ b/packages/react/src/Testimonial/Testimonial.test.tsx
@@ -1,0 +1,516 @@
+import React from 'react'
+import {render} from '@testing-library/react'
+import '@testing-library/jest-dom'
+import {axe, toHaveNoViolations} from 'jest-axe'
+import {Testimonial} from './Testimonial'
+
+expect.extend(toHaveNoViolations)
+
+const leftDoubleQuote = '“'
+
+describe('Testimonial', () => {
+  const mockQuote = 'GitHub helps us ensure that we have our security controls baked into our pipelines.'
+  const mockName = 'David Ross'
+  const mockPosition = 'Staff Security Engineer'
+  const mockAvatarSrc = '/images/avatar-mona.png'
+  const mockAvatarAlt = 'David Ross avatar'
+
+  const DefaultTestimonial = () => (
+    <Testimonial>
+      <Testimonial.Quote>{mockQuote}</Testimonial.Quote>
+      <Testimonial.Name position={mockPosition}>{mockName}</Testimonial.Name>
+      <Testimonial.Avatar src={mockAvatarSrc} alt={mockAvatarAlt} />
+    </Testimonial>
+  )
+
+  it('has no accessibility violations', async () => {
+    const {container} = render(<DefaultTestimonial />)
+    const results = await axe(container)
+    expect(results).toHaveNoViolations()
+  })
+
+  it('renders with default props', () => {
+    const {getByRole, getByText} = render(<DefaultTestimonial />)
+
+    expect(getByRole('figure')).toBeInTheDocument()
+    expect(getByText(mockQuote)).toBeInTheDocument()
+    expect(getByText(mockName)).toBeInTheDocument()
+    expect(getByText(mockPosition)).toBeInTheDocument()
+    expect(getByRole('img', {name: mockAvatarAlt})).toBeInTheDocument()
+  })
+
+  it('renders quote mark with default color', () => {
+    const {getByText} = render(<DefaultTestimonial />)
+
+    const quoteMark = getByText(leftDoubleQuote)
+    expect(quoteMark).toHaveClass('Testimonial__quoteMark--default')
+  })
+
+  it('applies custom className', () => {
+    const {getByRole} = render(
+      <Testimonial className="custom-testimonial">
+        <Testimonial.Quote>Quote text</Testimonial.Quote>
+        <Testimonial.Name>Name</Testimonial.Name>
+      </Testimonial>,
+    )
+
+    expect(getByRole('figure')).toHaveClass('custom-testimonial')
+  })
+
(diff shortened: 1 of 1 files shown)


packages/react/src/Testimonial/Testimonial.test.tsx
```diff
@@ -0,0 +1,516 @@
+import React from 'react'
+import {render} from '@testing-library/react'
+import '@testing-library/jest-dom'
+import {axe, toHaveNoViolations} from 'jest-axe'
+import {Testimonial} from './Testimonial'
+
+expect.extend(toHaveNoViolations)
+
+const leftDoubleQuote = '“'
```
rezrah: Could we move this inside the describe block with the other mock values?

packages/react/src/Testimonial/Testimonial.test.tsx
```diff
@@ -0,0 +1,516 @@
...
+  })
+
+  it('renders quote mark with default color', () => {
+    const {getByText} = render(<DefaultTestimonial />)
+
+    const quoteMark = getByText(leftDoubleQuote)
+    expect(quoteMark).toHaveClass('Testimonial__quoteMark--default')
+  })
+
+  it('applies custom className', () => {
```
rezrah:
```suggestion
  it('allows forwarding of custom classes', () => {
```

packages/react/src/Testimonial/Testimonial.test.tsx
```diff
@@ -0,0 +1,516 @@
...
+    const {getByRole} = render(
+      <Testimonial className="custom-testimonial">
+        <Testimonial.Quote>Quote text</Testimonial.Quote>
+        <Testimonial.Name>Name</Testimonial.Name>
+      </Testimonial>,
+    )
+
+    expect(getByRole('figure')).toHaveClass('custom-testimonial')
```
rezrah: Could we also ensure custom classes don't overwrite the default classes please? One of the general edge case tests that should apply to all components, as it's led to bug reports in the past.

Example:

```suggestion
    const el = getByRole('figure');
    expect(el).toHaveClass('custom-testimonial')
    expect(el).toHaveClass('Testimonial')'
```

Also just to note that for readability and style preference, can we extract the element matcher outside of assertion please

packages/react/src/Testimonial/Testimonial.test.tsx
```diff
@@ -0,0 +1,516 @@
...
+  })
+
+  it('supports RefObject for Testimonial component', () => {
+    const refObject = React.createRef<HTMLElement>()
```
rezrah: Can we prefer useRef instead of createRef for these tests to be more representative of end-user usage?

packages/react/src/Testimonial/Testimonial.test.tsx
```diff
@@ -0,0 +1,516 @@
...
+      <Testimonial className="custom-testimonial">
+        <Testimonial.Quote>Quote text</Testimonial.Quote>
+        <Testimonial.Name>Name</Testimonial.Name>
+      </Testimonial>,
+    )
+
+    expect(getByRole('figure')).toHaveClass('custom-testimonial')
+  })
+
+  it('supports RefObject for Testimonial component', () => {
```
rezrah: For conciceness, and could we remove from other case names too please.

```suggestion
  it('supports RefObject refs', () => {
```

packages/react/src/Testimonial/Testimonial.test.tsx
```diff
@@ -0,0 +1,516 @@
...
+  it('applies custom quote mark color', () => {
+    const {getByText} = render(
+      <Testimonial quoteMarkColor="blue">
+        <Testimonial.Quote>Quote text</Testimonial.Quote>
+        <Testimonial.Name>Name</Testimonial.Name>
+      </Testimonial>,
+    )
+
+    const quoteMark = getByText(leftDoubleQuote)
+    expect(quoteMark).toHaveClass('Testimonial__quoteMark--blue')
```
rezrah: In some other tests, I like to put `expectedColor` at the top to set expectations early. Can we do the same here please. 

E.g. 

```suggestion
    const expectedColor = 'blue'
    
    const {getByText} = render(
      <Testimonial quoteMarkColor={expectedColor}>
        <Testimonial.Quote>Quote text</Testimonial.Quote>
        <Testimonial.Name>Name</Testimonial.Name>
      </Testimonial>,
    )

    const quoteMark = getByText(leftDoubleQuote)
    expect(quoteMark).toHaveClass(`Testimonial__quoteMark--${expectedColor}`)

```

packages/react/src/Testimonial/Testimonial.test.tsx
```diff
@@ -0,0 +1,516 @@
...
+        </Testimonial.Quote>
+        <Testimonial.Name>Name</Testimonial.Name>
+      </Testimonial>,
+    )
+
+    const quoteText = getByText('This is', {exact: false})
+    expect(quoteText).toHaveClass('Testimonial-quote--muted')
+  })
+
+  it('applies muted variant when quote contains emphasized text', () => {
```
rezrah:
```suggestion
  it('applies muted text color when the quote contains emphasized text', () => {
```

packages/react/src/Testimonial/Testimonial.test.tsx
```diff
@@ -0,0 +1,516 @@
...
+        <Testimonial.Avatar src={mockAvatarSrc} alt={mockAvatarAlt} />
+      </Testimonial>,
+    )
+
+    const avatar = getByRole('img', {name: mockAvatarAlt})
+    expect(avatar).toBeInTheDocument()
+    expect(avatar).toHaveAttribute('src', mockAvatarSrc)
+  })
+
+  it('applies testimonial avatar class', () => {
```
rezrah: What is this testing for? It feels like it could be rolled into the tests above? Could consolidate all individual tests for avater into one case?

packages/react/src/Testimonial/Testimonial.test.tsx
```diff
@@ -0,0 +1,516 @@
...
+    const name = getByText(mockName)
+
+    const figureChildren = Array.from(figure.children)
+    const quoteIndex = figureChildren.findIndex(child => child.contains(quote))
+    const nameIndex = figureChildren.findIndex(child => child.contains(name))
+
+    expect(quoteIndex).toBeLessThan(nameIndex)
+  })
+
+  it('renders avatar, logo, and name in media section', () => {
```
rezrah: This one feels a little like implementation detail. It seems like we're expecting elements to render in a specific, generic div but not sure if that's necessary. wdyt?

packages/react/src/Testimonial/Testimonial.test.tsx
```diff
@@ -0,0 +1,516 @@
...
+        <Testimonial.Name>Name</Testimonial.Name>
+      </Testimonial>,
+    )
+
+    expect(getByRole('figure')).toBeInTheDocument()
+    expect(getByText('Quote text')).toBeInTheDocument()
+    expect(getByText('Name')).toBeInTheDocument()
+  })
+
+  it('handles mixed children types', () => {
```
rezrah:
```suggestion
  it('filters out children that aren't part of the Testimonial namespace', () => {
```

packages/react/src/Testimonial/Testimonial.test.tsx
joshfarrant: I can do, but here I'm specifically testing the Testimonial component (as opposed to one of the compound components), so I'd lean towards being explicit here. For example, we have all of these tests within this test suite

```
it('supports functional ref for Testimonial component')
it('supports functional ref for Testimonial.Name component')
it('supports functional ref for Testimonial.Logo component')
it('supports functional ref for Testimonial.Quote component')
```

Happy to change if you'd like, although this could be a good use-case for nested `describe` blocks so we can keep test names clean whilst still clearly scoping them to their related component.

packages/react/src/Testimonial/Testimonial.test.tsx
joshfarrant: I went with `createRef` here for a couple of reasons.

1. To keep things consistent with [how we're testing in the Checkbox component](https://github.com/primer/brand/blob/main/packages/react/src/forms/Checkbox/Checkbox.test.tsx#L90)
2. Testing with `useRef` is a bit awkward as `useRef` is a hook, so will need to be called from within a component. The following would work, but it doesn't feel great IMO
```tsx
  it('supports RefObject for Testimonial component', () => {
    let refObject

    const TestComponent = () => {
      refObject = React.useRef<HTMLElement>(null)

      return (
        <Testimonial ref={refObject}>
          <Testimonial.Quote>Quote text</Testimonial.Quote>
          <Testimonial.Name>Name</Testimonial.Name>
        </Testimonial>
      )
    }

    const {getByRole} = render(<TestComponent />)
    const figure = getByRole('figure')
    expect(refObject.current).toBe(figure)
  })
```

Both `useRef()` and `React.createRef()` return the same underlying `RefObject` as far as I'm aware.

If you'd like me to implement the above approach though just let me know and I'm happy to do so.

packages/react/src/Testimonial/Testimonial.test.tsx
joshfarrant: Yep happy to remove it 👍

packages/react/src/Testimonial/Testimonial.test.tsx
joshfarrant: Looks like something I was playing around with then didn't remove, sorry about that, removed

packages/react/src/Testimonial/Testimonial.test.tsx
joshfarrant: I tend to favour the explicit approach for simple tests like this. It's immediately clear we're passing 'blue' and expecting the 'blue' class — we don't have to deal with mentally tracking a variable or building template strings.

Happy to do it if you strongly prefer this approach, but adding unnecessary logic and abstraction to a test feels like it goes against good testing practise to me.

packages/react/src/Testimonial/Testimonial.test.tsx
rezrah: ah cool, makes sense. Happy to keep it this way, but could we please update the first one to be: `it supports functional ref for the base|root (whichever ones makes sense) Testimonial component`. IMO this sets an earlier expectaion that others similar tests will follow for the child components.

packages/react/src/Testimonial/Testimonial.test.tsx
rezrah: Yeah, makes sense. I was wondering if we could do what we did for SubNav [here](https://github.com/primer/brand/blob/main/packages/react/src/SubNav/SubNav.test.tsx#L274).

rezrah: The reason being that I think it mimics how users will use the component. 1) Creating the wrapper 2) Create a ref with the hook 3) pass to our component 4) Doing something with it in their own handler or effect.

packages/react/src/Testimonial/Testimonial.test.tsx
rezrah: Sounds good, fine to keep it this way. Might come down to a style preference, and I don't have a super strong opinion on it right now. I think in many other parts of the codebase we do what I was suggesting. Could you clarify the part of "goes against good testing practise to me." please, as this feels more like a style preference to me.

packages/react/src/Testimonial/Testimonial.test.tsx
```diff
@@ -0,0 +1,515 @@
...
+        <Testimonial.Quote>Quote text</Testimonial.Quote>
+        <Testimonial.Name>Name</Testimonial.Name>
+      </Testimonial>,
+    )
+
+    const quoteMark = getByText(leftDoubleQuote)
+    expect(quoteMark).toHaveClass('Testimonial__quoteMark--blue')
+  })
+
+  it('applies custom color as CSS variable', () => {
```
rezrah: Is this test case different from the one above? Feels like they are part of the same logic, as they share the same color prop. Could we consider merging them into one test?

packages/react/src/Testimonial/Testimonial.test.tsx
joshfarrant: > Could you clarify the part of "goes against good testing practise to me." please

Here I'm referring to the practise of having tests contain minimal logic and be "obviously correct" at a glance. Adding variables and template literals for single-use values introduces small amounts of logic that require mental parsing, whereas hardcoded values are immediately clear.

As I say, if you did feel strongly that we should continue with your suggested approach then we can do though.

packages/react/src/Testimonial/Testimonial.test.tsx
joshfarrant: Yes good spot, thanks 👍

packages/react/src/Testimonial/Testimonial.test.tsx
```diff
@@ -0,0 +1,515 @@
...
+      <Testimonial>
+        <Testimonial.Quote>Quote text</Testimonial.Quote>
+        <Testimonial.Name>Name</Testimonial.Name>
+        <Testimonial.Avatar src={mockAvatarSrc} alt={mockAvatarAlt} size={64} />
+      </Testimonial>,
+    )
+
+    const avatarContainer = getByAltText(mockAvatarAlt)
+    expect(avatarContainer).toBeInTheDocument()
+  })
```
rezrah: I think the test name here is wrong? Prop value is `64` but test name is `48`?. I might be missing something? 🤔. Should this be asserting for a class or specific dimension property too rather than checking it's in the document?

joshfarrant: The test name could be clearer here. I suppose what I'm actually testing is that the size of the `Avatar` _can't_ be overridden as it's intentionally fixed to 48px. Let me make that a bit clearer, and improve the assertion.

joshfarrant: Improved in 412e52de200b7633f1db7f535451c46bcde24a6a, lmk what you think

packages/react/src/Testimonial/Testimonial.test.tsx
rezrah: Cool, thanks for clarifying. No strong opinions on this beyond the mentioned inconsistency in other test files. Ironically I see it the other, and find it harder to read this way, because I don't really care what the entire className was, just that the expectedValue appears in the final class. Fine to leave as-is for this PR though.

packages/react/src/Testimonial/Testimonial.test.tsx
rezrah: Yep, makes sense 🙌

rezrah (approved):

PR #1110, opened 2025-07-16
add MinimalFooter tests

joshfarrant:
## Summary

Backfills tests for MinimalFooter component to bring coverage up to 100%.

## Contributor checklist

## Reviewer checklist


packages/react/src/MinimalFooter/MinimalFooter.test.tsx
```diff
@@ -0,0 +1,478 @@
...
+describe('MinimalFooter', () => {
+  it('has no accessibility violations', async () => {
+    const {container} = render(<MinimalFooter />)
+    const results = await axe(container)
+    expect(results).toHaveNoViolations()
+  })
+
+  it('renders as a footer element', () => {
+    const {getByRole} = render(<MinimalFooter />)
+    expect(getByRole('contentinfo')).toBeInTheDocument()
```
rezrah: Can we please extract the element matcher outside of the assertion please, like you did further down [here](https://github.com/primer/brand/pull/1110/files#diff-6441670ec35bc1550b32bcb90a2bcace70a7e2719e12ee7ea8a3af9323947e83R57). Helps with readability. The main place I see it useful doing it this way is when we need to reassign the value, but that's not the case here.

```suggestion
    const el = getByRole('contentinfo');
    expect(el).toBeInTheDocument()
```

packages/react/src/MinimalFooter/MinimalFooter.test.tsx
```diff
@@ -0,0 +1,478 @@
...
+    expect(getByText(`© ${currentYear} GitHub. All rights reserved.`)).toBeInTheDocument()
+  })
+
+  it('renders custom copyright statement', () => {
+    const customCopyright = 'Custom Copyright 2024'
+    const {getByText} = render(<MinimalFooter copyrightStatement={customCopyright} />)
+    expect(getByText(customCopyright)).toBeInTheDocument()
+  })
+
+  it('renders copyright statement as React element', () => {
```
rezrah, line 64: This is cool, didn't know we support ReactNode and not just string values.

packages/react/src/MinimalFooter/MinimalFooter.test.tsx
```diff
@@ -0,0 +1,478 @@
...
+  it('renders GitHub logo with custom href', () => {
+    const customHref = 'https://custom.github.com'
+    const {getByRole} = render(<MinimalFooter logoHref={customHref} />)
+    const logoLink = getByRole('link', {name: 'GitHub'})
+    expect(logoLink).toHaveAttribute('href', customHref)
+  })
+
+  it('applies custom className', () => {
+    const {getByRole} = render(<MinimalFooter className="custom-footer" />)
+    expect(getByRole('contentinfo')).toHaveClass('custom-footer')
```
rezrah: Can we update this test to ensure that custom classes don't also overwrite the default classes?

rezrah: A good one to add to the general checklist too, as this bug has snuck in a few times before.

packages/react/src/MinimalFooter/MinimalFooter.test.tsx
```diff
@@ -0,0 +1,478 @@
...
+
+expect.extend(toHaveNoViolations)
+
+// Helper function to filter social links
+const getSocialLinks = (links: HTMLElement[]) => {
+  return links.filter(
+    link =>
+      link.getAttribute('href')?.includes('x.com') ||
+      link.getAttribute('href')?.includes('github.com/github') ||
+      link.getAttribute('href')?.includes('linkedin.com') ||
```
rezrah: Can we do an equality check instead of a partial match? Also wondering if we can just duplicate the same `socialLinkData` here instead, as that would give better confidence.

packages/react/src/MinimalFooter/MinimalFooter.test.tsx
```diff
@@ -0,0 +1,478 @@
...
+  })
+
+  it('applies custom className', () => {
+    const {getByRole} = render(<MinimalFooter className="custom-footer" />)
+    expect(getByRole('contentinfo')).toHaveClass('custom-footer')
+  })
+
+  it('renders all social links by default', () => {
+    const {getAllByRole} = render(<MinimalFooter />)
+    const socialLinks = getSocialLinks(getAllByRole('link'))
```
rezrah: This feels a little hard to read in isolation because the core test logic is abstracted away into getSocialLinks. Related comment here. Can we instead have individual assertions like you did for tests below. It's more verbose, but easier to reason about IMO

packages/react/src/MinimalFooter/MinimalFooter.test.tsx
```diff
@@ -0,0 +1,478 @@
...
+    expect(socialLinks).toHaveLength(8)
+  })
+
+  it('renders specific social links when provided', () => {
```
rezrah:
```suggestion
  it('can optionally render a subset of social links', () => {
```

packages/react/src/MinimalFooter/MinimalFooter.test.tsx
joshfarrant: I've tweaked this based on your feedback, lmk what you think.

We _could_ inline this everywhere, but it will result in us having the following in a few different places.

```tsx
const socialLinks = allLinks.filter(
  link =>
    link.getAttribute('href') === socialLinkData.x ||
    link.getAttribute('href') === socialLinkData.github ||
    link.getAttribute('href') === socialLinkData.linkedin ||
    link.getAttribute('href') === socialLinkData.youtube ||
    link.getAttribute('href') === socialLinkData.facebook ||
    link.getAttribute('href') === socialLinkData.twitch ||
    link.getAttribute('href') === socialLinkData.tiktok ||
    link.getAttribute('href') === socialLinkData.instagram,
)
```

I'm not opposed to having that across multiple tests, but lmk what you think on the current approach as it hopefully strikes a good balance by not abstracting core logic and instead just leveraging [the getByRole query's](https://testing-library.com/docs/queries/byrole#api) name prop with a [TextMatch](https://testing-library.com/docs/queries/about#textmatch) matcher function.

rezrah: I think the new suggestion works. I think the name of the name matcher is throwing me though. Can we go with something a bit more decriptive like `isSupportedSocialLinkName`?

packages/react/src/MinimalFooter/MinimalFooter.test.tsx
```diff
@@ -0,0 +1,601 @@
...
+          <Text>Valid footnote</Text>
+        </MinimalFooter.Footnotes>
+      </MinimalFooter>,
+    )
+
+    const validFootnote = getByText('Valid footnote')
+    expect(validFootnote).toBeInTheDocument()
+  })
+
+  describe('Color modes', () => {
```
rezrah: Can we remove the nested describe block please, the tests should fall under the main describe block. E.g. One component => One describe block

packages/react/src/MinimalFooter/MinimalFooter.test.tsx
```diff
@@ -0,0 +1,601 @@
...
+          removeListener: jest.fn(),
+          addEventListener: jest.fn(),
+          removeEventListener: jest.fn(),
+          dispatchEvent: jest.fn(),
+        })),
+      })
+    })
+
+    afterEach(() => {
+      Object.defineProperty(window, 'matchMedia', {
```
rezrah: Does jest.clearAllMocks() not work here?

packages/react/src/MinimalFooter/MinimalFooter.test.tsx
```diff
@@ -0,0 +1,601 @@
...
+        </MinimalFooter.Footnotes>
+      </MinimalFooter>,
+    )
+
+    const validFootnote = getByText('Valid footnote')
+    expect(validFootnote).toBeInTheDocument()
+  })
+
+  describe('Color modes', () => {
+    let originalMatchMedia: typeof window.matchMedia
```
rezrah: Guessing this is needed for a particular reason? If not, can we match the way we've mocked matchMedia in other components [like this?](https://github.com/primer/brand/blob/main/packages/react/src/ThemeProvider/ThemeProvider.test.tsx#L11)

packages/react/src/MinimalFooter/MinimalFooter.test.tsx
joshfarrant: I found a nicer way to do this which leverages [Testing Library's TextMatch function](https://testing-library.com/docs/queries/about/#textmatch), which essentially lets Testing Library figure it out, lmk what you think

packages/react/src/MinimalFooter/MinimalFooter.test.tsx
joshfarrant: Updated the mocking as you suggested [here](https://github.com/primer/brand/pull/1110#discussion_r2251570431)

packages/react/src/MinimalFooter/MinimalFooter.test.tsx
rezrah: Yeah love this. thanks for updating it. Let's use this approach going forward.

PR #1114, opened 2025-07-17
Add table support to Prose

### LINKED ISSUE #692: [modification request] table styling in Prose
## Problem

SWP asked for getting HTML table support for SEO and solution page using the `Prose` component. It is not supported as of today

Historical context:
- Prose only supports paragraphs, headings, quotes, lists and links today
- We kept the initial scope light, assuming that support for other tags will improve over time on an as-needed basis
- The inherited styling from the ReadMe project got cut from the current Prose component

## Solution

Per discussion between Site and Primer Brand, we would like to implement basic styling (aka- support) for of the `<table>` markup.

Here is an in-situ suggestion using our text, color, border primitives
Image

We cant restrict the amount of column but would highly suggest to limit to 5 or perhaps fix the min-width of `<td>` and have an `overflow-y: scroll`
Image

I do not recommend using a boxed design to minimize the visual noise and fit the editorial styling we have been moving toward brand-wise
Image

## Urgency

No major urgency, SWP is working on scaling the template and could use capability anytime

### LINKED ISSUE #1092: [Feature request] Add table support in Prose component
Hi Primer team,

The SWP team has received a [request](https://github.com/github/marketing-engineering/issues/4660) to add tables to our articles where we leverage the `Prose` component. Currently, the `table` component is not supported. 

Below is a screenshot of how the table currently behaves in the Prose component

Sample table in the prose component

Ideally, the Prose component would support the `table` element so all consumers of the `Prose` will get the styling.

Also, if we could get a prescribed limit of columns/rows, we can work with the editors to try to prevent tables that will not fit in the Prose component.

Thanks as always!

rezrah:
## Summary

Closes #692 
Closes #1092 

Adds table support to the Prose component

🔗 [Preview (docs)](https://primer-d9fa5ebf97-26139705.drafts.github.io/brand/components/Prose/)
🔗 [Storybook examples ](https://primer-d9fa5ebf97-26139705.drafts.github.io/brand/storybook/?path=/story/components-prose-features--default-table)

## List of notable changes:

- **added** basic styling support to Prose stylesheet per the design guidelines [here](https://github.com/primer/brand/issues/692#issuecomment-2313053549)
- **added** two layout modes:
  - Default responsive layout (shrinks horizontally on narrow viewports)
  - Optional stacked column layout, which is opt-in via use of data attributes. This is opt-in because Prose assumes that input HTML will be as plain as possible.
- **added** relevant storybook stories, including translations where relevant

- Go through the examples and verify that the HTML table support works as expected
- Go through visual diffs to ensure content is localizing correctly

## Steps to test:

1. [Review docs example and make sure it works okay](https://primer-58edafb791-26139705.drafts.github.io/brand/)
1. [Review various Storybook examples](https://primer-d9fa5ebf97-26139705.drafts.github.io/brand/storybook/?path=/story/components-prose-features--default-table)

## Contributor checklist:

## Reviewer checklist:

## Screenshots:

<table>
<tr>
</tr>
<tr>
<td valign="top">

Screenshot 2025-07-17 at 16 28 55

 </td>
<td valign="top">

Screenshot 2025-07-17 at 16 31 17

</td>
</tr>
</table>


simmonsjenna (approved):
LGTM 👍

apps/next-docs/content/components/Prose/index.mdx
```diff
@@ -69,6 +69,67 @@ The `Prose` component renders native HTML elements with Primer Brand styling. `P
     <p><code>for-each-ref</code> is extremely useful for listing references, finding which references point at a given object (with <code>--points-at</code>), which references have been merged into a given branch (with <code>--merged</code>), or which references contain a given commit (with <code>--contains</code>).</p>
     <h6>Heading level 6</h6>
     <p>Pellentesque non ornare ligula. Suspendisse nibh purus, pretium id tortor sit amet, tincidunt gravida augue.</p>
+    <table>
```
joshfarrant, line 72: Accessibility-wise this looks good, just one additional thing required when comparing this with the [ARIA Authoring Practises Guide (APG)](https://www.w3.org/WAI/ARIA/apg/patterns/table/#wai-ariaroles,states,andproperties).

> If there is an element in the user interface that serves as a label for the table, [aria-labelledby](https://w3c.github.io/aria/#aria-labelledby) is set on the table element with a value that refers to the labelling element. Otherwise, a label is specified for the table element using [aria-label](https://w3c.github.io/aria/#aria-label).

In this case an `aria-label` is probably the simplest, although a `<caption>` may also satisfy this requirement, although the APG isn't particularly clear on when and why you should use one over the other. In this case I'd imagine that either option would suffice.

```html
<table aria-label="Developer population and growth by country">
<!-- or -->
<caption>Developer population and growth by country</caption>
```

Doing the same in our stories would be an improvement too, although it isn't strictly required as it's not public facing documentation. 

---

Do you think we should add a note to the docs reminding consumers of the component to include a label for the table via `aria-label` or `aria-labelledby`, or even `<caption>` as a bit of accessibility guidance?

packages/react/src/Prose/Prose.features.stories.tsx
```diff
@@ -114,3 +117,291 @@ UnorderedList.decorators = [
...
+    <tr>
+      <th scope="col"><span class="visually-hidden">${t('country')}</span></th>
+      <th scope="col">${t('number_of_developers')}</th>
+      <th scope="col">${t('yoy_growth')}</th>
+    </tr>
+  </thead>
+  <tbody>
+    <tr>
+      <td>${t('singapore')}</td>
+      <td data-label="${t('number_of_developers')}">>${t('developers')}</td>
```
joshfarrant: We're missing some numbers in these examples, e.g.

```suggestion
      <td data-label="${t('number_of_developers')}">>1M ${t('developers')}</td>
```

packages/react/src/Prose/Prose.features.stories.tsx
```diff
@@ -114,3 +117,291 @@ UnorderedList.decorators = [
...
+export const TableWithDataLabelsNarrowView = () => {
+  const {t} = useTranslation('Prose')
+  return <Prose html={dataLabelsMarkup(t)} enableFullWidth />
+}
+TableWithDataLabelsNarrowView.parameters = {
+  viewport: {
+    defaultViewport: 'iphonexr',
+  },
+}
+TableWithDataLabelsNarrowView.storyName = 'With a vertically-stacked table (narrow)'
```
joshfarrant, line 265: Do you know if there have been any previous accessibility investigations into this pattern?

I understand the purpose and the problem we're solving, but rendering a table's rows and columns vertically on narrow viewports makes me wonder if this could be an accessibility issue.

From my research, I spotted [WCAG 1.3.1 Info and Relationships](https://www.w3.org/WAI/WCAG21/Understanding/info-and-relationships.html) and [Failure F43](https://www.w3.org/WAI/WCAG21/Techniques/failures/F43) which I think are both relevant.

We're using the same table markup across all viewports, but the narrow viewport styling creates a visual presentation that no longer matches the semantic table structure. Users get table semantics (headers, columns, cell relationships, etc) but _see_ a vertical list layout, which results in a mismatch between the document structure and visual presentation and could potentially be confusing for Assistive Tech users.

apps/next-docs/content/components/Prose/index.mdx
rezrah: > Do you think we should add a note to the docs reminding consumers of the component to include a label for the table via aria-label or aria-labelledby, or even <caption> as a bit of accessibility guidance?

I've already added a caption to all of the Storybook examples, which I preferred because it's visible compared to the aria-label. I'm missing one in the docs though, so I'll go ahead and add that in 👍

packages/react/src/Prose/Prose.features.stories.tsx
joshfarrant: Looking back on this F43 might not be relevant as the markup is still structured in a way which accurately reflects the content, it's just that visually it's displayed differently. 🤔

SC 1.3.1 might still be relevant though. I've reached out for more info on this, although I'm guessing we've already had this behaviour signed off since we use a similar pattern on the Copilot Plans page.

packages/react/src/Prose/Prose.features.stories.tsx
rezrah: Thanks Josh, that's fixed 👌

packages/react/src/Prose/Prose.features.stories.tsx
rezrah: > I've reached out for more info on this

Where can I follow along @joshfarrant? There's some urgency to landing this feature. If it's going to be a while, I can disable the stories for this optional layout and ship the basic support.

joshfarrant: Discussed in [the a11y channel here](https://github.slack.com/archives/C0FSWLQ0Y/p1753088525478789), but the tl;dr is that what we're doing here is fine 👍 :shipit:

packages/react/src/SubdomainNavBar/SubdomainNavBar.visual.spec.ts-snapshots/Visual-Comparison-SubdomainNavBar-SubdomainNavBar-No-Search-1-linux.png
joshfarrant, line 1: Not sure why this has changed — think it's worth reverting it?

joshfarrant (approved):
Looks good! One small note about a snapshot, but otherwise ✅

PR #1111, opened 2025-07-16
Improvements to various components in Windows High Contrast Mode

joshfarrant:
## Summary

Improvements to a variety of components when viewed in Windows High Contrast Mode.

There are likely further improvements that could be made across the board and with some more thorough style tweaks we could improve our support for High Contrast Mode across PB, but this PR sets a good baseline and addresses some low-hanging fruit. 

## List of notable changes

- `FAQGroup`: Improved selected styles for selected tab.
- `Pagination`: Improved selected styles for current page item.
- `VideoPlayer`: Made all controls and text visible and legible.
- `Checkbox`: Improved focus and checked styles.
- `Radio`: Improved focus and checked styles.
- `TextInput`: Improved focus styles.
- `Textarea`: Improved focus styles.
- `ActionMenu`: Improved focus styles.

## What should reviewers focus on?

- Make sure the improvements work as expected in Windows High Contrast Mode
- Make sure non-High Contrast styles aren't affected
  - _*Some small tweaks have been made to the non-High Contrast Mode VideoPlayer focus styles_

## Steps to test

1. Find a Windows machine (I can share instructions on how to set up a VM)
2. Enable High Contrast Mode
3. Test the components and states affected by this PR

### SWP
- Closes https://github.com/github/primer/issues/5123
### Tetralogical accessibility audit
- Closes https://github.com/github/primer/issues/3866
- Closes https://github.com/github/primer/issues/3790
### HCL accessibility audit
- Closes https://github.com/github/accessibility-audits/issues/10401
- Closes https://github.com/github/accessibility-audits/issues/10375
- Closes https://github.com/github/accessibility-audits/issues/10361
- Closes https://github.com/github/accessibility-audits/issues/10369
- Closes https://github.com/github/accessibility-audits/issues/10397
- Closes https://github.com/github/accessibility-audits/issues/10395

## Contributor checklist

- [x] All new and existing CI checks pass
- [x] Tests prove that the feature works and covers both happy and unhappy paths
- [x] Any drop in coverage, breaking changes or regressions have been documented above
- [x] UI Changes contain new visual snapshots (generated by adding `update snapshots` label to the PR)
- [x] All developer debugging and non-functional logging has been removed
- [x] Related issues have been referenced in the PR description

## Reviewer checklist

- [ ] Check that pull request and proposed changes adhere to our [contribution guidelines](../../CONTRIBUTING.md) and [code of conduct](../../CODE_OF_CONDUCT.md)
[...]

PR #1116, opened 2025-07-18
Regen VideoPlayer snapshots

rezrah:
Regens stale video player snapshots from https://github.com/primer/brand/pull/1111


joshfarrant (approved):
Ah sorry, looks like I forgot to regenerate after rounding off those focus indicators

PR #1121, opened 2025-07-29
Fix layout shift in PricingOptions accordion 

rezrah:
## Summary

Resolves https://github.com/github/primer/issues/5563

Fixes an bug in the PricingOptions component, where feature lists that should be open by default in certain SSR-contexts have a delayed appearance and cause subsequent layout shift.

This PR fixes that issue, while preserving most of the previous behavior. 

## List of notable changes:

- updated pricing list accordions to not rely on `useWindowSIze` to determine initial open/close state. 
- disabled fade in animations from the accordion, as they were contributing to the suboptimal initial opening state
- added a story that handles responsive behaviour, this was missing previously

- The intended fix can only be tested in dotcom
- Instead, we should look for regressions in general pricing option behaviour. This can be done by referencing the preview Storybook examples against the current Live counterparts.

## Steps to test:

1. Open the [preview storybook ](https://primer-a2e9fb099b-26139705.drafts.github.io/brand/storybook/?path=/story/components-pricingoptions--default)
1. Open the [production storybook](https://primer.style/brand/storybook/?path=/story/components-pricingoptions--default)
1. Compare each story and make sure there's no noticeable regression

## Contributor checklist:

## Reviewer checklist:

## Screenshots:

Tested in dotcom 👇. Open state now renderering correctly pre-hydration.

<table>
<tr>
</tr>
<tr>
<td valign="top">

https://github.com/user-attachments/assets/38f6a555-7086-41e7-a8c4-f587fdc3fcad

 </td>
<td valign="top">

https://github.com/user-attachments/assets/63dfbb01-d346-4160-a6b1-6bbb55b0a47c

</td>
</tr>
</table>


packages/react/src/PricingOptions/PricingOptions.tsx
```diff
@@ -382,35 +396,50 @@ type ResponsiveExpandableProps = {
...
+        if (isRegular) return true
+        return false
+      }
 
-    const {isMedium: isRegular, isXLarge: isWide} = useWindowSize()
+      // finally handle responsive objects, and go from largest to smallest viewports
+      const {narrow, regular, wide} = expanded
+      if (isWide) return wide
+      if (isRegular) return regular
+      if (isNarrow) return narrow
```
joshfarrant: Minor thing, but this check is redundant.

If `isNarrow` is truthy then we return `narrow`. If it's not, we still return `narrow`.
```suggestion
```

packages/react/src/PricingOptions/PricingOptions.tsx
```diff
@@ -382,35 +396,50 @@ type ResponsiveExpandableProps = {
...
+  ({children, className, 'data-testid': testId, hasDivider = true, expanded, accordionAs = 'h4', ...rest}, ref) => {
+    const {
+      allFeatureListsExpanded,
+      updateFeatureListExpanded,
+      featureListUserInteracted,
+      setFeatureListUserInteracted,
+    } = usePricingOptions()
+    const {isMedium: isRegular, isXLarge: isWide, isSmall: isNarrow} = useWindowSize()
+
+    React.useEffect(() => {
```
joshfarrant, line 414: As we're modifying the logic around the `expanded` prop, do you think we should add some unit tests to assert that each of the allowed values of the `expanded` prop function as expected?

packages/react/src/PricingOptions/PricingOptions.tsx
```diff
@@ -382,35 +396,50 @@ type ResponsiveExpandableProps = {
...
+      if (expanded === undefined || typeof expanded === 'object') {
+        setFeatureListUserInteracted(false)
+      }
```
joshfarrant, line 417: Could you help me understand the purpose of this logic please?

I understand the mechanics of the logic (if the viewport size changes, or if the value of the `expanded` prop changes to `undefined` or any object, then we set `featureListUserInteracted` back to false) but I don't understand why we need this check.

packages/react/src/PricingOptions/PricingOptions.tsx
rezrah: Yep, removed 👍

packages/react/src/PricingOptions/PricingOptions.tsx
rezrah: Just noticed that we don't have any tests around this 😨, so yeah this should be added but either in a follow up or through the unit test epic. Don't want to hold this change up, let's manual test for now.

packages/react/src/PricingOptions/PricingOptions.tsx
rezrah: Sure thing, I added this to reset the featureListUserInteracted state as windows size changes. I recall there being a bug when window size was being updated while a responsive object was passed. Without resetting this, the following code ended up making subsequent lines from being reacheable, as it stays true:

```
 if (featureListUserInteracted) {
    return allFeatureListsExpanded
}
```

I'm going to leave this as-is unless you see something as blocking here? I don't want to reintroduce that bug.

joshfarrant: No all good, thanks for the explanation.

packages/react/src/PricingOptions/PricingOptions.tsx
joshfarrant: Ok no problem, I'll add the test coverage as part of the Epic once I get to this component.

PR #1132, opened 2025-08-19
Add new Tabs component

rezrah:
## Summary

Adds a Tabs component

Towards https://github.com/github/primer/issues/4783

🔗 [Storybook preview](https://primer-963f46b7ce-26139705.drafts.github.io/brand/storybook/?path=/story/components-tabs--default)
🔗 [Tabs docs](https://primer-963f46b7ce-26139705.drafts.github.io/brand/components/Tabs/)

## List of notable changes:

- [x] Base react component
- [x] Tokens
- [x] Storybook 
  - [x] default and playground
  - [x] features
  - [x] examples
- [x] VRT snapshots
- [ ] Localized VRT
- [x] Documentation for React
- [ ] Interface guidelines
- [x] Unit tests
- [ ] Added to Kitchen Sink

- Review the design to ensure were happy with the different variants, and overall design
- Review the code

## Steps to test:

1. Use [these Storybook examples](https://primer-963f46b7ce-26139705.drafts.github.io/brand/storybook/?path=/story/components-tabs--default)

- Custom homepage tabs
- Custom pricing plans tabs
- Review accessibility

## Contributor checklist:

## Reviewer checklist:

## Screenshots:

<table>
<tr>
<th width="50%"> Dotcom (custom tabs) </th> <th> After (Primer tabs) </th>
</tr>
<tr>
<td valign="top">

https://github.com/user-attachments/assets/7359b946-359a-4b05-be33-f8ae92066650

 </td>
<td valign="top">

https://github.com/user-attachments/assets/00ae5047-5f13-4813-ba22-8ec08c17b7dd

</td>
</tr>
</table>

<table>
<tr>
<th width="50%"> Dotcom (custom tabs) </th> <th> After (Primer tabs) </th>
</tr>
<tr>
<td valign="top">

https://github.com/user-attachments/assets/f110afe8-8ff3-442b-8a29-d9cb1a765c34

 </td>
<td valign="top">

https://github.com/user-attachments/assets/c1f35fe4-91fb-46c9-a1de-d56755c79b69

</td>
</tr>
</table>


simmonsjenna (approved):
Looks good! 👍

packages/react/src/hooks/useTabs.ts
```diff
@@ -199,6 +210,14 @@ export const useTabs = ({
...
       ref: (element: HTMLElement | null) => {
         if (element) registerTab(id, element)
+
+        if (externalRef) {
+          if (typeof externalRef === 'function') {
+            externalRef(element)
+          } else {
+            ;(externalRef as {current: HTMLElement | null}).current = element
+          }
+        }
```
joshfarrant: What do you think about the following approach as an alternative? It achieves the same thing, but doesn't require a type assertion

```suggestion
        if (!externalRef) return

        if (typeof externalRef === 'function') {
          externalRef(element)
          return
        }

        if ('current' in externalRef) {
          externalRef.current = element
          return
        }
```

If you take a look at the type of `externalRef` after that last `if` statement it should have the type of `never`, indicating that our checks are exhaustive and we've handled all possible input values.

wdyt?

packages/react/src/hooks/useTabs.ts
```diff
@@ -188,7 +199,7 @@ export const useTabs = ({
   )
 
   const getTabProps = useCallback(
-    (id: string) => ({
+    (id: string, externalRef?: React.Ref<HTMLElement>) => ({
```
joshfarrant: I think `React.ForwardedRef` might be a more accurate type here. It's essentially the same type as `React.Ref`, with the key difference being that `ref.current` isn't `readonly`, which is what we want here given that we're writing to `ref.current` on L218.
 
It would need to be changed on L30 too.

packages/react/src/hooks/useTabs.ts
joshfarrant, line 1: Some good changes here — thanks!

packages/react/src/Tabs/Tabs.test.tsx
```diff
@@ -0,0 +1,493 @@
...
+    expect(toolbar).toHaveAttribute('aria-label', 'Custom tab controls')
+
+    const prevButton = getByRole('button', {name: 'Go to previous tab'})
+    const nextButton = getByRole('button', {name: 'Go to next tab'})
+
+    expect(prevButton).toBeInTheDocument()
+    expect(nextButton).toBeInTheDocument()
+  })
+
+  it('applies custom className if needed', () => {
```
joshfarrant, line 395: Should we add a check here to make sure that the default class (`"Tabs-container"`) is still applied?

packages/react/src/Tabs/Tabs.tsx
```diff
@@ -0,0 +1,394 @@
...
+        const tabPanelProps = getTabPanelProps(panelId)
+        const isActive = activeTab === panelId
+
+        return cloneElement(child, {
+          ...tabPanelProps,
+          id: child.props.id ?? tabPanelProps.id,
+          key: String(index),
+          value: String(index),
+          activeValue: activeTab,
+          hidden: !isActive,
```
joshfarrant: This is handled by `tabPanelProps`, you don't need to pass it yourself.

packages/react/src/Tabs/Tabs.tsx
```diff
@@ -0,0 +1,394 @@
...
+  const conditionalRender = animation ? (
+    <AnimationProvider autoStaggerChildren={false}>
+      <Animate animate={animation}>{children}</Animate>
+    </AnimationProvider>
+  ) : (
+    children
+  )
+
+  return (
+    <div className={clsx(styles['Tabs__panel'], className)} data-testid={testIds.panel} hidden={hidden} {...rest}>
```
joshfarrant, line 404: `hidden` is handled for you by the hook (and included in `...rest` here) so you don't need to pass it explicitly here

packages/react/src/Tabs/Tabs.test.tsx
```diff
@@ -0,0 +1,493 @@
...
+    const {getByRole} = render(
+      <Tabs aria-label="Test tabs" variant="underline">
+        <Tabs.Item>Tab one</Tabs.Item>
+        <Tabs.Panel>Panel one</Tabs.Panel>
+      </Tabs>,
+    )
+
+    const tabsContainer = getByRole('tablist')
+    expect(tabsContainer).toHaveClass('Tabs--underline')
+  })
```
joshfarrant, line 278: Feel free to ignore this, but you could use `it.each` here to remove some duplication

```suggestion
  it.each(['default', 'accent', 'underline'] as const)('renders %s variant correctly', variant => {
    const {getByRole} = render(
      <Tabs aria-label="Test tabs" variant={variant}>
        <Tabs.Item>Tab one</Tabs.Item>
        <Tabs.Panel>Panel one</Tabs.Panel>
      </Tabs>,
    )

    const tabsContainer = getByRole('tablist')
    expect(tabsContainer).toHaveClass(`Tabs--${variant}`)
  })
```

packages/react/src/Tabs/Tabs.test.tsx
```diff
@@ -0,0 +1,493 @@
...
+    })
+
+    await user.keyboard('{ArrowRight}')
+    expect(tabTwo).toHaveAttribute('aria-selected', 'true')
+
+    await user.keyboard('{ArrowRight}')
+    expect(tabThree).toHaveAttribute('aria-selected', 'true')
+
+    await user.keyboard('{ArrowLeft}')
+    expect(tabTwo).toHaveAttribute('aria-selected', 'true')
```
joshfarrant, line 241: Do you think it would be worth checking that other items aren't aria-selected? And that the panels are being shown/hidden correctly?

packages/react/src/Tabs/Tabs.tsx
```diff
@@ -0,0 +1,394 @@
...
+   */
+  'aria-label': string
+  /**
+   * @param internalAccessibleLabels - Customizable labels for screen readers to improve accessibility. These are applied internally using the aria-label attribute.
+   * @param internalAccessibleLabels.controls - Label for the tab navigation controls, used by screen readers. Default is "Tab navigation controls".
+   * @param internalAccessibleLabels.controlsNext - Label for the "next tab" control button, used by screen readers. Default is "Next tab".
+   * @param internalAccessibleLabels.controlsPrev - Label for the "previous tab" control button, used by screen readers. Default is "Previous tab".
+   */
+  internalAccessibleLabels?: {
+    ['controls']: string
```
joshfarrant: Minor nit for consistency

```suggestion
    controls: string
```

packages/react/src/Tabs/Tabs.test.tsx
```diff
@@ -0,0 +1,493 @@
...
+        <Tabs.Panel>Panel one</Tabs.Panel>
+        <Tabs.Panel>Panel two</Tabs.Panel>
+      </Tabs>,
+    )
+
+    const tabs = getAllByRole('tab')
+    const [tabOne] = tabs
+
+    expect(refObject.current).toBe(tabOne)
+  })
```
joshfarrant, line 438: In other test suites, I've been testing whether the component continues to function when passing both kinds of ref. Specifically I make sure to test around any functionality which explicitly relies on `ref.current` being defined. Up to you ofc, but you might want to do the same here and check that basic component functionality works when passing the two different refs.

packages/react/src/Tabs/Tabs.tsx
```diff
@@ -0,0 +1,394 @@
...
+  onChange?: (activeTab: string) => void
+
+  /**
+   * Additional CSS class names to apply to the tabs container for custom styling.
+   */
+  className?: string
+  /**
+   * Accessible label for the entire tabs component, used by screen readers to describe the tab group.
+   */
+  'aria-label': string
```
joshfarrant: We have this as a required prop, but it's perfectly valid to _not_ pass `aria-label` if you instead pass `aria-labelledby` if there is a visible element (eg a heading) which describes the component.

We always need exactly one of them to be passed though — do you think we should reflect that in the typing here?

```tsx
type LabelOrLabelledBy =
  | {'aria-label': string; 'aria-labelledby'?: never}
  | {'aria-labelledby': string; 'aria-label'?: never}
```

```tsx
export type TabsProps = {
  // ...
} & LabelOrLabelledBy
```

I've got this same check in [my useTabs type improvements PR](https://github.com/primer/brand/pull/1138/files#diff-ab93c6c285d81a44ac26ea3f2573ac6d00c3c5fa4800eefbecd97289771c3e4aR48), so it might be good to have the same guardrail on this component too, wdyt?

packages/react/src/Tabs/Tabs.stories.tsx
```diff
@@ -0,0 +1,68 @@
...
+    variant: 'default',
+    'aria-label': 'Tabs',
+  },
+  argTypes: {
+    align: {
+      options: ['start', 'center'],
+      control: {type: 'inline-radio'},
+    },
+    variant: {
+      options: ['default', 'accent', 'minimal', 'underline'],
```
joshfarrant: Is `"minimal"` a valid variant? Setting it does have an (as it removes some styling), but I can't see it explicitly defined as a valid variant in the Tabs types, or in the CSS

packages/react/src/Tabs/Tabs.test.tsx
joshfarrant, line 1: We've got a few lines of coverage missing from this file (152,161-222,371) any chance you could cover those gaps and bring it up to 100%? 🙏

packages/react/src/Tabs/Tabs.tsx
```diff
@@ -0,0 +1,394 @@
...
+      className,
+      'aria-label': ariaLabel,
+      internalAccessibleLabels = defaultInternalAccessibleLabels,
+      ...props
+    },
+    ref,
+  ) => {
+    const {activeTab, activateTab, getTabListProps, getTabProps, getTabPanelProps} = useTabs({
+      defaultTab: defaultActiveTab ? defaultActiveTab : undefined,
+      onTabActivate: onChange ? (id: string) => onChange(id) : undefined,
```
joshfarrant: Out of interest, is there a reason we can't do the following?

```suggestion
      defaultTab: defaultActiveTab,
      onTabActivate: onChange,
```

packages/react/src/Tabs/Tabs.tsx
```diff
@@ -0,0 +1,394 @@
...
+        const panelId = String(index)
+        const tabPanelProps = getTabPanelProps(panelId)
+        const isActive = activeTab === panelId
+
+        return cloneElement(child, {
+          ...tabPanelProps,
+          id: child.props.id ?? tabPanelProps.id,
+          key: String(index),
+          value: String(index),
+          activeValue: activeTab,
```
joshfarrant: I might be missing something here, but I'm not sure we need to pass `value` or `activeValue` as they don't appear to be used

packages/react/src/Tabs/Tabs.tsx
```diff
@@ -0,0 +1,394 @@
...
+          hidden: !isActive,
+        })
+      })
+
+    useEffect(() => {
+      if (!tabsContainerRef.current || !sliderRef.current) return
+
+      const activeTabButton = tabsContainerRef.current.querySelector(`[data-value="${activeTab}"]`) as HTMLButtonElement
```
joshfarrant: This could potentially be simplified to find the element with `aria-selected="true"`, negating the need for `data-value`

packages/react/src/Tabs/Tabs.tsx
```diff
@@ -0,0 +1,394 @@
...
+      .map((tab, index) => {
+        const tabId = String(index)
+        const originalRef = (tab as React.ReactElement & {ref?: React.Ref<HTMLButtonElement>}).ref
+        const tabProps = getTabProps(tabId, originalRef)
+
+        return cloneElement(tab, {
+          ...tabProps,
+          id: tab.props.id ?? tabProps.id,
+          'aria-controls': tab.props['aria-controls'] ?? tabProps['aria-controls'],
+          value: tabId,
```
joshfarrant, line 132: `value` doesn't seem to be used by the `Tabs.Item` component, so you may be able to remove it

packages/react/src/Tabs/Tabs.tsx
```diff
@@ -0,0 +1,394 @@
...
+const TabsItem = forwardRef<HTMLButtonElement, TabsItemProps & React.ComponentPropsWithoutRef<'button'>>(
+  ({children, value, className, isActive, variant, ...props}, ref) => {
+    const tabRef = useProvidedRefOrCreate(ref)
+
+    return (
+      <button
+        ref={tabRef}
+        className={clsx(
+          styles.Tabs__item,
+          isActive && styles.active,
```
joshfarrant: This `isActive` prop appears to only be used to apply a style to `Tabs.Item` when it's active.

Rather than creating and passing additional props to the component, you could instead target `.Tabs__item[aria-current="true"]` in your styles, which would allow you to delete this prop, and no longer pass it in on L128.

joshfarrant: If you don't want to go with this approach then you might want to update the name of this style as it's currently not using the `Tabs` component namespace. e.g. `styles['Tabs__item--active']`

Not sure if that's an issue or not, but thought I'd mention

packages/react/src/Tabs/Tabs.examples.stories.tsx
```diff
@@ -0,0 +1,256 @@
...
+}
+
+export const WithCustomPanels: StoryFn<typeof Tabs> = args => {
+  const [activeTab, setActiveTab] = useState<string>('0')
+
+  const handleTabChange = useCallback((id: string) => {
+    setActiveTab(id)
+  }, [])
+
+  return (
```
joshfarrant, line 189: The markup in this example is slightly incorrect as both panels (above and below) use the same id (`"panel-[1/2/3]"`).

We could update the ids to be `"panel-[1/2/3]-a"` for the first set of boxes and `"panel-[1/2/3]-b"` for the second set.

We'd also then need to update the `aria-controls` of each tab to be `"panel-[1/2/3]-a panel-[1/2/3]-b"` so the tab correctly points to both elements it controls.

Also tabpanels without any focusable elements should have `tabindex="0"` on them to make them focusable.

apps/next-docs/content/components/Tabs/index.mdx
```diff
@@ -0,0 +1,229 @@
...
+
+```jsx live
+function CustomPanels() {
+  const [activeTab, setActiveTab] = React.useState('0')
+
+  const handleTabChange = React.useCallback(id => {
+    setActiveTab(id)
+  }, [])
+
+  return (
```
joshfarrant, line 99: We're missing `aria-controls` and `aria-labelledby` in this example, as well as ids on the tabpanels. Also tabpanels without any focusable elements should have `tabindex="0"` on them to make them focusable.

packages/react/src/Tabs/Tabs.tsx
```diff
@@ -0,0 +1,394 @@
...
+    return (
+      <div
+        className={clsx(styles['Tabs-container'], styles[`Tabs-container--align-${align}`], className)}
+        ref={ref}
+        data-testid={testIds.root}
+        {...props}
+      >
+        {/* Narrow only indicators and controls */}
+        {tabs.length >= 3 && (
+          <div className={styles.Tabs__controls} role="toolbar" aria-label={internalAccessibleLabels.controls}>
```
joshfarrant: Do you mind if I bring this to the a11y team to get their input on the role and any other suggested attributes/functionality? There are a few things I don't necessarily know the answer to and I'd like some expert input; e.g. the "toolbar" role, whether the next/prev buttons need `aria-disabled` when the first/last page is selected, etc.

I don't think it would need to block this PR

joshfarrant: I should note that, from looking at the [Toolbar Pattern APG](https://www.w3.org/WAI/ARIA/apg/patterns/toolbar/) it does feel to me that `"toolbar"` is an appropriate role.

To fully match the spec defined there we would also need to set an `aria-label` or `aria-labelledby` on the toolbar, and modify the focus behaviour to allow the prev/next controls to be navigated using the arrow keys.

joshfarrant: Ah, although [MDN states
](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/toolbar_role#description)

> Only use the toolbar role to group 3 or more controls.

So this probably does warrant a look from someone more with more a11y knowledge than me

packages/react/src/hooks/useTabs.ts
rezrah: Unfortunately this doesn't work because the ref is being interpreted as readonly, and not mutable.  

`Cannot assign to 'current' because it is a read-only property.`

To make this more obvious, I've instead casted as a MutableRefObject. Should do the same thing at runtime though, while keeping compiler happy.

packages/react/src/hooks/useTabs.ts
rezrah: Ah this explains this comment this issue I had. TIL about this type, this'll come in handy. Thanks.

packages/react/src/hooks/useTabs.ts
rezrah: Update: This now works after changing the base type for externalRef above. Worked like a charm and agree the type narrowing is cleaner

joshfarrant: Great! Yes sorry I should probably have grouped this and my below comment together into one as this depends on that change 👍

apps/next-docs/content/components/Tabs/index.mdx
rezrah: 👍 updated to match the storybook version

packages/react/src/Tabs/Tabs.tsx
rezrah: 👍 thanks, updated the type defs to support aria-labelledby too

packages/react/src/Tabs/Tabs.tsx
rezrah: We can for both yeah. I recall putting this in because there was a type mismatch in the test file, which actualyl turned out to be a problem in the test file which didn't have the right shape of callback. I've fixed this now 👍

packages/react/src/Tabs/Tabs.tsx
rezrah: I needed these at some point, but not anymore. Thanks 👍

packages/react/src/Tabs/Tabs.tsx
rezrah: Much better, thanks

packages/react/src/Tabs/Tabs.tsx
rezrah: Going with second option for now, thanks for flagging

packages/react/src/Tabs/Tabs.examples.stories.tsx
rezrah: Should be fixed now

packages/react/src/Tabs/Tabs.tsx
```diff
@@ -0,0 +1,403 @@
...
+
+export type TabsProps = {
+  /**
+   * Visual style variant of the tabs. Affects the appearance and styling theme.
+   */
+  variant?: 'default' | 'accent' | 'underline'
+  /**
+   * Horizontal alignment of the tab navigation. Defaults to 'start'.
+   */
+  align?: 'start' | 'center'
```
danielguillan, line 60: Could we provide an `end` option so users don't need extra markup to position tabs on the right side of the page? Or is it a use case we want to explicitly disallow?

danielguillan (commented):
This looks great! 

Can we ensure that the rendered height of the component matches other controls, such as the Button and the ActionMenu trigger, at `48px`?

Image

packages/react/src/Tabs/Tabs.tsx
rezrah: @simmonsjenna indicated the need for a `start` option, but `end` wasn't discussed. @simmonsjenna what do you think? Do you ever see that end / right alignment of tabs being needed?

packages/react/src/Tabs/Tabs.tsx
joshfarrant: For posterity, the recommendation after speaking with the a11y team was to remove  `role="toolbar"` and `aria-label={internalAccessibleLabels.controls}`.

packages/react/src/Tabs/Tabs.test.tsx
rezrah: The missing lines of coverage would require too much implementation detail being mocked out. E.g. unmounting of observers, etc. 

I've added two more tests to cover animation behaviours, but otherwise I think all API features are covered.

Feel free to take another look during your sweep. Curious if you see any other opportunities.

Screenshot 2025-09-02 at 16 36 10

apps/next-docs/content/components/Tabs/index.mdx
```diff
@@ -0,0 +1,248 @@
...
+  <Tabs.Panel>
+    <Text>Panel two</Text>
+  </Tabs.Panel>
+  <Tabs.Panel>
+    <Text>Panel three</Text>
+  </Tabs.Panel>
+</Tabs>
+```
+
+### Custom panels
```
joshfarrant, line 89: It would be nice to have a sentence explaining here which explains that, by using custom panels, the user is responsible for handling some of the accessibility functionality, specifically:

- Ensuring `Tabs.Item` have an `aria-controls` attribute which points to the id of the associated panel.
- Ensuring custom panels have an `aria-labelledby` attribute which points to the id of the associated `Tabs.Item`.
- Ensuring custom panels have attribute `role="tabpanel"`
- Ensuring custom panels have a `tabIndex={0}`
- Ensuring custom panels have the attribute `hidden` when they are not the active tab

Optionally, a link to [the relevant section of the APG for the Tabs pattern](https://www.w3.org/WAI/ARIA/apg/patterns/tabs#wai-ariaroles,states,andproperties) could lend useful context too.

If you agree @rezrah then I'll add this in a follow-up PR

packages/react/src/Tabs/Tabs.test.tsx
joshfarrant: Ok no problem, I'll take a look as I think that we can probably cover a bit more than this with some minor tweaks.

I'll take a look and, since this PR is now merged, I'll open a new PR with the coverage improvements.

packages/react/src/Tabs/Tabs.test.tsx
joshfarrant: Did you have any thoughts on this @rezrah? I'm happy to address this in a follow-up PR?

rezrah: Yeah that's a good point, we're missing function ref coverage and/or support here is what you're calling out? As you're planning to open a follow up PR, feel free to add there, or I can look at it later. LMK what you prefer.

joshfarrant: I'll add the functional ref coverage, along with asserting core functionality, in a follow-up PR

PR #1133, opened 2025-08-21
Add tests for Accordion component

joshfarrant:
## Summary

Backfills tests for Accordion component to bring coverage up to 100%.

## Contributor checklist

## Reviewer checklist

### DIFF
--- a/packages/react/src/Accordion/Accordion.test.tsx
+++ b/packages/react/src/Accordion/Accordion.test.tsx
@@ -1,0 +1,535 @@
+import React, {createRef} from 'react'
+import {fireEvent, render} from '@testing-library/react'
+import '@testing-library/jest-dom'
+import userEvent from '@testing-library/user-event'
+
+import {Accordion, AccordionToggleColors} from '../'
+import {axe, toHaveNoViolations} from 'jest-axe'
+
+expect.extend(toHaveNoViolations)
+
+describe('Accordion', () => {
+  it('has no accessibility violations', async () => {
+    const {container} = render(
+      <Accordion>
+        <Accordion.Heading>Test heading</Accordion.Heading>
+        <Accordion.Content>
+          <p>Test content</p>
+        </Accordion.Content>
+      </Accordion>,
+    )
+
+    const results = await axe(container)
+
+    expect(results).toHaveNoViolations()
+  })
+
+  it('applies the correct classes for the default variant', () => {
+    const {getByRole} = render(
+      <Accordion variant="default">
+        <Accordion.Heading>Test heading</Accordion.Heading>
+        <Accordion.Content>Test content</Accordion.Content>
+      </Accordion>,
+    )
+
+    const details = getByRole('group')
+    expect(details).toHaveClass(`Accordion--default`)
+  })
+
+  it('applies the correct classes for the emphasis variant', () => {
+    const {getByRole} = render(
+      <Accordion variant="emphasis">
+        <Accordion.Heading>Test heading</Accordion.Heading>
+        <Accordion.Content>Test content</Accordion.Content>
+      </Accordion>,
+    )
+
+    const details = getByRole('group')
+    expect(details).toHaveClass(`Accordion--emphasis`)
+
+    const summary = getByRole('heading', {name: 'Test heading'}).parentElement
+    expect(summary).toHaveClass('Accordion__summary--emphasis')
+  })
+
+  it('allows forwarding of custom classes', () => {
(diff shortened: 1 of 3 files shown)


packages/react/src/Accordion/Accordion.tsx
```diff
@@ -80,6 +80,7 @@ export const AccordionRoot = forwardRef<HTMLDetailsElement, AccordionRootProps>(
 
     useEffect(() => {
       const detailsElement = ref.current
+      /* istanbul ignore next */
```
joshfarrant: This seems tricky to test, are you happy for me to ignore this particular line?

packages/react/src/Accordion/Accordion.tsx
```diff
@@ -112,6 +113,11 @@ export type AccordionHeadingProps = HTMLAttributes<HTMLElement> & {
   toggleColor?: (typeof AccordionToggleColors)[number]
 }
 
+/**
```
joshfarrant, line 116: Let me know if you'd like this to be captured elsewhere

packages/react/src/Accordion/Accordion.test.tsx
```diff
@@ -0,0 +1,535 @@
...
+
+    expect(ref.current).toBe(details)
+
+    expect(details).not.toHaveAttribute('open')
+
+    await user.click(heading)
+    expect(details).toHaveAttribute('open')
+
+    await user.click(heading)
+    expect(details).not.toHaveAttribute('open')
```
rezrah, line 274: Is the reason for adding these extra assertions that you're checking the passing of ref doesn't cause open/close regressions? Was wondering how they're related to testing the ref.

It feels quite duplicative for every ref test, but could make sense if the goal is to catch regressions through side effects

packages/react/src/Accordion/Accordion.tsx
rezrah: 👍 can you add this to a new tasklist on the same issue please, as remedial follow-up work.

rezrah: You'll need to add a changeset to this PR @joshfarrant. Can you add one explaining it's a noop update to this file.

packages/react/src/Accordion/Accordion.tsx
rezrah: We have added this line on some places, but could you elaborate on why we need to add it here please? Haven't looked into it, so lacking context. It's an effect, so looks relatively important.

packages/react/src/Accordion/Accordion.test.tsx
joshfarrant: Yes exactly, the goal is to make sure that the component still functions in the expected way when refs of different types are passed, given the fact that we rely on the ref internally for some of the component's core functionality.

packages/react/src/Accordion/Accordion.tsx
joshfarrant: The comment ignores the next line from coverage, which in this case is

```tsx
if (!detailsElement) return
```

Practically speaking, `detailsElement` (`ref.current`) is always going to be defined and I don't think there's a way to make it `undefined` in a test without some very unrealistic mocking. If you have a suggestion of how to do so though I'd be happy to explore it.

Another approach could be to use a type assertion

```tsx
const detailsElement = ref.current as HTMLDetailsElement
```

which would negate the need for the `if` and the `istanbul ignore` entirely, but that would result in our code being less defensive, which feels like a step backwards

rezrah: > detailsElement (ref.current) is always going to be defined

I can't recall the exact reason anymore, but it would seem that this early return was very intentional 🤔 

The correct fix here would instead be to wrap all handlers inside an if statement, checking if both ref and ref.current are defined? That seems like a common use-case that we're handling elsewhere too?

Curious what you think. I'm fine ignoring coverage for this, as we have adequate coverage elsewhere now. Mostly wondering if there's any other ways to work around this and not ignore it entirely.

joshfarrant: Ok, let's wrap it in an `if` rather than doing the early return.

Do you want me to do that as part of this work, or as a follow-up?

rezrah: I think it's fine to mark as a follow up item (in the tasklist), and add a TODO annotation 👍

joshfarrant: Great, ty, will do

PR #1139, opened 2025-08-26
Add tests for Checkbox component

joshfarrant:
## Summary

Backfills tests for Checkbox component to bring coverage up to 100%.

## Contributor checklist

## Reviewer checklist

### DIFF
--- a/packages/react/src/forms/Checkbox/Checkbox.test.tsx
+++ b/packages/react/src/forms/Checkbox/Checkbox.test.tsx
@@ -74,13 +74,17 @@ describe('Checkbox', () => {
   })
 
   it('remains clickable when passing an arbitrary function as a ref', async () => {
-    // eslint-disable-next-line @typescript-eslint/no-empty-function
-    const {getByRole} = render(<Checkbox value="test" ref={() => {}} />)
+    const user = userEvent.setup()
+    const mockRef = jest.fn()
+
+    const {getByRole} = render(<Checkbox value="test" ref={mockRef} />)
 
     const checkbox = getByRole('checkbox')
     expect(checkbox).not.toBeChecked()
 
-    const user = userEvent.setup()
+    expect(mockRef).toHaveBeenCalledTimes(1)
+    expect(mockRef).toHaveBeenCalledWith(checkbox)
+
     await user.click(checkbox)
     expect(checkbox).toBeChecked()
   })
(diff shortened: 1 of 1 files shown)


packages/react/src/forms/Checkbox/Checkbox.test.tsx
```diff
@@ -103,4 +107,321 @@ describe('Checkbox', () => {
...
+    expect(checkbox.indeterminate).toBe(false)
+    expect(checkbox).not.toBeChecked()
+    expect(span).not.toHaveClass('Checkbox--indeterminate')
+    expect(checkbox).not.toHaveAttribute('aria-checked')
+    expect(dashPath).not.toBeInTheDocument()
+    expect(checkmarkPath).toBeInTheDocument()
+  })
+
+  it('has indeterminate state when indeterminate prop is true', () => {
+    const {getByRole, container} = render(<Checkbox value="test" indeterminate onChange={jest.fn()} />)
```
rezrah: Stylistic nit, could we pull jest.fn out to a separate variable instead of calling it directly here? You did something like this further down too:

```suggestion
const mockOnChange = jest.fn();
const {getByRole, container} = render(<Checkbox value="test" indeterminate onChange={mockOnChange} />)
```

packages/react/src/forms/Checkbox/Checkbox.test.tsx
```diff
@@ -103,4 +107,321 @@ describe('Checkbox', () => {
...
+
+    const checkbox = getByRole('checkbox') as HTMLInputElement
+
+    const span = container.querySelector('span.Checkbox')
+    const checkmarkPath = container.querySelector('path.Checkbox-checkmark-path')
```
rezrah, line 193: Do you think this might be too much implementation detail? Is it needed for coverage? Wondering if just checking the classname is sufficient?

joshfarrant: Ideally I think we'd have a class or test-id on the indeterminate dash icon which allowed us to identify it, but the current implementation of the component doesn't include such a class, so we're left with this way of selecting it. Agreed it isn't ideal.

I don't think it's strictly _needed_ for coverage, but asserting that the correct icon is displayed on indeterminate checkboxes is something that we should be testing imo.

We could either leave it as-is, remove it, or add a class/test-id to the indeterminate icon.

Let me know your preferred approach here @rezrah and I'll make any required changes before merging.

joshfarrant: As discussed with @rezrah, we'll keep this as-is for now and add a follow-up task to add some way of uniquely identifying the dash icon to the `Checkbox` component, at which point we can refactor this test.

PR #1146, opened 2025-09-03
Joshfarrant/tabs component follow up

joshfarrant:
## Summary

Improves to Tabs component test suite and simplifies Tab centering logic in component.

## List of notable changes

- Replaces manual tab centering logic with a call to `activeTabButton.scrollIntoView()` to remove manual logic. This has the added benefit of allowing this scroll behaviour to be easily tested.
- Adds functional ref testing
- Adds some core functionality testing to refs tests
- Replaces 3 variant tests with a single `it.each()`
- Adds some more assertions to some test cases
- Explicitly ignores some lines from coverage to achieve 100% coverage. Ignored lines are
  - 3 guards at the top of `useEffect` calls, which are difficult to test and practically are unlikely to happen.
  - 1 function which relies heavily on layout measurement and isn't practical to test.

- Check that, on narrow viewports, the active tab is still scrolled to the center as it was before.
- Are you happy with all of the test changes? If not, I'm happy to revert 👍

## Steps to test

1. Take a look at the component on narrow viewports
2. Check that the active tab is still automatically scrolled to the center

- Follow on work from outstanding comments in https://github.com/primer/brand/pull/1132

## Contributor checklist

## Reviewer checklist


packages/react/src/Tabs/Tabs.tsx
```diff
@@ -150,6 +150,7 @@ const _TabsRoot = forwardRef<HTMLDivElement, TabsProps>(
       })
 
     useEffect(() => {
+      /* istanbul ignore next */
```
joshfarrant, line 154: I've not added any _why_ comments for these ignores as I think adding ignores to null checks at the top of `useEffect` blocks could become somewhat common in the codebase as we work towards 100% coverage.

These ignores _could_ be removed by inverting the `if` logic and wrapping the whole block in the if — [as we discussed here](https://github.com/primer/brand/pull/1133#discussion_r2290657405) — but I wonder whether the negative readability impact of that additional layer of nesting is actually worse than this innocuous `istanbul ignore` 🤔

lmk your opinion on this

packages/react/src/Tabs/Tabs.tsx
```diff
@@ -181,24 +183,9 @@ const _TabsRoot = forwardRef<HTMLDivElement, TabsProps>(
...
-      const tabCenter = tabLeft + tabWidth / 2
-
-      const targetScrollLeft = tabCenter - containerWidth / 2
-
-      const maxScrollLeft = tabsContainer.scrollWidth - containerWidth // defines boudnary
-      const finalScrollLeft = Math.max(0, Math.min(targetScrollLeft, maxScrollLeft))
-
-      tabsContainer.scrollTo({
-        left: finalScrollLeft,
+      activeTabButton.scrollIntoView({
```
joshfarrant, line 186: I've tested this and the functionality appears to be the same, but another set of 👀 would be appreciated

PR #1157, opened 2025-09-09
Improve minimal Card outline

### LINKED ISSUE #1143: 🐛 [BUG] - Border radius on minimal card variants are too large
### Describe the bug

Update the border radius on Minimal card variants from `xlarge` to `medium`.

`xlarge` is appropriate for normal card variants where a border is present but not minimal.

Site Design confirmed that `medium` is the correct size value.

| Actual | Expected |
|--------|--------|
| Image | Image | 

### Reproduction steps

```bash
1. Go to this page: https://github.com/resources/articles/ai
2. Click on or focus on a Card item
3. Observe that the focus outline is too close to the hovered thumbnail
```

### Expected behavior

```bash
1. Card with `minimal` variants use `--brand-borderRadius-medium` CSS variable instead of `--brand-borderRadius-xlarge`
```

### Screenshots

```bash

```

### Browsers

_No response_

### OS

_No response_

danielguillan:
## Summary

Resolves #1143. Reduces the border radius of the `minimal` `Card` variant, which appears on the focus ring.

## List of notable changes:

- Updated `.Card__outer` styling to apply a smaller border radius for the `minimal` variant.

- [Storybook](https://primer-2c763065f7-26139705.drafts.github.io/brand/storybook/?path=/story/components-card-features--minimal)

- #1143

## Contributor checklist:

## Reviewer checklist:

## Screenshots:

<table>
<tr>
</tr>
<tr>
<td valign="top">

screenshot-QsErYXTM-000397@2x

 </td>
<td valign="top">

screenshot-fwgAmyA6-000395@2x

</td>
</tr>
</table>

### DIFF
--- a/packages/react/src/Card/Card.module.css
+++ b/packages/react/src/Card/Card.module.css
@@ -101,7 +101,9 @@
 }
 
 .Card__outer:has(.Card--variant-minimal):focus-within {
-  outline-offset: var(--base-size-16);
+  outline-offset: var(
+    --base-size-12
+  ); /* This value is adjusted so that, when combined with the container's border radius, the total border radius equals `var(--brand-borderRadius-medium)` */
 }
 
 .Card--icon {


packages/react/src/Card/Card.module.css
```diff
@@ -92,6 +92,10 @@
   border-radius: var(--brand-borderRadius-xlarge);
 }
 
+.Card__outer:has(.Card--variant-minimal) {
+  border-radius: var(--brand-borderRadius-small);
```
rezrah, line 96: Was this supposed to be medium instead of small? I recall some conversation about this on the call on using medium. Personal opinion, but I think it looks better as small as possible. cc. @simmonsjenna.

danielguillan: I see why this is confusing. We’re using ⁠`--brand-borderRadius-small` (`4px`) on the container, but because of the `16px` offset of the outline ring, the effective border radius is `20px`. 

That's close to `--brand-borderRadius-medium` (`16px`) but not exact.

There are two approaches we could take, as border-radius cannot be controlled directly in CSS outlines:

- We could use a pseudo-element to draw the focus ring. But I'd not go that route since that would require refactor and introduce some risk. as pseudo-elements are already used for the effects.
- A simpler approach would be to match the medium size radius by compensating the offset instead:

```diff
+.Card__outer:has(.Card--variant-minimal) {
+  border-radius: var(--brand-borderRadius-small);
+}

.Card__outer:has(.Card--variant-minimal):focus-within {
-  outline-offset: var(--base-size-16);
+  outline-offset: var(--base-size-12); /* A comment to document that this value influences the effective border radius */
}
```

What do you think?

rezrah: Yeah that works me for personally (simpler approach), thanks @danielguillan

simmonsjenna: @rezrah The reason for achieving `--brand-borderRadius-medium (16px)` here is because the image inside the card uses `--brand-borderRadius-small`, and the container (outline) should have a larger border-radius since that is usually how we approach them (e.g. parent should have a slightly larger border radius than its children to create an optically balanced curve).

PR #1163, opened 2025-09-30
Add FlexSection component

joshfarrant:
## Summary

FlexSection implementation.

All components have been transferred over to this template and are configurable via the Storybook controls.

There are some bugs, however I believe these are faithful to the source FlexSection implementation. As this is a demo/internal component, typing is a bit looser than it would otherwise be if this were a public component.

Take a look at [the new FlexSection component in Storybook](https://primer-5e296a4fbb-26139705.drafts.github.io/brand/storybook/?path=/story/recipes-flextemplate-flexsection--flex-section-default), play with the controls, see if you spot any issues.

-
-

## Contributor checklist:

## Reviewer checklist:


packages/react/src/recipes/FlexTemplate/fixtures/images/security/river-bg-dark-1.png
rezrah, line 1: @joshfarrant could you remove these assets from here please, as they aren't watermarked and I don't believe all are used in flex section. They may have been carried over from the original feature branch from Stefan, but we can repurpose existing assets if needed.

packages/react/src/recipes/FlexTemplate/FlexSection/FlexSection.tsx
```diff
@@ -0,0 +1,1011 @@
...
+                      style={{
+                        backgroundColor:
+                          statistics.fields.variant === 'boxed'
+                            ? `var(--brand-color-canvas-${backgroundColor === 'default' ? 'subtle' : 'default'})`
+                            : undefined,
+                      }}
+                    >
+                      <Statistic.Heading as="p">
+                        $2M+
+                        {statistics.fields.showHeadingFootnotes ? (
```
rezrah: @joshfarrant FYI these aren't rendering correctly. Let me know if I'm misunderstanding something though.


<table>
<tr>
<th> Dotcom flexsection </th> <th> This PR </th>
</tr>
<tr>
<td valign="top">

Screenshot 2025-10-02 at 11 19 51



 </td>
<td valign="top">


Screenshot 2025-10-02 at 11 18 56


</td>
</tr>
</table>

packages/react/src/recipes/FlexTemplate/FlexSection/FlexSection.tsx
```diff
@@ -0,0 +1,1011 @@
...
+                  ))
+                : []}
+            </RiverStoryScroll>
+          )}
+
+          {testimonials && testimonials.fields.testimonialCount > 0 && (
+            <FlexSectionTestimonials testimonials={testimonials} className={styles.normalizeMargin} />
+          )}
+
+          {breakoutBanner && (
```
rezrah, line 469: Breakout banner looking a bit squished here? Is that how it's configured in Dotcom?

Screenshot 2025-10-02 at 11 26 52

packages/react/src/recipes/FlexTemplate/FlexSection/FlexSection.tsx
```diff
@@ -0,0 +1,1011 @@
...
+                    <Card.Description>
+                      Duis aute irure dolor in reprehenderit in voluptate velit esse cillum dolore.
+                    </Card.Description>
+                  </Card>
+                </Box>
+              </Grid.Column>
+            </Grid>
+          )}
+
+          {featuredBento && (
```
rezrah, line 363: Is this supposed to be so narrow? 

Screenshot 2025-10-02 at 11 30 00



Dotcom storybook for featured bento looks like like this, so wondering if we could match that? Or am I looking at a different component?

Screenshot 2025-10-02 at 11 30 17

packages/react/src/recipes/FlexTemplate/FlexSection/FlexSection.tsx
```diff
@@ -0,0 +1,1011 @@
...
+                        </PricingOptions.Footnote>
+                      )}
+                    </PricingOptions.Item>
+                  ))}
+                </PricingOptions>
+              </Grid.Column>
+            </Grid>
+          )}
+
+          {segmentedControlPanel && (
```
rezrah, line 636: Interesting, this is the custom tabs component but I can't see it in the storybook flex section story: https://ui.githubapp.com/storybook/?path=/story/others-mkt-swp-contentful-contentfulflexsection--default

It feels way too low down to be here. @joshfarrant is this confirmed part of flex section? Feels like it should sit inbetween multiple flex sections inside of flex template instead 🤔

packages/react/src/recipes/FlexTemplate/fixtures/images/security/river-bg-dark-1.png
joshfarrant: Good spot, thanks

packages/react/src/recipes/FlexTemplate/FlexSection/FlexSection.tsx
joshfarrant: This is one of those cases where I believe this is a bug in the original implementation and/or with our `Statistic` component.

The `ContentfulStatistic` can render multiple children into `<Statistic.Heading>` depending on whether footnotes are enabled.

If you do try to render multiple children into `<Statistic.Heading>`, then you get something like this.

```tsx
<Statistic>
  <Statistic.Heading>+25%<span>Test</span></Statistic.Heading>
  <Statistic.Description>increase in developer speed</Statistic.Description>
</Statistic>
```

image

Try that in [our docs](https://primer.style/brand/components/Statistic/#default) to reproduce the issue.

I'll open an issue for that `Statistic` bug, and I'll remove the `statisticsShowHeadingFootnotes` control from the story now, which means we no longer trigger that bug.

packages/react/src/recipes/FlexTemplate/FlexSection/FlexSection.tsx
joshfarrant: Yes this is taken straight from [the FlexSection implementation](https://github.com/github/github-ui/blob/main/packages/swp-core/components/contentful/ContentfulFlexSection/ContentfulFlexSection.tsx#L205-L211). It's not configured to show in those stories so it doesn't show by default. If you were to update the `component` object with the correct attributes then it would show.

It does effectively sit between multiple flex sections as each of the rendered tabpanels [is its own FlexSection](https://github.com/github/github-ui/blob/main/packages/swp-core/components/contentful/ContentfulSegmentedControlPanel/ContentfulSegmentedControlPanel.tsx#L46). I'm not sure that level of infinite nesting is practical to implement in our Storybook though, but I can look into it if you think it's worthwhile?

packages/react/src/recipes/FlexTemplate/FlexSection/FlexSection.tsx
joshfarrant: Thanks! Took a while but I got to the bottom of it 😄

packages/react/src/recipes/FlexTemplate/FlexSection/FlexSection.tsx
joshfarrant: Ah thanks, I was missing a class here.

packages/react/src/recipes/FlexTemplate/FlexSection/FlexSection.tsx
rezrah: 🤔 Yeah that's not supposed to work that way, so I don't think this is a bug in our component. 

Could you clarify why multiple children are being passed to statistic via Flex Section, and how? I'm looking at dotcom Storybook and can't see any examples of this. Only ones where the description is being used correctly.👇 

Screenshot 2025-10-03 at 15 35 19

AFAICT the component itself is also [configured to render rich text directly as a single React node](https://github.com/github/github-ui/blob/main/packages/swp-core/components/contentful/ContentfulStatistic/ContentfulStatistic.tsx#L62), so this shouldn't technically happen. Let me know if I'm misunderstanding something though @joshfarrant.

rezrah: Update to ☝️, I just spotted the issue now, so you can ignore some of the above. It's the footnotes thing 😞. It looks like passing a Fragment doesn't work either, which is weird. That bug will need to be fixed before we can enable an inline foot note like this. Fine to disable for now 👍

packages/react/src/recipes/FlexTemplate/FlexSection/FlexSection.tsx
rezrah: > I'm not sure that level of infinite nesting is practical to implement in our Storybook though, but I can look into it if you think it's worthwhile?

Is it possible to create an additional story that replicates this page:

https://github.com/features/copilot/plans

and put it under FlexTemplate > Examples > Pricing

packages/react/src/recipes/FlexTemplate/FlexSection/FlexSection.stories.tsx
```diff
@@ -0,0 +1,766 @@
...
+      description: 'Show bottom border',
+      table: {category: 'Visual Settings'},
+    },
+  },
+}
+
+export default meta
+
+// @ts-expect-error As this is purely a demo component the types are not exact
+export const FlexSectionDefault: StoryFn<typeof FlexSection> = args => <FlexSection component={createMockData(args)} />
```
rezrah: Could we switch this to CSF3 please? We're currently trying to move over to CSF3 everywhere.

packages/react/src/recipes/FlexTemplate/FlexSection/FlexSection.stories.tsx
```diff
@@ -0,0 +1,766 @@
...
+        verticalGap: args.verticalGap,
+        hasBorderBottom: args.hasBorderBottom,
+        enableRiverStoryScroll: args.enableRiverStoryScroll,
+      },
+    },
+  },
+})
+
+const meta: Meta<FlexSectionStoryArgs> = {
+  title: 'Recipes/FlexTemplate/FlexSection',
```
rezrah, line 230: Could we also get stories for FlexTemplate, which is just flex section + a hero so we start seeing this flexsection in practice?

E.g. Could we have a story here: recipes/flextemplate/examples/models

Which reproduces this page https://github.com/features/models, which I believe was built on flex template and should look like-for-like if flexsection is correctly implemented.

packages/react/src/recipes/FlexTemplate/FlexSection/FlexSection.tsx
joshfarrant: The Copilot plans page isn't a FlexSection page, so replicating that isn't possible with the current FlexSection. I believe it can be configured somewhat by Contentful, but it's not fully built using it.

packages/react/src/recipes/FlexTemplate/FlexSection/FlexSection.tsx
```diff
@@ -0,0 +1,1075 @@
...
+      width: 418,
+    },
+    right: {
+      width: 433,
+    },
+  },
+}
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+function FlexSectionTestimonials({testimonials, className}: any) {
```
rezrah: Could we move components like these into a `/components` subfolder given there's quite a few of them in this file?

packages/react/src/recipes/FlexTemplate/FlexSection/FlexSection.stories.tsx
```diff
@@ -0,0 +1,767 @@
+import React from 'react'
```
rezrah, line 1: Could you please add some visual snapshots for these stories @joshfarrant, we can see visual diffs in future PR's then.

PR #1167, opened 2025-10-05
Fix link to grid vs stack documentation

rbadillap:
This pull request updates the documentation for the Stack component to point to the correct location for the "grid vs stack" section. The current link seems to be outdated and redirects to the homepage.

Documentation update:

* Updated the Stack component usage guidelines to link to the correct "grid vs stack" section under `/brand/layout/Grid/#grid-vs-stack` instead of `/components/Grid#grid-vs-stack`.

### DIFF
--- a/apps/next-docs/content/layout/Stack/index.mdx
+++ b/apps/next-docs/content/layout/Stack/index.mdx
@@ -10,7 +10,7 @@ tab-label: Guidelines
 
 ## Usage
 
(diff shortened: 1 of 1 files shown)


rezrah (approved):
Thanks for fixing this @rbadillap. Some CI checks are failing as this is a fork (we'll fix that later), but the code change itself looks good 👍

PR #1214, opened 2025-12-19
[A11Y] Increase line number contrast in IDE

danielguillan:
## Summary

Resolves https://github.com/github/accessibility-audits/issues/14453. Improves the visual contrast of IDE line numbers. Changing color from `var(--brand-color-text-subtle)` to `var(--brand-color-text-muted)` in `IDE.module.css`.

| before | after |
|---|---|
| image    | image |

Snapshots haven't been updated, likely due to a very small threshold difference between the two colors.

### DIFF
--- a/.changeset/sixty-nails-sit.md
+++ b/.changeset/sixty-nails-sit.md
@@ -2,4 +2,4 @@
 '@primer/react-brand': patch
 ---
 
-Increase IDE line numbers contrast
+Increase `IDE` component line numbers contrast


.changeset/sixty-nails-sit.md
```diff
@@ -0,0 +1,5 @@
+---
+'@primer/react-brand': patch
+---
+
+Increase IDE line numbers contrast
```
rezrah:
```suggestion
Increase `IDE` component line numbers contrast
```

rezrah (approved):
LGTM. While the VRT is passing, the thresholds are low enough to not pick up subtle changes to color like this. Could we try regenerating them to see if there is an update @danielguillan?

danielguillan:
@rezrah I tried regenerating the snapshots again, but there's no updated.

PR #1210, opened 2025-12-17
Update Docs to use React v19

danielguillan:
## Summary

Towards https://github.com/github/primer/issues/6134. Updates React, NextJS, Doctocat, and Primer React to latest versions in docs.

## List of notable changes:

- Upgrade `@primer/doctocat-nextjs` from 0.7.0 to 0.8.0                                    
- Upgrade `@primer/react` from 37.11.2 to 38.6.1                                           
- Upgrade `next` from 15.5.2 to 15.5.9                                                     
- Update React version constraints to support React 18 and 19                              
- Remove `styled-components` dependency (no longer required by @primer/react 38.x)         
- Remove `motion-dom` override (no longer needed)                                          
- Refactor color-scales components to use CSS modules instead of @primer/react Box with sx prop

## What should reviewers focus on?

- [Review updated docs](https://primer-03973f7832-26139705.drafts.github.io/brand/)
- Ensure [Color primitives page](https://primer-03973f7832-26139705.drafts.github.io/brand/primitives/color/) works as expected

## Supporting resources (related issues, external links, etc):

- https://github.com/github/primer/issues/6134
- [Doctocat v0.8.0](https://github.com/primer/doctocat-nextjs/releases/tag/%40primer%2Fdoctocat-nextjs%400.8.0)

PR #1222, opened 2026-01-14
Fixes CI issues introduced in next-docs React 19 update

danielguillan:
Fixes CI failures introduced in https://github.com/primer/brand/pull/1210 by aligning React type dependencies in the `next-docs` workspace and updating configuration for proper dependency resolution.


package.json
```diff
@@ -35,7 +35,6 @@
     "format": "npm run format --workspaces --if-present",
     "format:fix": "npm run format:fix --workspaces --if-present",
     "lint": "npm run lint --workspaces --if-present",
-    "postinstall": "npm install --prefix apps/next-docs",
```
rezrah, line 38: Did we need to remove this @danielguillan? It was a convenience feature.

danielguillan: The issue with that script is that it treats the `next-docs` directory as a separate project and creates its own `package-lock.json`, which causes conflicts now that dependenciess are hoisted to the root. We could use actual workspaces for this, or perhaps we simply keep the script and pass the `npm_config_package_lock=false` environment variable? What do you think?

rezrah: ohh of course, I forgot we hoist those dependencies now. This is totally fine 👍

rezrah (approved):

PR #1229, opened 2026-01-26
Rename font files for better URL compatibility

danielguillan:
Fixes an issue with font file paths by replacing square brackets with hyphens, ensuring better compatibility.

Related: https://github.com/github/primer/issues/6303

### DIFF
--- a/.changeset/smooth-doors-dance.md
+++ b/.changeset/smooth-doors-dance.md
@@ -1,5 +1,5 @@
 ---
-'@primer/react-brand': major
+'@primer/react-brand': minor
 ---
 
 Renamed font files to replace URL-unsafe characters (square brackets and commas) with hyphens


packages/react/CHANGELOG.md
```diff
@@ -41,7 +41,7 @@
 
     ```diff
     - @primer/react-brand/fonts/MonaSans.woff2
-    + @primer/react-brand/fonts/MonaSansVF[wdth,wght,opsz].woff2
+    + @primer/react-brand/fonts/MonaSansVF-wdth,wght,opsz.woff2
```
rezrah: This is still valid for the 0.61.0 tag, so can we leave this as-is please?

packages/fonts/MonaSansVF-wdth-wght-opsz.woff2
rezrah, line 1: @danielguillan can we replace the commas too with dashes also?

rezrah: Theory being that the comma might have the same issue

packages/react/CHANGELOG.md
danielguillan: You're right, good catch! Thanks!

.changeset/smooth-doors-dance.md
```diff
@@ -0,0 +1,15 @@
+---
+'@primer/react-brand': major
```
rezrah: @danielguillan could we change this to `minor` please, this will turn us v1 😅 

```suggestion
'@primer/react-brand': minor
```

danielguillan: Yes, sorry just catched the mistake 😅

PR #1355, opened 2026-05-14
Add image background support in RiverAccordion

danielguillan:
## Summary

Towards https://github.com/github/brand-experience/issues/73

Adds image background support to `RiverAccordion.Visual` and makes the full-bleed visual background the default in the gridline variant.

## List of notable changes:

- Make the new full-bleed background the default visual layout for the gridline variant
- Added new examples stories and docs
- Fixed an issue where images get cropped on all variants

- Verify that the variant aligns closely with [the design](https://www.figma.com/design/66oUQUIP0bfGXTW4iuAwe1/?node-id=908-148819&t=7I3Pgj9cdZOLgxyu-11)
- Verify that there are no regressions on existing variants

## Steps to test:

1. Check the [Storybook examples](https://primer-2009f33e75-26139705.drafts.github.io/brand/storybook/?path=/story/components-riveraccordion-features-gridline-variants--grid-line-with-background)
2. Review the [docs](https://primer-2009f33e75-26139705.drafts.github.io/brand/components/RiverAccordion/#visual-with-background)

- [Designs](https://www.figma.com/design/hUT1gnfCRTX7iPZSjOg8Vp/?node-id=3863-84497&t=fCfS3CIrHR3k73Jk-11)

## Contributor checklist:

## Reviewer checklist:

## Screenshots:
screenshot-ek95yi4g-001050@2x


packages/react/src/river/RiverAccordion/RiverAccordion.tsx
```diff
@@ -256,10 +258,24 @@ const RiverAccordionContent = ({className, children, ...props}: RiverAccordionCo
   )
 }
 
-export type RiverAccordionVisualProps = React.HTMLAttributes<HTMLDivElement>
+export type RiverAccordionVisualProps = React.HTMLAttributes<HTMLDivElement> & {
+  /**
+   * Applies a background color with padding around the media.
+   */
+  imageBackgroundColor?: RiverAccordionVisualBackgroundColor
```
rezrah: @danielguillan does this need to be a prop? I'm wondering if we enable the background color by default when parent variant is gridline? I'm struggling to see when someone will use this in gridline layouts without a background color now, and do we want to encourage that? Similarly, if someone isn't using gridline variant, does this work as expected? (Can't see stories for that).

If we can justify keeping it, i think we should call it `backgroundColor` for consistency with other similar props in other components. We don't need to qualify it with `image`. Let me know if I'm missing anything here.

danielguillan: @rezrah, there's a precedent. This follows the exact same API we have in `River` ([example](https://stunning-chainsaw-j82glqz.pages.github.io/?path=/story/recipes-flexsuite-overview--ai)). 

That said, I completely agree with the concern. I suggest we proceed with this approach for now to unblock upcoming work and revisit it in a follow-up. Happy to work on that right after the next release.

rezrah: Thanks @danielguillan 

> and revisit it in a follow-up. Happy to work on that right after the next release.

Sorry, could you elaborate on what that means? Does that mean we'd ship `imageBackgroundColor` and change it to something else later? Wouldn't that be a breaking change that we can avoid by doing it correctly now? 

Also checking we even need a prop for this too:

> does this need to be a prop? I'm wondering if we enable the background color by default when parent variant is gridline.

Could the background color be baked into the `gridline` variant, or does it need to be configurable?

I'm totally fine with shipping this, but if it means we need to change the prop name in the next release we should get it right now.

danielguillan: > Could the background color be baked into the gridline variant, or does it need to be configurable?

This is definitely my preference, but it requires a visual breaking change, along with an API breaking change in River if we follow the same rule. I'm checking with @jesussandreas to see if we can make this the default behavior for the gridline variant and remove the other layout. I'll update as needed.

rezrah: cool. Let me know when the PR is ready for re-review and I'll ✅. Thanks for following up on this.

jesussandreas: @danielguillan yeah, my preference is that the media background should be baked into the gridline variant by default.

I don’t think we need to expose it as a separate prop unless there’s a clear use case where gridline should not have that background treatment. From a design perspective, the background feels like part of what makes the gridline variant work, so making it automatic should help keep usage more consistent.

danielguillan: Thanks everyone! Implemented the new approach. Ready for re-review.

rezrah (approved):

PR #1363, opened 2026-06-03
[Brand Refactor] Multiple component + recipe updates

rezrah:
## Summary

Resolves https://github.com/orgs/github/projects/23995/views/1?pane=issue&itemId=159782690
Resolves https://github.com/orgs/github/projects/23995/views/1?pane=issue&itemId=186264527
Resolves https://github.com/orgs/github/projects/23995/views/1?pane=issue&itemId=185437164
Resolves https://github.com/orgs/github/projects/23995/views/1?pane=issue&itemId=185436819
Resolves https://github.com/orgs/github/projects/23995/views/1?pane=issue&itemId=185436530
Resolves https://github.com/orgs/github/projects/23995/views/1?pane=issue&itemId=185436530
Resolves https://github.com/orgs/github/projects/23995/views/1?pane=issue&itemId=185436322
Resolves https://github.com/orgs/github/projects/23995/views/1?pane=issue&itemId=185436229
Resolves https://github.com/orgs/github/projects/23995/views/1?pane=issue&itemId=185434044
Resolves https://github.com/orgs/github/projects/23995/views/1?pane=issue&itemId=185432260
Resolves https://github.com/orgs/github/projects/23995/views/1?pane=issue&itemId=185430224

Lots of different updates in this PR that are clearing items in the QA board. Mostly minor spacing/sizing/type changes but some bigger changes to Accordion and base size tokens (additive).

Towards https://github.com/github/brand-marketing-design/issues/2606

## List of notable changes:

- Updates the Accordion animation to match the [Figma prototype](https://www.figma.com/proto/hUT1gnfCRTX7iPZSjOg8Vp/branch/9dhtrAAxDMJwsDi0gWByWK/BrandX-UI-System?page-id=5166%3A13470&node-id=5291-7037&viewport=-5577%2C187%2C0.3&t=D8IweXFrRVROGzAh-1&scaling=min-zoom&content-scaling=fixed&starting-point-node-id=5291%3A17622&show-proto-sidebar=1)
- Updated the Flex Suite overview page, where qa items are relevant to only that
- Updated components in non-breaking way everywhere else
- Added new base size tokens for `--base-size-60` and `--base-size-88`
- 🆕 Updated River accordion to use triangles instead of chevrons

- Go through the QA items if you're feeling brave and cross-check changes are there
- Otherwise just focus on a code review and check the preview.
- Review the FAQ animations though, they've been completely updated.

## Steps to test:

1. See new accordion animations [here](https://primer-67f583a5f4-26139705.drafts.github.io/brand/storybook/?path=/story/components-faq--default)
1. See the revised FlexSuite [here](https://primer-67f583a5f4-26139705.drafts.github.io/brand/storybook/?path=/story/recipes-flexsuite-overview--ai)

## Contributor checklist:

## Reviewer checklist:

## Screenshots:

<table>
<tr>
</tr>
<tr>
<td valign="top">

https://github.com/user-attachments/assets/9fb2c33f-cce2-44a1-97cd-b86743f4cea3

 </td>
<td valign="top">

https://github.com/user-attachments/assets/2f792b88-f251-46f3-828c-b6bbaeb5e724

</td>
</tr>
</table>

<table>
<tr>
</tr>
<tr>
<td valign="top">

Screenshot 2026-06-04 at 09 49 22

 </td>
<td valign="top">

Screenshot 2026-06-04 at 09 49 03

</td>
</tr>
</table>


packages/react/src/Accordion/Accordion.tsx
```diff
@@ -52,15 +52,87 @@ const useAccordionContext = (): AccordionContextType => {
...
+
+        isClosingRef.current = true
+        details.classList.add(styles['Accordion--closing'])
+
+        closeAnimationTimeout.current = setTimeout(() => {
+          details.open = false
+          isClosingRef.current = false
+          details.classList.remove(styles['Accordion--closing'])
+          closeAnimationTimeout.current = null
+        }, 400)
```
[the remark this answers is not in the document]
rezrah: Fixed

packages/react/src/Accordion/Accordion.module.css
```diff
@@ -1,8 +1,29 @@
 .Accordion {
+  --brand-Accordion-animation-duration-open: 900ms;
+  --brand-Accordion-animation-duration-close: 400ms;
+  --brand-Accordion-animation-duration: var(--brand-Accordion-animation-duration-open);
+  --brand-Accordion-animation-easing-open: cubic-bezier(0.04, 0.62, 0.16, 0.94);
+  --brand-Accordion-animation-easing-close: cubic-bezier(0.04, 0.62, 0.16, 0.94);
+  --brand-Accordion-animation-easing: var(--brand-Accordion-animation-easing-open);
```
danielguillan: The accordion opening animation feels a bit too slow, especially in the final part of the easing curve. It sort of lingers near the end of the transition, which makes the interaction feel less polished. It seems much slower than the one in the Figma prototype. I’d suggest shortening the duration slightly and/or using an easing curve that resolves more decisively. cc @jesussandreas

packages/react/src/Accordion/Accordion.tsx
```diff
@@ -30,6 +30,7 @@ import styles from './Accordion.module.css'
 export type AccordionRootProps = DetailsHTMLAttributes<HTMLDetailsElement> & {
   children: ReactElement<AccordionHeadingProps | AccordionContentProps>[]
   variant?: 'default' | 'emphasis'
+  disableAnimation?: boolean
```
danielguillan, line 35: We're missing the docs and storybook control for the new prop.

packages/react/src/Accordion/Accordion.tsx
```diff
@@ -50,17 +51,118 @@ const useAccordionContext = (): AccordionContextType => {
...
+        }
+
+        const summary = target.closest('summary')
+
+        if (summary?.parentElement === details) {
+          event.preventDefault()
+
+          if (!isClosingRef.current) {
+            closeWithAnimation(details)
+          }
```
danielguillan, line 171: This seems to be breaking consumer's custom `onClick`. I think it should run the close logic only after consumer's `onClick` has had a chance to call `preventDefault()`. For example: the [Always Exactly One Panel Open](https://primer-67f583a5f4-26139705.drafts.github.io/brand/storybook/?path=/story/components-accordion-features--always-exactly-one-panel-open) story breaks, since clicking on the open panel closes it.

packages/react/src/Accordion/Accordion.module.css
danielguillan: Also, should we perhaps move these variables into their own design-tokens JSON file?

packages/react/src/constants.tsx
```diff
@@ -19,5 +19,5 @@ export const BiColorGradients = ['blue-purple', 'green-blue', 'pink-blue', 'purp
 export const TriColorGradients = ['green-blue-purple'] as const
 
 // TODO: consider generating the scale from  style dictionary and serve from the brand-primitives package
-export const BaseSizeScale = [4, 8, 12, 16, 20, 24, 28, 32, 36, 40, 44, 48, 64, 80, 96, 112, 128] as const
+export const BaseSizeScale = [4, 8, 12, 16, 20, 24, 28, 32, 36, 40, 44, 48, 60, 64, 80, 88, 96, 112, 128] as const
```
danielguillan, line 22: Nit: my OCD can’t unsee this. 😅

Could we add the missing values so the 4px increments are complete up to 64, and the 8px increments stay consistent after that?

```suggestion
export const BaseSizeScale = [4, 8, 12, 16, 20, 24, 28, 32, 36, 40, 44, 48, 52, 56, 60, 64, 72, 80, 88, 96, 112, 128] as const
```

rezrah: Yeah i definitely see the issue, but just to clarify, this would expand the current token API beyond what is immediately required. AFAIK we don't need those additional values yet, but we could create load bearing dependency on them. Not sure we should do that right now. Do you have any thoughts on that Dani? Do you feel strongly enough that the trade off is worth. it, or are you happy leaving it as a nit that we can live with for now?

packages/react/src/Accordion/Accordion.tsx
rezrah: Nice catch Dani, thanks for this. Yeah there was an issue, which is now fixed but how did you spot that out of interest? was it comparing the prod to preview? Just curious as this slipped through my tests and want to put some checks in place for next time.

packages/react/src/Accordion/Accordion.module.css
rezrah: Removed most of these. Only one remains which is used for programmatic height calculation, recommend keeping that local here.

packages/react/src/Accordion/Accordion.tsx
rezrah: Added storybook story, but we don't have public docs for Accordion so left that out. Or did you mean FAQ @danielguillan?

jesussandreas (approved):
heelllll yeeeeea

packages/react/src/Accordion/Accordion.tsx
danielguillan: I spotted the issue by clicking around the [Always Exactly One Panel Open](https://primer-67f583a5f4-26139705.drafts.github.io/brand/storybook/?path=/story/components-accordion-features--always-exactly-one-panel-open). I had to compare it with the production version though because I wasn't completely confident that the behavior was correct, even though the story name is quite clear.

packages/react/src/Accordion/Accordion.tsx
danielguillan: Should we document this prop in FAQ, or do we expect it to always animate? We can omit it if that's the only behavior we anticipate everywhere.

packages/react/src/constants.tsx
danielguillan: I see the concern, but I don’t think this introduces much risk in practice. These values would still be part of the same scale logic rather than arbitrary additions, and the consistent increments make the scale more predictable and useful from a systems perspective.

That said, I’m fine not introducing them now if we don’t have an immediate need. We can leave it as-is and revisit later. 👍

danielguillan (approved):
🚀 Great updates!

PR #1364, opened 2026-06-03
[Brand Refactor] Tiles gridlines update

danielguillan:
## Summary

Resolves https://github.com/orgs/github/projects/23995/views/5?pane=issue&itemId=185435286

Addresses QA feedback from the [board](https://github.com/orgs/github/projects/23995/views/5) for the `Tiles` gridline variant.

Towards https://github.com/github/brand-marketing-design/issues/2606

## List of notable changes:

- Hides the lateral gridline container borders when a `Tiles` row is underfilled at the largest breakpoint.
- Keeps full-row and multi-row gridline behavior unchanged for default and compact layouts.
- Adds a Storybook story covering underfilled default and compact gridline rows.

- Verify that underfilled `Tiles` gridline rows do not show lateral container borders at the largest breakpoint.
- Verify that full default and compact gridline rows still show the expected outer gridlines.

## Steps to test:

1. Open the `Gridlines Underfilled Rows` story in the [Storybook preview](https://primer-f15e1f51c5-26139705.drafts.github.io/brand/storybook/?path=/story/components-tiles-features--gridlines-underfilled-rows).
1. Verify the default and compact examples at the largest breakpoint.

## Contributor checklist:

## Reviewer checklist:

## Screenshots:

image


packages/react/src/Tiles/Tiles.module.css
```diff
@@ -63,6 +64,13 @@
   right: 0;
 }
 
+@media screen and (min-width: 63.25rem) {
+  .Tiles--variant-gridlines.Tiles--layout-default .Tiles-grid:not(:has(> .Tiles-item:nth-child(6))),
```
rezrah, line 68: oof that's a rough selector, can we do anything to make this simpler @danielguillan? Also `:has` has performance problems in Safari. Could we pass a class through React to cover >6 items instead?

danielguillan: I’d prefer to keep this encapsulated in CSS, even at the expense of a slightly more complex selector, because the condition only exists to support this specific layout rule and can stay in a single file. We could add a comment to clarify what the selector does.

I also don’t think this is likely to cause meaningful performance issues. The `:has()` usage is scoped to the Tiles component rather than being applied globally to something broad like `body` or `*`, and this component does not appear in large numbers on any page, which are the more common scenarios where `:has()` can become problematic. The selector also checks for a direct child, so the matching work stays small and predictable.

A React + CSS approach would work, and I can change it if you feel strongly about this approach, but it would require the component logic to know about a presentational item-count condition that is only relevant to this CSS rule.

rezrah: Okay thanks for clarification. Let's move ahead. The readability isn't _that_ bad, it's just in start contrast to how complex selectors get around the codebase, but I can live with it for sure.

> but it would require the component logic to know about a presentational item-count condition that is only relevant to this CSS rule

🤔 I'm probably misunderstanding but we do that in a lot of components already. E.g. SubNav items, Tabs, AnchorNav.. and anything that requires different presentational and interaction behaviour based on children length. You'll see this in our test approach too, where we treat our classes as 'business logic' to assert against. 

We're fine from a precedent perspective so just don't let that hold you back in case that was a factor.

rezrah (approved):

PR #1367, opened 2026-06-09
Fix docs links

danielguillan:
## Summary

Fixes internal links in the docs so they work in both local dev and the production base path (`/brand`), and resolves intermittent scroll jank on navigation.

## List of notable changes:

- **fixed** internal markdown links to use `next/link`, so they respect the base path and resolve to current docs routes
- **fixed** broken link targets in component docs (e.g. `/components/Heading`, `/components/Link`, `EyebrowText`, `Popover`) and outdated Storybook story IDs
- ~**added** trailing-slash normalization for internal links to match the `trailingSlash: true` static export and avoid 308 redirects~
- **fixed** scroll-to-top on navigation by adding `data-scroll-behavior="smooth"` to `<html>` (so Next.js disables smooth scroll during route transitions)

- Internal links resolve correctly in dev and in production under the `/brand` base path
- Clicking links reliably scrolls to the top, while in-page anchor/TOC links still scroll smoothly

## Steps to test:

1. Run the docs locally + check the [preview deployment](https://primer-bb7e877539-26139705.drafts.github.io/brand/)
2. Open a component page, scroll down, on "Related components" links → page should navigate directly and jump to the top
3. Click a table-of-contents/heading anchor → should scroll smoothly within the page

- https://nextjs.org/docs/messages/missing-data-scroll-behavior

### DIFF
--- a/apps/next-docs/content/components/ActionMenu/index.mdx
+++ b/apps/next-docs/content/components/ActionMenu/index.mdx
@@ -5,7 +5,7 @@ keywords: ['dropdown', 'select', 'menu', 'options', 'popup']
 show-tabs: false
 ready: true
 figma: 'https://www.figma.com/file/BJ95AjraesmRCWsKA013GS/Primer-Brand?node-id=4912%3A32799'
-source: https://github.com/primer/brand/tree/main/packages/react/src/ActionMenu/ActionMenu.tsx
+source: https://github.com/primer/brand/blob/main/packages/react/src/ActionMenu/ActionMenu.tsx
 storybook: '/brand/storybook/?path=/story/components-actionmenu--default'
 thumbnail: '/images/thumbnails/action-menu-thumbnail.png'
 thumbnail_darkMode: '/images/thumbnails/action-menu-thumbnail-dark.png'
--- a/apps/next-docs/content/components/Card/react.mdx
+++ b/apps/next-docs/content/components/Card/react.mdx
@@ -5,7 +5,7 @@ keywords: ['card', 'link', 'summary', 'content', 'information']
 show-tabs: true
 tab-label: React
 ready: true
-source: https://github.com/primer/brand/tree/main/packages/react/src/Card/Card.tsx
+source: https://github.com/primer/brand/blob/main/packages/react/src/Card/Card.tsx
 storybook: '/brand/storybook/?path=/story/components-card--playground'
 a11yReviewed: true
 ---
--- a/apps/next-docs/content/components/Card/react.mdx
+++ b/apps/next-docs/content/components/Card/react.mdx
@@ -225,7 +225,7 @@ Forwards all the props from the [Image component](/components/Image), including 
 | `variant`  | <CardLabelVariantsProp />  | `token` | `false`  | Presentation of the label above the heading |
 
 Forwards common span props such as `className`, `id`, and `ref`. `variant="token"` renders a
-[Token](/components/Token), and `variant="accent-text"` renders [EyebrowText](/components/EyebrowText)
+[Token](/components/Token), and `variant="accent-text"` renders [EyebrowText](https://primer.style/brand/storybook/?path=/story/components-eyebrowtext--default)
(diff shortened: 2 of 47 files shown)


apps/next-docs/content/components/ActionMenu/index.mdx
```diff
@@ -5,7 +5,7 @@ keywords: ['dropdown', 'select', 'menu', 'options', 'popup']
 show-tabs: false
 ready: true
 figma: 'https://www.figma.com/file/BJ95AjraesmRCWsKA013GS/Primer-Brand?node-id=4912%3A32799'
-source: https://github.com/primer/brand/tree/main/packages/react/src/ActionMenu/ActionMenu.tsx
+source: https://github.com/primer/brand/blob/main/packages/react/src/ActionMenu/ActionMenu.tsx
```
rezrah, line 8: Using blob is the right call as it's a file, but they still resolve to the same place. Was the redirect causing an issue @danielguillan or is this just cleanup?

apps/next-docs/content/components/Card/react.mdx
```diff
@@ -225,7 +225,7 @@ Forwards all the props from the [Image component](/components/Image), including
 | `variant`  | <CardLabelVariantsProp />  | `token` | `false`  | Presentation of the label above the heading |
 
 Forwards common span props such as `className`, `id`, and `ref`. `variant="token"` renders a
-[Token](/components/Token), and `variant="accent-text"` renders [EyebrowText](/components/EyebrowText)
+[Token](/components/Token), and `variant="accent-text"` renders [EyebrowText](https://primer.style/brand/storybook/?path=/story/components-eyebrowtext--default)
```
rezrah: @danielguillan can we make these env-relative like our Storybook links at the top of page? E.g. in preview these should link to the preview deployment not prod. We should be able to reuse the functionality from the page header links.

apps/next-docs/mdx-components.js
```diff
@@ -22,7 +22,33 @@ import NextLink from 'next/link'
 // eslint-disable-next-line import/extensions
 import {Pre} from './src/components/Pre/Pre.tsx'
 
-export const Link = ({href = '', ...props}) => <NextLink href={href} {...props} />
+const isInternalPath = href => href.startsWith('/') && !href.startsWith('//')
+
+const hasFileExtension = pathname => /\/[^/]+\.[^/]+$/.test(pathname)
+
+const withTrailingSlash = href => {
```
rezrah: I don't think you need to worry about this as the primer.style server will auto-append the trailing slash for you.

apps/next-docs/mdx-components.js
```diff
@@ -22,7 +22,33 @@ import NextLink from 'next/link'
...
+  }
+
+  return `${pathname}/${suffix}`
+}
+
+export const Link = ({href = '', children, ...props}) => {
+  const normalizedHref = isInternalPath(href) ? withTrailingSlash(href) : href
+
+  return isInternalPath(href) ? (
+    <NextLink href={normalizedHref} {...props}>
```
rezrah: Does this our links will no longer look like PB-styled links? Can we restore the styles from Link here, even if it's manual? Or temporarily use the legacyBehaviour prop to use the Link internally for presentational reasons?

apps/next-docs/content/components/ActionMenu/index.mdx
danielguillan: Some redirects where causing a 404, so I took the opportunity to clean up all the links

apps/next-docs/content/components/Card/react.mdx
danielguillan: Good idea! I'll look into it. 👍

apps/next-docs/mdx-components.js
danielguillan: It's removed.

rezrah (approved):
🚀

PR #1370, opened 2026-06-11
Add a JSON color map export to brand-primitives

stefankp:
## Summary

Adds a JSON color map export to `@primer/brand-primitives`. The build now emits `lib/design-tokens/json/colors.json` — a one-dimensional JSON file listing every brand color custom property, keyed by its CSS variable name (including the leading `--`) and mapped to its literal value.

This gives lint rules (and other tooling) an authoritative list of brand color tokens to validate against, without recreating that list from file paths or token names on every release. It's data-only — no rule logic or config ships here.

## List of notable changes:

- **added** a generic `json/one-dimensional` Style Dictionary format (`src/formats/json-one-dimensional.js`) that flattens tokens to a sorted `--<name>: value` map. It carries no type logic, so the same format can back future per-type exports.
- **added** `src/filters/isColorValue.js`, which selects color tokens by inspecting the emitted value (hex / rgb / rgba / hsl / hsla / color-mix / var / `transparent` / `currentColor`) rather than by file path, so non-color values living in `colors.*` files (gradients, shadows, filters, `none`, `url()`) are excluded.
- **added** a build step in `scripts/build-tokens.js` that writes `lib/design-tokens/json/colors.json` (516 color tokens).
- **removed** the earlier color-specific format and its build-time `$type`-from-path annotation.
- **added** unit tests for the format and the filter.

- Color selection happens by value shape in the export, not in source — brand tokens keep their legacy `value`/`dark` shape with no `$type`, so token sources are untouched. When tokens later gain real `$type`, the export can read it and the heuristic goes away.
- `IDE.default.editor.tabs.bgColor` is included because its light value is a color (only its dark value is a gradient), and we filter on the emitted (light) value. Flagging as an accepted edge case.
- No `exports` map was added (the package publishes via `files: ["lib"]`), so the artifact resolves at `@primer/brand-primitives/lib/design-tokens/json/colors.json`.

## Steps to test:

1. `npm run build -w @primer/brand-primitives`
1. Confirm `packages/design-tokens/lib/design-tokens/json/colors.json` exists — 516 entries, each a `--`-prefixed name mapping to a color value.
1. `npm test -w @primer/brand-primitives` — format and filter unit tests pass.

## Contributor checklist:


packages/design-tokens/src/formats/stylelint-colors.js
```diff
@@ -0,0 +1,48 @@
...
+
+    for (const token of dictionary.allTokens) {
+      if (!isColorTokenFilePath(token.filePath)) continue
+
+      const name = `--${token.name}`
+      entries[name] = {
+        name,
+        value: token.value,
+        $type: 'color',
+        filePath: token.filePath,
```
rezrah: @stefankp do you need the `filePath` property for your linting? It "leaks" out internal source path, and is extra noise that I think we don't need here. If you're not directly referencing the property, could we remove it here please?

```suggestion
```

packages/design-tokens/src/formats/stylelint-colors.js
```diff
@@ -0,0 +1,48 @@
...
+    const entries = {}
+
+    for (const token of dictionary.allTokens) {
+      if (!isColorTokenFilePath(token.filePath)) continue
+
+      const name = `--${token.name}`
+      entries[name] = {
+        name,
+        value: token.value,
+        $type: 'color',
```
rezrah: Think CCR pointed this out already, but this isn't always a color unit, it can be an arbitrary string. Recommend removing this, unless you absolutely need it @stefankp. 

```suggestion
```

.changeset/brand-stylelint-colors-export.md
```diff
@@ -0,0 +1,7 @@
+---
+'@primer/brand-primitives': minor
+---
+
+Added a stylelint color allowlist export at `lib/design-tokens/styleLint/colors.json`.
```
rezrah: @stefankp I don't think stylelint is the right folder for this. It's too opinionated. Sibling folders have names like `css`, `js`, `ts`, etc... and given that this is just a JSON object for all color tokens, can we please change the output location to:

```suggestion
Added a new JSON-based color map export at /lib/design-tokens/json/colors.json, listing color tokens from @primer/brand-primitives.
```

☝️ this new output location would need to be wired in too, will leave separate comment about that.

.changeset/brand-stylelint-colors-export.md
```diff
@@ -0,0 +1,7 @@
...
+
+This one-dimensional JSON file lists every brand color custom property keyed by its CSS variable name (including the leading `--`), each annotated with `$type: "color"`. It gives stylelint plugins (and other tooling) an authoritative, type-aware allowlist of brand color tokens — mirroring the `@primer/primitives` `dist/styleLint/*.json` exports — so consumers can validate color values against brand tokens without relying on brittle name or path heuristics.
```
rezrah: Let's drop this from the public changeset. It's a bit too opinionated about stylelint.

```suggestion
```

packages/design-tokens/scripts/build-tokens.js
```diff
@@ -306,6 +309,33 @@ const darkJson = require('../src/tokens/base/colors/dark')
...
+   * (`--base-color-scale-*`); functional and component tokens keep the `brand` prefix.
+   */
+  buildPrimitives({
+    source: [`tokens/base/colors/color-scales.json`, `tokens/functional/colors/global.json`, ...filesForColorModes],
+    namespace,
+    platforms: {
+      styleLint: {
+        prefix: namespace,
+        addPrefix: token => token.isSource && !token.filePath.replace(/\\/g, '/').includes('/base/colors/'),
+        buildPath: `${outputPath}/styleLint/`,
```
rezrah: Per previous suggestion...

```suggestion
        buildPath: `${outputPath}/json/`,
```

packages/design-tokens/src/formats/stylelint-colors.js
```diff
@@ -0,0 +1,48 @@
...
+  const fp = (filePath || '').replace(/\\/g, '/')
+  return (
+    /\/base\/colors\//.test(fp) ||
+    /\/functional\/colors\//.test(fp) ||
+    /\/functional\/components\/[^/]+\/colors\.(js|json)$/.test(fp)
+  )
+}
+
+module.exports = {
+  name: 'json/stylelint-colors',
```
rezrah: Based on other comments can we call this somethign more generic, but still similar to primer/primitives...

```suggestion
  name: 'json/one-dimensional',
```

packages/design-tokens/src/formats/stylelint-colors.js
stefankp: Removed. The export is now a flat `--name: value` map with no per-entry metadata, so `filePath` is gone.

packages/design-tokens/src/formats/stylelint-colors.js
stefankp: Agreed, removed. Instead of annotating a type, the export now selects color tokens by value shape (a small `isColorValue` filter) and emits just the value — so non-color values that live in `colors.*` files (gradients, shadows, filters, `none`, `url()`) get excluded rather than mislabelled.

.changeset/brand-stylelint-colors-export.md
stefankp: Done — output is now `lib/design-tokens/json/colors.json`, and I've used your changeset wording. Build wiring updated to match.

.changeset/brand-stylelint-colors-export.md
stefankp: Dropped — the changeset is now a single generic line with no stylelint framing.

packages/design-tokens/scripts/build-tokens.js
stefankp: Done — `buildPath: ${outputPath}/json/`.

packages/design-tokens/src/formats/stylelint-colors.js
stefankp: Done — renamed to `json/one-dimensional`. It's now a generic flattener (no color/type logic — color selection happens in the build filter), so it can back future per-type exports too, matching the primer/primitives shape.

stefankp:
@rezrah I pushed an update addressing your points and updated the PR description, based on what we discussed yesterday.

This is an example output from JSON file:

```json
{
  "--base-color-scale-black-0": "#000000",
  "--base-color-scale-blue-0": "#DDF4FF",
  "--base-color-scale-blue-1": "#BCECFF",
  "--brand-color-accent-primary": "var(--base-color-scale-green-7)",
  "--brand-color-accent-secondary": "var(--base-color-scale-yellow-5)",
  "--brand-color-border-default": "var(--base-color-scale-gray-4)",
  "--brand-Accordion-border-color-default": "var(--brand-color-border-muted)",
  "--brand-Accordion-indicator-active": "var(--brand-color-accent-primary)",
  "--brand-Accordion-indicator-hover": "var(--base-color-scale-gray-5)"
}
```

.changeset/brand-color-map-export.md
```diff
@@ -0,0 +1,5 @@
+---
+'@primer/brand-primitives': minor
```
rezrah, line 2: As this affects more than one package...

```suggestion
'@primer/brand-primitives': minor
'@primer/brand-css': minor
'@primer/react-brand': minor
```

rezrah (approved):
🚀

PR #1378, opened 2026-06-17
Add NavList component

danielguillan:
## Summary

Towards https://github.com/github/brand-experience/issues/223

Adds a new `NavList` component for vertical navigation patterns such as docs/editorial sidebars, with support for flat link lists, expandable sections, nested sub-navigation, and optional visuals.

🔗 [Docs preview](https://primer-369fe91f37-26139705.drafts.github.io/brand/components/NavList/)
🔗 [Storybook preview](https://primer-369fe91f37-26139705.drafts.github.io/brand/storybook/?path=/story/components-navlist--playground)

## List of notable changes:

- **added** `NavList`, `NavList.Group`, `NavList.Item`, and `NavList.SubNav`
- **added** support for nested navigation up to five levels
- **added** expandable section rows with `defaultExpanded`, controlled `expanded`, and `onExpandedChange`
- **added** current item handling with `aria-current`, including auto-expanding parents with current descendants
- **added** `leadingVisual` and `trailingVisual` slots

- Review the `NavList.Item` + `NavList.SubNav` API and confirm it covers expected navigation use cases.
- Verify expandable section accessibility: section rows render as buttons, expose `aria-expanded`/`aria-controls`, and label their nested lists.
- Check current page behavior, especially auto-expanding parents with current descendants.
- Check visual behavior across flat lists, multiple nested levels, long labels, light/dark modes, leading/trailing visuals, hover/focus/current/disabled states, and collapsed sections.
- Review controlled expansion and custom link component behavior.

## Steps to test:

1. Open the Storybook preview and review `Components/NavList`.
2. Test the default, playground, sections, multiple levels, collapsed current descendant, long labels, visuals, and dark mode stories.
3. Open the docs preview and verify the examples, accessibility guidance, and prop tables.
4. Interact with expandable sections using mouse and keyboard, and confirm collapsed nested links are not reachable until expanded.

- Towards https://github.com/github/brand-experience/issues/223

## Contributor checklist:

## Reviewer checklist:

---

image


kellyquesnel:
@danielguillan this is looking great, huge shoutout for building this with so much attention to detail.

One small QA note: could we tuck the triangle icons a bit closer to the nested labels and shift them slightly left? Right now the hierarchy is working, but tightening that spacing would make it clearer that each subcategory/sub-subcategory is nested under its parent Level 1 category. Will attach a screenshot comparing the preview and design below!

Left: preview / Right: prototype

CleanShot 2026-06-18 at 11 40 53@2x

packages/react/src/NavList/NavList.tsx
```diff
@@ -0,0 +1,462 @@
...
+    return (
+      <ul
+        ref={setSubNavRef}
+        className={clsx(styles.NavList__subNav, className)}
+        aria-labelledby={ariaLabelledBy ?? labelledBy}
+        {...rest}
+        aria-hidden={expanded ? undefined : 'true'}
+        data-testid={testId || testIds.subNav}
+        data-expanded={expanded ? 'true' : 'false'}
+        inert={!expanded}
```
rezrah: FYI inert has typing issues in React 18, which is why we add it using an effect in other components. This might cause downstream compilation issues so can we please align to the existing way of setting the attribute for now.

packages/react/src/NavList/NavList.tsx
```diff
@@ -0,0 +1,462 @@
...
+      href,
+      id: accordionButtonIdProp,
+      ...accordionButtonProps
+    } = rest as React.ButtonHTMLAttributes<HTMLButtonElement> & {
+      href?: string
+    }
+    const controlledAccordionButtonId = accordionButtonIdProp ?? accordionButtonId
+
+    if (hasSubNav && !canExpand) {
+      throw new Error('NavList supports up to 5 levels. Level 5 items cannot contain NavList.SubNav.')
```
rezrah: Throwing an error feels pretty aggressive. We've always added a console.warn instead behind a NODE_ENV flag so it appears in dev only. For consistency, can we:

- Switch to a `console.warn`
- Silently error in the UI (not render)

packages/react/src/NavList/NavList.tsx
```diff
@@ -0,0 +1,462 @@
...
+  get toggle() {
+    return `${this.root}-toggle`
+  },
+}
+
+const DefaultNavigationLabel = 'Navigation'
+
+export type NavListRootProps = {
+  /**
+   * Accessible label for the navigation landmark. Defaults to "Navigation".
```
rezrah: Should aria-label be required Dani? I'm not sure if Navigation is sufficient from a desciptor perspective in all use cases as there might be secondary navigation on the page. Can (or should) we make aria-label required? I notice PRC also do this using a title prop, but I think I prefer this approach as it's a bit more explicit.

packages/react/src/NavList/NavList.tsx
```diff
@@ -0,0 +1,462 @@
...
+  if (!visual) return null
+
+  const visualElement = isValidElement(visual) ? visual : React.createElement(visual)
+
+  if (!isValidElement<{className?: string; ['aria-hidden']?: string; focusable?: string}>(visualElement)) return null
+
+  return (
+    <span className={className}>
+      {React.cloneElement(visualElement, {
+        className: clsx(visualElement.props.className),
```
rezrah:
```suggestion
        className: visualElement.props.className,
```

packages/react/src/NavList/NavList.module.css
```diff
@@ -0,0 +1,408 @@
...
+  background-color: var(--brand-color-canvas-subtle);
+}
+
+.NavList__item--leaf:not(.NavList__item--level-1) > .NavList__itemContent .NavList__link:hover {
+  background-color: transparent;
+  padding-inline-start: var(--base-size-8);
+}
+
+[data-color-mode='dark'] .NavList {
+  --brand-NavList-activeIndicator-color: var(--brand-color-accent-primary);
```
rezrah: @danielguillan what's the purpose of this? If you already have --brand-NavList-activeIndicator-color, should this not be assigned in the json already? Or if this is a local var, can we move it to design tokens folder please.

packages/react/src/NavList/NavList.tsx
```diff
@@ -0,0 +1,462 @@
...
+          >
+            {children}
+          </ul>
+        </NavListLevelContext.Provider>
+      </nav>
+    )
+  },
+)
+
+type Visual = ReactElement | React.ElementType
```
rezrah: As these are Octicons i think you'd need to add `Icon` from `@primer/octicons-react` to the union to prevent strict typing issues.

packages/react/src/NavList/NavList.tsx
```diff
@@ -0,0 +1,462 @@
...
+type NavListSubNavElement = ReactElement<NavListSubNavProps>
+
+type ElementWithChildren = ReactElement<{children?: ReactNode}>
+type ElementWithCurrent = ReactElement<{children?: ReactNode; 'aria-current'?: AriaAttributes['aria-current']}>
+
+function isNavListSubNavElement(node: ReactNode): node is NavListSubNavElement {
+  return isValidElement<NavListSubNavProps>(node) && node.type === NavListSubNav
+}
+
+function isCurrentValue(value: AriaAttributes['aria-current']) {
```
rezrah: Nit: extracted functions like this, in particular ones that are single-use should be inlined to the named compound component or parent function to help readability. In this case,  it'll leader to fewer lines of code to just inline it, so unsure what the value is in extracting it. Also it's the typical convention followed for other components and I'd love if we could stick to that 🙏. If we are legit using it in multiple places though, let's move to bottom of the file where it's not going to affect readability.

packages/react/src/NavList/NavList.features.stories.tsx
```diff
@@ -0,0 +1,222 @@
...
+      </NavList.Item>
+    </NavList>
+  )
+}
+
+export const Sections: Story = {
+  render: () => <ExpandedDocsNavigation />,
+}
+
+export const MultipleLevels: Story = {
```
rezrah: Love this story. @danielguillan could you please add some more stories that stress test this. I want to specifically see how stable rendering performance is. I'm seeing some potential issues but would love to have a live story to validate it. Could we add

- Navlist with 100 links through iteration
- Navlist with 5 sections with 20 links in each group and open them by default at the same time

packages/react/src/NavList/NavList.tsx
```diff
@@ -0,0 +1,462 @@
...
+      onClick,
+      onExpandedChange,
+      trailingVisual,
+      'aria-current': ariaCurrent,
+      'data-testid': testId,
+      ...rest
+    }: NavListItemProps<C>,
+    ref: Ref<HTMLElement>,
+  ) => {
+    const Component = as || 'a'
```
rezrah: Should this be type constrained to 'button' | 'a' instead of allowing a loose generic? It would make the typing a bit easier to reason about here if we did, unless there's a good reason to allow other element types?

packages/react/src/NavList/NavList.tsx
```diff
@@ -0,0 +1,462 @@
...
+    const childrenArray = Children.toArray(children)
+    const subNavChildren = childrenArray.filter(isNavListSubNavElement)
+    const subNav = childrenArray.find(isNavListSubNavElement)
+    const labelChildren = subNav ? childrenArray.filter(child => child !== subNav) : childrenArray
+    const hasLabelContent = Children.toArray(labelChildren).length > 0
+    const hasSubNav = Boolean(subNav)
+    const canExpand = level < MaxNavListLevel
+    const isLeafItem = !hasSubNav
+    const controlledSubNavId = subNav?.props.id ?? subNavId
+    const hasCurrentSubNavItem = hasCurrentDescendant(subNav?.props.children)
```
rezrah: I think we should memoize values like as it will run this type of logic on each re-render otherwise.

Something like this:

```suggestion
    const hasCurrentSubNavItem = useMemo(() => hasCurrentDescendant(subNav?.props.children), [subNav?.props.children])
```

Some others here might warrant it too. Think we should generally do a performance pass to make sure we've mitigated low-hanging performance issues.

packages/react/src/NavList/NavList.module.css
```diff
@@ -0,0 +1,408 @@
...
+}
+
+.NavList__item:not(.NavList__item--level-1):has(
+    > .NavList__itemContent .NavList__accordionButton[aria-expanded='false']:not(:disabled)
+  ):has(> .NavList__subNav [aria-current]:not([aria-current='false']))
+  > .NavList__itemContent
+  .NavList__accordionButton:not(:disabled)
+  .NavList__labelArea {
+  background-color: var(--brand-color-canvas-subtle);
+}
```
rezrah, line 135: @danielguillan I'm finding selectors like this pretty hard to read. Same for you, or just me? 

This file contains a lot of deeply nested `:has` usage, but most seem to already be pre-computed in the JS. I totally appreciate you're using higher specificity for `:has` but in React i can see you're already computing things like `hasCurrentSubNavItem` so can we potentially shorten the selectors and deduplicate any work where possible? The intent being to not do the same work twice.

packages/react/src/NavList/NavList.tsx
```diff
@@ -0,0 +1,462 @@
...
+        assignRef(ref, node)
+      },
+      [ref],
+    )
+
+    React.useEffect(() => {
+      setSubNavAndAncestorHeights()
+    }, [children, expanded, setSubNavAndAncestorHeights])
+
+    React.useEffect(() => {
```
rezrah: Nit... @danielguillan we seem to have a mix of destructured hooks + namespaced ones in this feel, could we clean those up please 🙏

E.g. here you're using the React namespace but you've already destructured `useEffect` at the top of the file. There's also others that we should consider destructuring for consistency like `React.useCallback`

packages/react/src/NavList/NavList.features.stories.tsx
```diff
@@ -0,0 +1,222 @@
...
+  ))
+
+const ExpandedDocsNavigation = ({colorMode}: {colorMode?: 'light' | 'dark'}) => {
+  const nav = (
+    <NavList aria-label="Docs navigation">
+      <NavList.Item defaultExpanded>
+        GitHub Copilot
+        <NavList.SubNav>
+          {renderArticleItems([
+            'Overview',
```
rezrah: @danielguillan could you backfill the l10n translations for these stories please? You can point Copilot at Accordion if it helps. Would love to see how this behaves across different localizations.

packages/react/src/NavList/NavList.features.stories.tsx
```diff
@@ -0,0 +1,222 @@
...
+}
+
+const renderArticleItems = (labels: string[], currentLabel?: string) =>
+  labels.map(label => (
+    <NavList.Item key={label} href="#" aria-current={label === currentLabel ? 'page' : undefined}>
+      {label}
+    </NavList.Item>
+  ))
+
+const ExpandedDocsNavigation = ({colorMode}: {colorMode?: 'light' | 'dark'}) => {
```
rezrah: Sorry stupid question probably but why are we allowing a prop here for color mode? Isn't this already handled in the global storybook decorator? If you need a local override it should be a storywide decorator too.

rezrah: Also could we refactor this file so that all the components are inlined to the CSF3 story declarion please? Makes the code output in storybook ui cleaner and file easier to read E.g. 1 story renders 1 JSX. consts for data sharing is legit though.

packages/react/src/NavList/NavList.module.css
```diff
@@ -0,0 +1,408 @@
+.NavList {
+  --brand-NavList-activeIndicator-color: var(--brand-color-text-emphasized);
```
rezrah: Could we move this out to Brand Primitives please?

packages/react/src/NavList/NavList.module.css
```diff
@@ -0,0 +1,408 @@
...
+
+  font-family: var(--brand-fontStack-sansSerif);
```
rezrah, line 3: Do we need this? Isn't it already in our main stylesheet? 🤔

packages/react/src/SubNav/SubNav.visual.spec.ts-snapshots/Visual-Comparison-SubNav-Mobile-viewport-test-d89c8-av-Variant-SubNav-Narrow-Anchor-Nav-Variant-1-darwin.png
rezrah, line 1: @danielguillan this looks like a flakey snapshot. okay to reset this?

packages/react/src/IDE/IDE.visual.spec.ts-snapshots/Visual-Comparison-IDE-IDE-Playground-1-darwin.png
rezrah, line 1: FYI Another stray snapshot 👇. Could we reset this please. Also reminds me we need to kill the IDE 🔥

apps/next-docs/content/components/NavList/index.mdx
```diff
@@ -0,0 +1,242 @@
...
+      <NavList.Item href="#">Secret scanning</NavList.Item>
+      <NavList.Item href="#">Code scanning</NavList.Item>
+    </NavList.SubNav>
+  </NavList.Item>
+</NavList>
+```
+
+### Five levels
+
+NavList supports up to five levels. Levels one through four can optionally expand or collapse; level five is reserved for leaf article links.
```
rezrah: Nit.. i recall the convention was on react pages we use component pascal case in code tag, but on guidelines we use natural name

```suggestion
`NavList` supports up to five levels. Levels one through four can optionally expand or collapse; level five is reserved for leaf article links.
```

apps/next-docs/content/components/NavList/index.mdx
```diff
@@ -0,0 +1,242 @@
...
+    <NavList.SubNav>
+      <NavList.Item href="#">Overview</NavList.Item>
+      <NavList.Item href="#">Secret scanning</NavList.Item>
+      <NavList.Item href="#">Code scanning</NavList.Item>
+    </NavList.SubNav>
+  </NavList.Item>
+</NavList>
+```
+
+### Five levels
```
rezrah: Should we say Up to five levels so it's more about the upper-bound? Presumably we don't want to encourage using five.

```suggestion
### Maximum limit of five nested sub navigations
```

apps/next-docs/content/components/NavList/index.mdx
```diff
@@ -0,0 +1,242 @@
...
+  </NavList.Item>
+  <NavList.Item href="#">Quickstart</NavList.Item>
+  <NavList.Item href="#">Install GitHub Copilot</NavList.Item>
+  <NavList.Item href="#">Configure your editor</NavList.Item>
+  <NavList.Item href="#">Manage policies</NavList.Item>
+  <NavList.Item href="#">Troubleshooting</NavList.Item>
+</NavList>
+```
+
+### Sections
```
rezrah: Mostly curious @danielguillan. PRC seems to use the 'group' term to describe sections. They even have `.Group` as part of the API. What was the reasoning for us not doing that too?

packages/react/src/NavList/NavList.tsx
danielguillan: Thanks! It's fixed now.

packages/react/src/NavList/NavList.tsx
danielguillan: Agree! Updated to a warning in dev + not render

packages/react/src/NavList/NavList.features.stories.tsx
danielguillan: Done! Also did some perf validation on these stories and results look healthy. 👍

packages/react/src/NavList/NavList.tsx
danielguillan: Yup! It's updated now, I've memoized all repeated children filtering and lookups.

packages/react/src/NavList/NavList.tsx
danielguillan: Agree, I've made it required.

packages/react/src/NavList/NavList.module.css
danielguillan: Our main stylesheet does not define a default `font-family`, and NavList does not internally use any components that set it, so it's required here.

apps/next-docs/content/components/NavList/index.mdx
```diff
@@ -0,0 +1,255 @@
...
+      </NavList.Item>
+      <NavList.Item href="#" leadingVisual={WorkflowIcon}>
+        GitHub Actions
+      </NavList.Item>
+      <NavList.Item href="#" leadingVisual={CodeIcon}>
+        Codespaces
+      </NavList.Item>
+    </NavList.SubNav>
+  </NavList.Item>
+  <NavList.Item defaultExpanded trailingVisual={<KebabHorizontalIcon />}>
```
rezrah: The second item for me wasn't immediately clear what it was referring to. Would be nice to just have a single item that shows leading _and_ trailing visuals to keep this shorter? If this is intended for stylistic reasons though, ignore me (e.g. you shouldn't put leading + trailing on the same item, but if that's true, that shouldn't be possible in the code logic itself)

packages/react/src/NavList/NavList.stories.tsx
```diff
@@ -0,0 +1,117 @@
...
+      control: 'text',
+      description: 'Accessible label for the navigation landmark.',
+    },
+  },
+} satisfies Meta<typeof NavList>
+
+export default meta
+type Story = StoryObj<typeof NavList>
+
+const ArticleLabelKeys = [
```
rezrah: Nit... could we inline this to the default stories closure please. Only used there.

packages/react/src/NavList/NavList.test.tsx
```diff
@@ -0,0 +1,936 @@
...
+        </NavList.Item>
+      </NavList>,
+    )
+
+    const link = getByRole('link', {name: 'Getting started'})
+    const toggle = getByRole('button', {name: 'Guides'})
+
+    expect(link.querySelector('.NavList__leadingVisual')).toBeInTheDocument()
+    expect(link.querySelector('.NavList__trailingVisual')).toBeInTheDocument()
+    expect(toggle.querySelector('.NavList__leadingVisual')).toBeInTheDocument()
```
rezrah: As you already use the testIds convention in NavList @danielguillan, it'd be worth switching to that here so we can reduce burden on the css class here.

packages/react/src/NavList/NavList.tsx
```diff
@@ -0,0 +1,627 @@
...
+        </NavListLevelContext.Provider>
+      </nav>
+    )
+  },
+)
+
+type Visual = ReactElement | React.ElementType | Icon
+type NavListItemAs = 'a' | 'button'
+
+type ItemOwnProps = {
```
rezrah: Could we call this NavListItemBaseProps please @danielguillan? Feels more consistent with surrounding type names and other components. Sorry i missed this earlier.

packages/react/src/NavList/NavList.tsx
```diff
@@ -0,0 +1,627 @@
...
+  if (!isValidElement(node)) return false
+
+  if (node.type === React.Fragment) {
+    return Children.toArray((node as ElementWithChildren).props.children).some(childHasNavListGroupElement)
+  }
+
+  return isNavListGroupElement(node)
+}
+
+function getValidNavListGroupChildren(children: ReactNode): ReactNode[] {
```
rezrah: @danielguillan , sorry to be a pedant about this. This is only relevant to NavList.Group right? Could it be inlined to the component closure instead as it's directly manipulating children? I had to scroll to bottom of this file to read what this is doing, but not sure if that's just overhead or needed for perf reasons. LMK if i'm missing somethign though.

packages/react/src/NavList/NavList.tsx
```diff
@@ -0,0 +1,627 @@
...
+      {React.cloneElement(visualElement, {
+        className: visualElement.props.className,
+        'aria-hidden': 'true',
+        focusable: 'false',
+      })}
+    </span>
+  )
+}
+
+function getInvalidNavListItemMessage({
```
rezrah: Same with this one @danielguillan. I remember it being inline before, and now it's been hoisted out.

.changeset/navlist-component.md
```diff
@@ -0,0 +1,5 @@
+---
+'@primer/react-brand': minor
+---
+
+Adds the NavList component for vertical navigation with groups, nested expandable links, current item states, and custom link support.
```
rezrah:
```suggestion
Adds the `NavList` component for vertical navigation with groups, nested expandable links, current item states, and custom link support.

[See docs for usage examples](https://primer.style/brand/components/NavList)
```

rezrah (approved):
Really lovely work on this @danielguillan 👏 👏 ✨ 

Thanks for adding the extra stories and examples. Left a few more minor comments to look at but otherwise it's a  ✔️ from me to unblock you. Really looking forward to using this.

Deferring to @jesussandreas for a final design pass 🙏. We want to mitigate post-merge design QA issues from being reported on this, so can we please do a sweep now 🙇

jesussandreas (approved):
looks great dudee

PR #1394, opened 2026-07-08
Add JSON typography and size map exports to brand-primitives

stefankp:
## Summary

Adds two per-type JSON token maps to `@primer/brand-primitives`, mirroring the color map export merged in #1370: `lib/design-tokens/json/typography.json` and `lib/design-tokens/json/size.json`. Each is a one-dimensional map keyed by CSS custom-property name (`--brand-*` / `--base-*`) → value, reusing the existing `json/one-dimensional` Style Dictionary format.

These back downstream tooling (e.g. stylelint rules) that validates `var(--…)` references against the set of custom properties brand actually ships. It's a **name-based allowlist**, so each export is scoped to exactly match brand's shipped custom properties for that category. Colors were done in #1370; typography and size are next.

**Example — `typography.json`:**

```json
{
  "--base-text-weight-bold": 700,
  "--brand-body-fontFamily": "'Mona Sans', 'MonaSansFallback',-apple-system, BlinkMacSystemFont, 'Segoe UI', Helvetica, Arial, sans-serif, 'Apple Color Emoji', 'Segoe UI Emoji'",
  "--brand-fontStack-sansSerif": "'Mona Sans', 'MonaSansFallback',-apple-system, BlinkMacSystemFont, 'Segoe UI', Helvetica, Arial, sans-serif, 'Apple Color Emoji', 'Segoe UI Emoji'",
  "--brand-heading-lineHeight-400": "1.3",
  "--brand-text-size-200": "1rem",
  "--brand-text-size-800": "2rem",
  "--brand-text-weight-600": "var(--base-text-weight-normal)"
}
```

**Example — `size.json`:**

```json
{
  "--base-size-8": "0.5rem",
  "--base-size-16": "1rem",
  "--brand-borderRadius-medium": "0.5rem",
  "--brand-borderWidth-thin": "max(1px, 0.0625rem)",
  "--brand-box-spacing-normal": "1.5rem",
  "--brand-breakpoint-medium": "48rem",
  "--brand-control-medium-gap": "0.75rem"
}
```

## List of notable changes:

- **added** `lib/design-tokens/json/typography.json` (109 tokens) — base typography + functional responsive typography
- **added** `lib/design-tokens/json/size.json` (76 tokens) — base size + functional size, breakpoints and border widths/radii
- **added** `src/filters/isSizeValue.js` (+ unit tests) — value-shape predicate that restricts the size export to length-valued tokens
- **added** two `buildPrimitives` steps in `scripts/build-tokens.js` for the typography and size exports

**Scope decisions (intentional exclusions):**
- `functional/size/viewport` is excluded — it emits only `@custom-media` breakpoint definitions (0 custom properties), never referenced as `var()`.
- Animation values that live in the size sources (`--brand-control-animation-duration`, `--brand-control-animation-easing`) are dropped by `isSizeValue` — they belong to a future animation rule.
- `--brand-borderInset-*` `inset 0 0 0 …` composites are dropped by `isSizeValue` — they'll belong to a future border rule.
- The `-auto` pointer variants from `size-fine`/`size-coarse` are excluded — those files aren't imported by brand and the tokens are referenced 0× in brand CSS, so they aren't shipped custom properties.
- Upstream `functional/typography/typography.json` is excluded — `packages/react/src/css/stylesheets.ts` doesn't import the CSS built from it, so its 43 tokens (the `--brand-text-{body,title,caption,subtitle,display,code*}-*` styles plus `--brand-fontStack-sansSerifDisplay`) never become shipped custom properties. Same criterion as the `size-fine`/`size-coarse` exclusion above. It's also what keeps `--brand-fontStack-*` on the Mona Sans values shown in the sample above — that file was the only thing redefining them to the generic system stack.

Note: values in `size.json` are resolved literals (e.g. `--brand-borderRadius-medium: "0.5rem"`), whereas typography and colors keep `var()` references where present — but the key/name set, which is what the allowlist consumes, is complete and correct.

- The token name-set in each JSON file matches brand's shipped `--brand-*`/`--base-*` custom properties for that category.
- `isSizeValue` correctly keeps lengths / math functions / size `var()` refs and drops time, easing, `inset …` composites, and media-query ranges.

## Steps to test:

1. Run `npm run build -w @primer/brand-primitives`
1. Confirm `lib/design-tokens/json/typography.json` (109 entries) and `size.json` (76 entries) are generated
1. Run `npm test -w @primer/brand-primitives` — 29 tests pass (incl. the new `isSizeValue` suite)

- Follows #1370 (color map export)

## Contributor checklist:

- [ ] UI Changes contain new visual snapshots (generated by adding `update snapshots` label to the PR) — N/A, no UI changes

### Out of scope: pre-existing issue in `colors.json`

While cross-checking the typography map against brand's built CSS, I found one entry in the existing `colors.json` export that isn't a shipped custom property: `--brand-blinking-cursor-fgColor`. It's defined only in the vendored `lib/design-tokens/css/tokens/functional/components/blinking-cursor/colors-with-modes.css`, and nothing under `packages/react/src` imports that file — so the token is absent from `main.css` and from every per-component esm chunk.

This came in with the color map export in #1370 and predates this PR. I've deliberately left it alone to keep this diff focused; flagging it here for reviewers and for a follow-up. The other 516 entries in `colors.json` check out.


stefankp:
Worth calling out a change to the typography export, since it changes the shape of the output.

The map was originally built from three sources, including `tokens/functional/typography/typography.json`. That one is now excluded. `packages/react/src/css/stylesheets.ts` imports only `base/typography/typography.css` and `functional/typography/typography-responsive.css` — nothing imports the CSS built from the third file, so the 43 tokens it defines (`--brand-text-{body,title,caption,subtitle,display,code*}-*` plus `--brand-fontStack-sansSerifDisplay`) are never defined in the browser.

That matters more than the count. This export is consumed as a name-based allowlist, so including a token that's never defined green-lights `var()` references that fail at runtime. A reference to an undefined custom property makes the whole declaration invalid at computed-value time: it still wins the cascade and then computes to `inherit`, rather than falling through to a lower-precedence declaration. So `font-size: var(--brand-text-caption-size)` silently renders at the inherited size, and a `@media` override built on a dead token silently erases a valid base value.

The export goes from **152 → 109 tokens**. I cross-checked the emitted key list against the typography primitives actually defined in the built `main.css` bundle: exact 1:1 match, nothing missing in either direction, and none of the 109 surviving values changed.

Two things this also cleans up:

- The source-ordering workaround is gone. The `--brand-fontStack-*` collisions only existed because the legacy file redefined `sansSerif` / `monospace` / `system` with the old Apple system stack. With it removed, the two remaining sources have disjoint namespaces and order is irrelevant.
- `--brand-fontStack-sansSerifDisplay` existed *only* in the legacy file, so nothing overrode it and it leaked the generic stack. It was visible in this PR's original sample output (`"--brand-text-title-shorthand-large": "600 2rem/1.5 -apple-system, …"`). It's now correctly absent.

The description has been updated to match.

packages/design-tokens/scripts/build-tokens.js
```diff
@@ -339,6 +340,80 @@ const darkJson = require('../src/tokens/base/colors/dark')
...
+   * typography.json`. All values here are legitimate typography tokens (font family, weight, size,
+   * line height, letter spacing, font-feature settings), so no value-shape filter is needed. The
+   * sources are ordered so `typography-responsive.json` loads last and wins the handful of
+   * `--brand-fontStack-*` name collisions with the brand (Mona Sans) values. Base tokens are
+   * emitted without the `brand` prefix to match their CSS output (`--base-text-*`); functional
+   * tokens keep the `brand` prefix.
+   */
+  buildPrimitives({
+    source: [
+      `tokens/functional/typography/typography.json`,
```
rezrah: @stefankp this currently doesn't include any component-level typography tokens. Is that intentional?

E.g. `--brand-Testimonial-quote-fontSize-default` won't be included here.


The comment above it states: 

```
This is a name-based allowlist, so it must cover every `--brand-*` / `--base-*` typography custom property brand ships
```

So we'd need another way to grab those here as wildcard won't work. But then maybe the comment is wrong? Can you clarify whether you meant to include only global tokens and if so, update the comment.

packages/design-tokens/scripts/build-tokens.js
```diff
@@ -339,6 +340,80 @@ const darkJson = require('../src/tokens/base/colors/dark')
...
+   * `--base-*` typography custom property brand ships — including the core text-style tokens
+   * (`--brand-text-body-*`, `--brand-text-title-*`, `--brand-text-caption/subtitle/display/code*`
+   * and `--brand-fontStack-sansSerifDisplay`) that live only in `functional/typography/
+   * typography.json`. All values here are legitimate typography tokens (font family, weight, size,
+   * line height, letter spacing, font-feature settings), so no value-shape filter is needed. The
+   * sources are ordered so `typography-responsive.json` loads last and wins the handful of
+   * `--brand-fontStack-*` name collisions with the brand (Mona Sans) values. Base tokens are
+   * emitted without the `brand` prefix to match their CSS output (`--base-text-*`); functional
+   * tokens keep the `brand` prefix.
+   */
```
rezrah, line 345: This is super verbose, and i think it's overexplaining implementation detail.

```suggestion
  /**
   * Builds a flat JSON map from global typographic tokens.
   */
```

packages/design-tokens/scripts/build-tokens.js
rezrah: I don't think we need `tokens/functional/typography/typography.json` here @stefankp.

`tokens/functional/typography/typography-responsive.json` already covers it.

```suggestion

```

rezrah: Removing this should also fix this bit in your PR description. The file shouldn't be included here to begin with, and is causing product tokens to leak.

Image

packages/design-tokens/scripts/build-tokens.js
```diff
@@ -339,6 +340,80 @@ const darkJson = require('../src/tokens/base/colors/dark')
...
+   * functional size, border and breakpoint tokens) for downstream tooling. The source files are
+   * the same ones brand ships in its CSS (`base/size/size.css`, `functional/size/size.css`,
+   * `functional/size/breakpoints.css`, `functional/size/border.css`); the pointer-specific
+   * `size-fine` / `size-coarse` variants and the `viewport` media-query ranges are excluded
+   * because brand does not import them. Size tokens are additionally selected by inspecting the
+   * emitted value (`isSizeValue`), so non-size values that live alongside sizes (animation
+   * easing/duration, `inset ...` border composites) are dropped. Base tokens are emitted without
+   * the `brand` prefix to match their CSS output (`--base-size-*`); functional tokens keep the
+   * `brand` prefix.
+   */
```
rezrah, line 368: Similar to other comment, I'm struggling to read this (Is it just me? 😅). The backticks aren't helping here, and are superfluous.

Should be much shorter.


```suggestion
  /** Build a flat JSON map of our global size tokens for downstream tooling. */
```

packages/design-tokens/src/filters/isSizeValue.js
```diff
@@ -0,0 +1,39 @@
...
+ *   `none`, and `var(...)` references into the animation / motion domain.
+ *
+ * @param {string} value the token's resolved value
+ * @returns {boolean} `true` when the value is size-shaped
+ */
+const SIZE_FUNCTION_PREFIXES = ['max(', 'min(', 'clamp(', 'calc(']
+const LENGTH_VALUE = /^-?(\d+\.?\d*|\.\d+)(px|rem|em|vh|vw|vmin|vmax|ch)$/
+const NON_SIZE_VAR = /(animation|easing|duration|transition|motion|cubic)/
+
+const isSizeValue = value => {
```
rezrah, line 11: Just a note to say that we should pull isDimension from `@primer/primitives` in future. We can't do it right now, so this is fine as a patch. 

Want to add a TODO here @stefankp?

packages/design-tokens/src/filters/isSizeValue.js
```diff
@@ -0,0 +1,39 @@
...
+ *   / `calc()`), or a `var(...)` reference to another size token. It excludes time values
+ *   (`80ms`, `0.6s`), easing (`cubic-bezier(...)`), `inset ...` composites, media-query ranges,
+ *   `none`, and `var(...)` references into the animation / motion domain.
+ *
+ * @param {string} value the token's resolved value
+ * @returns {boolean} `true` when the value is size-shaped
+ */
+const SIZE_FUNCTION_PREFIXES = ['max(', 'min(', 'clamp(', 'calc(']
+const LENGTH_VALUE = /^-?(\d+\.?\d*|\.\d+)(px|rem|em|vh|vw|vmin|vmax|ch)$/
+const NON_SIZE_VAR = /(animation|easing|duration|transition|motion|cubic)/
```
rezrah: This isn't catching all non size vars @stefankp.

E.g.

Image

Long term we need to clean up the token metadata (use $type) and call `isDimension` from @primer/primitives.

In the meantime, do you want to double check if the output is accurate and not outputting noise?

packages/design-tokens/src/filters/isSizeValue.test.js
```diff
@@ -0,0 +1,65 @@
...
+    expect(isSizeValue('clamp(1rem, 2vw, 3rem)')).toBe(true)
+    expect(isSizeValue('calc(48rem - 0.02px)')).toBe(true)
+  })
+
+  it('keeps var() references to size tokens', () => {
+    expect(isSizeValue('var(--base-size-16)')).toBe(true)
+    expect(isSizeValue('var(--brand-borderWidth-thin)')).toBe(true)
+  })
+
+  it('excludes var() references into the animation domain', () => {
```
rezrah: It's worth adding another test for color values. Currently returning true, but shouldn't be.

Related: https://github.com/primer/brand/pull/1394/changes#r3704718207

packages/design-tokens/src/filters/isSizeValue.js
```diff
@@ -0,0 +1,39 @@
...
+ *   restrict a one-dimensional export to spacing/size tokens. Brand size sources mix value kinds:
+ *   the size files hold a couple of animation tokens (`--brand-control-animation-easing`,
+ *   `--brand-control-animation-duration`), the border file holds `inset ...` shadow composites,
+ *   and the viewport file holds media-query ranges (`(max-width: ...)`). Rather than trusting file
+ *   paths, this inspects the emitted value: a value is treated as a size when it is a bare length
+ *   (`px` / `rem` / `em` / viewport / `ch` units), a math function (`max()` / `min()` / `clamp()`
+ *   / `calc()`), or a `var(...)` reference to another size token. It excludes time values
+ *   (`80ms`, `0.6s`), easing (`cubic-bezier(...)`), `inset ...` composites, media-query ranges,
+ *   `none`, and `var(...)` references into the animation / motion domain.
+ *
```
rezrah:
```suggestion
 * @description Check whether a token value can be used as a CSS size. 
 *
```

.changeset/brand-functional-token-maps-export.md
```diff
@@ -0,0 +1,7 @@
+---
+'@primer/brand-primitives': minor
+'@primer/brand-css': minor
+'@primer/react-brand': minor
```
rezrah: Should only apply to one package:

```suggestion
'@primer/brand-primitives': patch
```

rezrah: Technically it also applies to `@primer/react-brand` too as we ship tokens there, but it'll need a new changeset with different NPM paths.

packages/design-tokens/scripts/build-tokens.js
rezrah, line 368:
```suggestion
  /** 
   * Build a flat JSON map of global size tokens. 
   */
```

packages/design-tokens/scripts/build-tokens.js
stefankp: Yes — global tokens only is intentional; component-scoped tokens remain internal to their component chunks. I shortened the comment to make that scope explicit.

packages/design-tokens/scripts/build-tokens.js
stefankp: Updated to the shorter global-typography wording.

packages/design-tokens/scripts/build-tokens.js
stefankp: Removed. The export now uses only the base and responsive typography sources, matching the global stylesheet.

packages/design-tokens/scripts/build-tokens.js
stefankp: Shortened this to the global-size wording.

packages/design-tokens/src/filters/isSizeValue.js
stefankp: Added a TODO to move to `isDimension` once all size tokens carry `$type` metadata.

packages/design-tokens/src/filters/isSizeValue.js
stefankp: Confirmed the generated map contains no noise: 76 entries, with only the two animation values and three border-inset composites excluded from the 81 source declarations. I also changed the predicate to reject every unresolved `var()` value, so color, font, and other non-size references cannot pass.

packages/design-tokens/src/filters/isSizeValue.test.js
stefankp: Added color coverage, including `var(--brand-color-text-default)`.

packages/design-tokens/src/filters/isSizeValue.js
stefankp: Shortened the description to this wording.

.changeset/brand-functional-token-maps-export.md
stefankp: Split this into patch changesets for `@primer/brand-primitives` and `@primer/react-brand`, and removed `@primer/brand-css`.

packages/design-tokens/scripts/build-tokens.js
stefankp: Updated to this shorter wording.

stefankp: AI generated comments are over the top 😄

rezrah (approved):
🚀 

Thanks for addressing the feedback @stefankp. LGTM.

PR #1406, opened 2026-07-15
Update CTA banner thumbnail

natalierharris:
## Summary
This is to update the CTA banner thumbnails for primer.style/brand

## List of notable changes:
Replaced old image assets with new ones with the same file path and name

-


natalierharris:
@danielguillan @rezrah This is also another thumbnail update, and should be good to go

rezrah (approved):

PR #1417, opened 2026-07-22
Fix missing form-control color tokens in the ESM build

stefankp:
## Summary

The `@primer/react-brand/esm` build never shipped the `--brand-control-*` **color** token family, so pure-ESM consumers that drop the global `lib/css/main.css` lose every form-control border and box: `TextInput` / `Select` / `Textarea` borders disappear and `Checkbox` / `Radio` render invisible (this is what broke the ContactSales form on github.com/enterprise/contact).

The control color tokens reached CSS only through a single side-effect import in the `forms/index.ts` re-export barrel. The ESM build (Vite + rolldown, `preserveModules`) rewrites the package entry to import the leaf component modules directly and never emits `esm/forms/index.js`, so that barrel's CSS side-effect — and the tokens with it — silently vanished from the ESM output. Every other component (Button, Card, Label, …) already co-locates its own `colors-with-modes.css` import inside the component, so its tokens ride along with the component module; the forms family was the lone exception relying on the barrel.

This moves the control-token import into each of the six form components to match that established per-component pattern, and drops the now-redundant barrel import. `lib/css/main.css` (the monolithic CSS-class build) is unaffected.

Both released `0.70.0` and `0.71.0` ship the ESM build without these tokens — no released version has ever shipped working ESM control tokens — so consumers on the ESM entry will need to upgrade to the release containing this fix.

## List of notable changes:

- **added** the `functional/components/control/colors-with-modes.css` import to `TextInput`, `Textarea`, `Select`, `Checkbox`, `Radio` and `FormControl`, so the `--brand-control-*` color tokens ship with the components in the ESM build (mirrors the existing `Button` pattern)
- **removed** the shared control-token import from `forms/index.ts`, since the ESM build drops that re-export barrel and it therefore never delivered the tokens to ESM consumers

- That the co-located import matches the existing per-component convention (e.g. `Button.tsx`), rather than the barrel-level import it replaces.
- That `lib/css/main.css` is unchanged in content — the fix only changes what the ESM build emits (the CSS-class build was already self-contained).
- That the control **sizing/animation** tokens (`--brand-control-large-*`, `--brand-control-medium-*`, `--brand-control-animation-*`), which already shipped in ESM, are not regressed.

## Steps to test:

1. Build the library: `npm run build:lib`.
2. Inspect `packages/react/esm/`: each of the six form component `.js` files now side-effect-imports a `functional/components/control/colors-with-modes-*.css`, and that token CSS is emitted into `esm/` (previously it was absent entirely).
3. Run the reference-vs-definition check over the built `esm/**/*.css`: **0** `--brand-control-*` tokens are now used without a definition (was **32**).
4. Or reproduce as a consumer: import `FormControl` / `TextInput` / `Select` / `Checkbox` / `Radio` from `@primer/react-brand/esm` in a bundler **without** importing `main.css`. Controls now render with borders/boxes; before this fix the borders collapsed and checkboxes/radios were invisible.

<details>
<summary>Reference-vs-definition check (run from <code>packages/react</code>, prints 0 after the fix)</summary>

```js
node -e "const fs=require('fs'),p=require('path'),f=[];(function w(d){for(const e of fs.readdirSync(d,{withFileTypes:1})){const q=p.join(d,e.name);e.isDirectory()?w(q):e.name.endsWith('.css')&&f.push(q)}})('esm');const D=/--(?:brand|base)-[\w-]+\s*:/g,R=/var\(\s*(--(?:brand|base)-[\w-]+)\s*([,)])/g,defs=new Set(),refs=new Map();for(const x of f){const s=fs.readFileSync(x,'utf8');let m;while(m=D.exec(s))defs.add(m[0].replace(/\s*:$/,''));while(m=R.exec(s)){const t=m[1];if(!refs.has(t))refs.set(t,0);if(m[2]===')')refs.set(t,1)}}console.log([...refs].filter(([t,n])=>n&&!defs.has(t)&&t.includes('control')).length)"
```

</details>

- Affects released `@primer/react-brand@0.70.0` and `@0.71.0` — both ship the ESM build without the `--brand-control-*` color tokens. Consumers using the ESM entry should upgrade once this ships.

## Contributor checklist:

- [x] All new and existing CI checks pass (`tsc`, ESLint with 0 warnings, Prettier, `jest src/forms` — 82 tests)
- [ ] Tests prove that the feature works and covers both happy and unhappy paths — N/A: this is a build/packaging fix with no unit-test surface; validated against the built `esm/` output (reference check + an ESM-only consumer harness)
- [ ] UI Changes contain new visual snapshots — N/A: no component markup/style changes; `main.css` is unchanged
- [ ] Related issues have been referenced in the PR description — N/A

## Reviewer checklist:

## Screenshots:

> Same ContactSales-style form rendered from `@primer/react-brand/esm` with **no** `main.css`, production-bundled.

<table>
<tr>
</tr>
<tr>
<td valign="top">

Before

 </td>
<td valign="top">

After

</td>
</tr>
</table>

### DIFF
--- a/.changeset/fix-esm-control-color-tokens.md
+++ b/.changeset/fix-esm-control-color-tokens.md
@@ -1,0 +1,7 @@
+---
+'@primer/react-brand': patch
+---
+
+Fixed missing form-control color tokens in the ESM build. `TextInput`, `Textarea`, `Select`, `Checkbox`, `Radio`, and `FormControl` now bundle the `--brand-control-*` color tokens directly, so their borders and boxes render correctly for consumers that import from `@primer/react-brand/esm` without also loading the global `@primer/react-brand/lib/css/main.css` stylesheet.
+
+Previously these tokens only reached CSS through a shared side-effect import in the forms barrel (`forms/index.ts`), which the ESM build drops. On ESM-only pages this left `var(--brand-control-*)` undefined, collapsing control borders (and rendering checkboxes and radios invisible). This affected the `0.70.0` and `0.71.0` releases; consumers using `@primer/react-brand/esm` should upgrade to this release.
--- a/.changeset/fix-esm-control-color-tokens.md
+++ b/.changeset/fix-esm-control-color-tokens.md
@@ -2,6 +2,4 @@
 '@primer/react-brand': patch
 ---
 
-Fixed missing form-control color tokens in the ESM build. `TextInput`, `Textarea`, `Select`, `Checkbox`, `Radio`, and `FormControl` now bundle the `--brand-control-*` color tokens directly, so their borders and boxes render correctly for consumers that import from `@primer/react-brand/esm` without also loading the global `@primer/react-brand/lib/css/main.css` stylesheet.
-
-Previously these tokens only reached CSS through a shared side-effect import in the forms barrel (`forms/index.ts`), which the ESM build drops. On ESM-only pages this left `var(--brand-control-*)` undefined, collapsing control borders (and rendering checkboxes and radios invisible). This affected the `0.70.0` and `0.71.0` releases; consumers using `@primer/react-brand/esm` should upgrade to this release.
+Fixed missing styles for ESM form controls, where the dependent design tokens were missing . Affects `TextInput`, `Textarea`, `Select`, `Checkbox`, `Radio`, and `FormControl`.
--- a/packages/react/src/forms/Checkbox/Checkbox.tsx
+++ b/packages/react/src/forms/Checkbox/Checkbox.tsx
@@ -6,6 +6,7 @@ import type {BaseProps} from '../../component-helpers'
 import type {FormValidationStatus} from '../form-types'
 import useLayoutEffect from '../../hooks/useIsomorphicLayoutEffect'
 
+import '@primer/brand-primitives/lib/design-tokens/css/tokens/functional/components/control/colors-with-modes.css'
 import styles from './Checkbox.module.css'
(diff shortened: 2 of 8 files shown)


.changeset/fix-esm-control-color-tokens.md
```diff
@@ -0,0 +1,7 @@
+---
+'@primer/react-brand': patch
+---
+
+Fixed missing form-control color tokens in the ESM build. `TextInput`, `Textarea`, `Select`, `Checkbox`, `Radio`, and `FormControl` now bundle the `--brand-control-*` color tokens directly, so their borders and boxes render correctly for consumers that import from `@primer/react-brand/esm` without also loading the global `@primer/react-brand/lib/css/main.css` stylesheet.
+
+Previously these tokens only reached CSS through a shared side-effect import in the forms barrel (`forms/index.ts`), which the ESM build drops. On ESM-only pages this left `var(--brand-control-*)` undefined, collapsing control borders (and rendering checkboxes and radios invisible). This affected the `0.70.0` and `0.71.0` releases; consumers using `@primer/react-brand/esm` should upgrade to this release.
```
rezrah: This is a bit too verbose for our changelog. We want to keep to a single, concise paragraph if it's about one fix as they're converted to bullets in the final output.

```suggestion
Fixed missing styles for ESM form controls, where the dependent design tokens were missing . Affects `TextInput`, `Textarea`, `Select`, `Checkbox`, `Radio`, and `FormControl`.
```

rezrah (approved):
LGTM. One change request for the changelog. Thanks for fixing this 🙇.

PR #1420, opened 2026-07-22
Export `useFocusTrap` hook

### LINKED ISSUE #1399: Export useFocusTrap as part of the public API
`@primer/react-brand` already ships a `useFocusTrap` hook, but it is not exported from the public API, so it is not reachable (it is not re-exported from the index, and the `exports` map has no path that resolves it at runtime).

Could we make it public so consumers can trap focus with Brand alone, without adding another focus-trap dependency? We are building marketing components that use `@primer/react-brand` as the only design-system dependency, and a header with a mobile menu needs this for its open menu.

Ideally exported as:

```ts
import {useFocusTrap} from '@primer/react-brand'
```

Happy to open a PR if that helps.

danielguillan:
## Summary

Resolves #1399

Exports `useFocusTrap` from the `@primer/react-brand` package.

The hook now creates missing refs automatically and captures previous focus when the trap activates.

## List of notable changes:

- **exported** `useFocusTrap` and `FocusTrapHookSettings` from the package root
- **updated** `useFocusTrap` to create missing refs while preserving supplied object refs
- **added** typed refs that work directly with specific HTML elements
- **moved** previous-focus capture into `useEffect` so rendering remains SSR-safe
- **reset** stored previous focus after cleanup so later activations capture the current element
- **added** direct hook, lifecycle, and SSR regression coverage

- #1399

## Contributor checklist:

## Reviewer checklist:

- [ ] Check that there are not other open Pull Requests for the same update/change

## Screenshots:

N/A

### DIFF
--- a/packages/react/src/hooks/useFocusTrap.ssr.test.tsx
+++ b/packages/react/src/hooks/useFocusTrap.ssr.test.tsx
@@ -1,0 +1,16 @@
+/** @jest-environment node */
+
+import React from 'react'
+import {renderToString} from 'react-dom/server'
+import {useFocusTrap} from './useFocusTrap'
+
+describe('useFocusTrap SSR', () => {
+  it('renders without a document', () => {
+    const TestComponent = () => {
+      useFocusTrap()
+      return null
+    }
+
+    expect(() => renderToString(<TestComponent />)).not.toThrow()
+  })
+})
--- a/packages/react/src/hooks/useFocusTrap.test.tsx
+++ b/packages/react/src/hooks/useFocusTrap.test.tsx
@@ -10,6 +10,17 @@ jest.mock('@primer/behaviors', () => ({
 const mockFocusTrap = jest.mocked(focusTrap)
 
 describe('useFocusTrap', () => {
+  const TestComponent = ({
+    disabled,
+    restoreFocusOnCleanUp = false,
+  }: {
+    disabled: boolean
+    restoreFocusOnCleanUp?: boolean
+  }) => {
+    const {containerRef} = useFocusTrap<HTMLDivElement>({disabled, restoreFocusOnCleanUp})
+    return <div ref={containerRef} />
+  }
+
   beforeEach(() => {
     mockFocusTrap.mockReturnValue(new AbortController())
   })
@@ -62,11 +73,6 @@ describe('useFocusTrap', () => {
     const abortSpy = jest.spyOn(abortController, 'abort')
     mockFocusTrap.mockReturnValue(abortController)
 
-    const TestComponent = ({disabled}: {disabled: boolean}) => {
-      const {containerRef} = useFocusTrap<HTMLDivElement>({disabled})
-      return <div ref={containerRef} />
-    }
-
     const {rerender} = render(<TestComponent disabled />)
 
     expect(mockFocusTrap).not.toHaveBeenCalled()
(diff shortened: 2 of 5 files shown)


packages/react/src/hooks/useFocusTrap.ts
```diff
@@ -61,9 +56,12 @@ export function useFocusTrap(
 
   React.useEffect(
     () => {
-      if (containerRef?.current instanceof HTMLElement) {
+      if (containerRef.current instanceof HTMLElement) {
         if (!disabled) {
-          abortController.current = focusTrap(containerRef.current, initialFocusRef?.current ?? undefined) ?? null
+          if (!previousFocusedElement.current) {
+            previousFocusedElement.current = document.activeElement
+          }
```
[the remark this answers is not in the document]
danielguillan: Valid finding. Fixed in 14fd6ba8 by clearing the stored previous focus unconditionally after cleanup. Added regression coverage for reactivation after an SVGElement was active.

packages/react/src/hooks/useFocusTrap.ts
```diff
@@ -34,21 +35,15 @@ export function useFocusTrap(
...
 ): {
-  containerRef: React.RefObject<HTMLElement | null> | undefined
-  initialFocusRef: React.RefObject<HTMLElement | null> | undefined
+  containerRef: React.RefObject<HTMLElement | null>
+  initialFocusRef: React.RefObject<HTMLElement | null>
 } {
-  const containerRef = settings?.containerRef
-  const initialFocusRef = settings?.initialFocusRef
+  const containerRef = useProvidedRefOrCreate<HTMLElement | null>(settings?.containerRef)
+  const initialFocusRef = useProvidedRefOrCreate<HTMLElement | null>(settings?.initialFocusRef)
```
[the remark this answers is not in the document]
danielguillan: Valid finding. Fixed in 14fd6ba8 by making useFocusTrap and FocusTrapHookSettings generic over the container and initial-focus element types. Generated refs now attach to concrete TSX elements without casts.

packages/react/src/hooks/useFocusTrap.ssr.test.tsx
rezrah, line 1: Does this need to be a separate file @danielguillan? Can we add the test case to the [useFocusTrap.test.tsx](https://github.com/primer/brand/pull/1420/changes#diff-f7293b53f1e6eab689ab15c2c8982888cacb945f26f89bd7ecfd748275c49d4f) file?

packages/react/src/hooks/useFocusTrap.test.tsx
```diff
@@ -0,0 +1,112 @@
...
+  it('captures fresh focus after cleaning up a non-HTMLElement active element', () => {
+    const svg = document.createElementNS('http://www.w3.org/2000/svg', 'svg')
+    svg.setAttribute('tabindex', '0')
+    document.body.append(svg)
+    svg.focus()
+
+    const nextFocusedElement = document.createElement('button')
+    document.body.append(nextFocusedElement)
+
+    const TestComponent = ({disabled}: {disabled: boolean}) => {
```
rezrah: We're duplicating this in several different test files. Can we hoist into the `describe` block and share it? It'd make the test easier to read IMO, and we do this in other test files.

rezrah (approved):
LGTM. Thanks for fixing this Dani.

packages/react/src/hooks/useFocusTrap.ssr.test.tsx
danielguillan: I believe so. My understanding is that Jest’s  node  environment is file-scoped.

PR #1429, opened 2026-07-30
[Brand Refactor] MinimalFooter updates

danielguillan:
## Summary

Towards https://github.com/github/brand-experience/issues/223

Updates `MinimalFooter` with refreshed visual treatment, an opt-in custom content slot, and a back-to-top control.

- 📚 [Docs](https://primer.style/brand/components/MinimalFooter/#footnotes)
- 🕹️ [Storybook](https://primer-779ad42239-26139705.drafts.github.io/brand/storybook/?path=/story/components-minimalfooter--default)

## List of notable changes:

- Reorganizes the footer into top, optional content, and bottom sections with updated spacing, dividers, typography, responsive behavior, and GitHub logomark.
- Adds `MinimalFooter.Content` for custom controls or information.
- Adds `MinimalFooter.BackToTop` with localized labels, configurable scroll behavior.
- Updates social icon colors and interaction states across light and dark modes.

- Verify the layout, spacing, dividers, typography, and content order across desktop, tablet, and mobile, with and without social links.
- Check the default and dark color modes, including social icon and Back to Top hover, focus-visible, active, and disabled states.
- Verify Back to Top behavior with pointer and keyboard activation, reduced motion, custom focus targets, and consumer cancellation.

## Steps to test:

1. Open the `MinimalFooter` Playground story.
1. Open the responsive medium and small stories, including the variants without social links, and verify the layout.
1. Open the Custom Content and Back to Top stories and verify their layout and interaction states.
1. Activate Back to Top with pointer and keyboard input. Verify it scrolls to the top, moves keyboard focus to the first `<main>` element, and respects disabled and reduced-motion behavior.

- https://github.com/github/brand-experience/issues/223
- [Minimal Footer design](https://www.figma.com/design/ktrli9XcPCMO7lk2H0vQgD/Docs-2026?node-id=123-6013)

## Contributor checklist:

## Reviewer checklist:

## Screenshots

| Before | After |
| ------ | ----- |
| image       | image       |


packages/react/src/MinimalFooter/MinimalFooter.visual.spec.ts-snapshots/Visual-Comparison-MinimalFooter-MinimalFooter-Default-1-darwin.png
rezrah, line 1: FYI @danielguillan there's alignment issues here for the links

Image

rezrah: The link hover is also really huge, is that intended? Wondering also if we're okay from an a11y standpoint not using the blue links here? Underline should be enough to not rely on colors, I think.

Image

packages/react/src/MinimalFooter/MinimalFooter.module.css
```diff
@@ -1,18 +1,25 @@
 .Footer {
   position: relative;
-  background-color: var(--brand-footer-bg-color);
+  border-top: var(--brand-borderWidth-thick) solid var(--brand-color-success-fg);
+  background-color: var(--brand-color-canvas-subtle);
```
rezrah: Can we move these to design tokens package please @danielguillan, we already have token for the background. 

Also a little concerned about the green border on older pages and subdomain sites, where the style doesn't naturally work. 

cc @natalierharris flagging this as a design risk to be aware of. 

The quickest fix is to turn the border-top value into a css variable, something like `--brand-footer-borderColor` so downstream user can customize it.

packages/react/src/MinimalFooter/MinimalFooter.module.css
```diff
@@ -132,6 +180,94 @@
...
+  }
+
+  .Footer__bottom--no-social .Footer__links {
+    width: auto;
+    justify-content: flex-start;
+    order: initial;
+  }
+}
+
+.Footer .Footer__backToTop {
```
rezrah: Some spacing / padding issues here @danielguillan 

Image

rezrah: Also most of these selectors have higher specificity than they need?

Can we just do this?

```suggestion
.Footer__backToTop {
```

packages/react/src/MinimalFooter/MinimalFooter.module.css
```diff
@@ -132,6 +180,94 @@
...
+  padding: 0;
+  border-color: transparent;
+  background: none !important;
```
rezrah: Is the important needed here? If there's a competing background color, can we make that opt-in so that downstream users can override this value if needed.

packages/react/src/MinimalFooter/MinimalFooter.module.css
```diff
@@ -1,18 +1,25 @@
 .Footer {
```
rezrah: @danielguillan while we're doing this refactor, can we find and replace the Footer class with MinimalFooter?

packages/react/src/MinimalFooter/MinimalFooter.module.css
```diff
@@ -132,6 +180,94 @@
   text-decoration: none;
 }
 
+.Footer__copyright,
+.Footer__link-text {
+  font-family: var(--brand-fontStack-monospace);
+  font-size: var(--base-size-12);
+  line-height: var(--brand-text-lineHeight-100);
+  letter-spacing: var(--brand-text-letterSpacing-100);
```
rezrah: These should be inheritable by setting size="100" on the <Text>? To cut down on redundant CSS can we do a sweep of CSS which is potentially duplicative?

packages/react/src/MinimalFooter/MinimalFooter.module.css
```diff
@@ -132,6 +180,94 @@
...
+}
+
+.Footer__backToTop:not([disabled], [aria-disabled='true']):hover .Footer__backToTop-icon {
+  border-color: var(--brand-button-subtle-borderColor-hover);
+  background-color: var(--brand-button-subtle-bgColor-hover);
+}
+
+.Footer__backToTop:not([disabled], [aria-disabled='true']):active .Footer__backToTop-icon {
+  border-color: var(--brand-button-subtle-borderColor-active);
+  background-color: var(--brand-button-subtle-bgColor-active);
```
rezrah: Using button tokens here feels a tad risky @danielguillan. Could we create MinimalFooter tokens for these and map them to the values or worst case Button? At least that way the coupling is limited to design tokens package.

packages/react/src/MinimalFooter/MinimalFooter.tsx
```diff
@@ -1,14 +1,20 @@
...
+import {useReducedMotion} from '../hooks/useReducedMotion'
 
 /**
  * Design tokens
  */
 import '@primer/brand-primitives/lib/design-tokens/css/tokens/functional/components/footer/colors-with-modes.css'
+/**
+ * `--brand-control-minTarget-coarse` (used by the BackToTop hit-area) isn't part of
+ * the shared stylesheet bundle, so it is imported explicitly here.
+ */
```
rezrah: Nice ESM callout but do we need all the tokens from there? Could we instead map to a specific problem from a new `@primer/brand-primitives/lib/design-tokens/css/tokens/functional/components/footer/base.css`? 

Alternatively, i'm wondering if this is a bug, because i'd expect these control tokens to technically be available in global import. That might be the right way to fix this, wdyt?

packages/react/src/MinimalFooter/MinimalFooter.module.css
rezrah: Using the `success-fg` token also feels brittle, as that can change at any time and is semantically irrelevant here. Can we swap this out for the equivalent value from `--base-color-scale-green-{x}`?

apps/next-docs/content/components/MinimalFooter/index.mdx
```diff
@@ -73,27 +75,86 @@ import {MinimalFooter} from '@primer/react-brand'
...
+  <MinimalFooter.BackToTop>Back to top</MinimalFooter.BackToTop>
+</MinimalFooter>
+```
+
+`children` supplies the required visible and accessible label:
+
+```jsx live
+<MinimalFooter socialLinks={false}>
+  <MinimalFooter.BackToTop>Retour en haut</MinimalFooter.BackToTop>
+</MinimalFooter>
```
rezrah: I think this is fairly self-explanatory. Let's remove to cut down page length 

```suggestion

```

packages/react/src/MinimalFooter/MinimalFooter.tsx
```diff
@@ -95,6 +99,44 @@ export type MinimalFooterProps = {
   copyrightStatement?: string | React.ReactElement
 } & BaseProps<HTMLElement>
 
+/**
+ * Recognizes only the first `MinimalFooter.Footnotes`, the first `MinimalFooter.Content`,
+ * the first `MinimalFooter.BackToTop`, and up to the first five `MinimalFooter.Link`
+ * children. All other children - including duplicates of the single-instance children
+ * and any Link children beyond the fifth - are ignored.
+ */
+function parseRootChildren(children: React.ReactNode) {
```
rezrah: @danielguillan we usually do this inside the component, can we move it back in?

packages/react/src/MinimalFooter/MinimalFooter.tsx
```diff
@@ -283,18 +310,116 @@ const Link = <C extends React.ElementType = 'a'>({as, children, ...rest}: PropsW
...
+
+        window.scrollTo({
+          top: 0,
+          behavior: prefersReducedMotion ? 'instant' : scrollBehavior ?? 'smooth',
+        })
+
+        if (event.detail === 0) {
+          const focusTarget = focusTargetId
+            ? document.getElementById(focusTargetId)
+            : document.querySelector<HTMLElement>('main')
```
rezrah: I don't think we can count of `main` always being present, especially on the subdomain sites. Could we fall back to the body or html tags?

packages/react/src/MinimalFooter/MinimalFooter.tsx
```diff
@@ -283,18 +310,116 @@ const Link = <C extends React.ElementType = 'a'>({as, children, ...rest}: PropsW
...
+      <Text variant="muted" size="200" className={styles['Footer__link-text']}>
         {children}
       </Text>
     </Component>
   )
 }
 
+/**
+ * Props for the optional Back to top control.
+ */
```
rezrah: Nit, this is a redundant annotation. Let's avoid these to keep file sizes down.

```suggestion
```

packages/react/src/MinimalFooter/MinimalFooter.tsx
```diff
@@ -283,18 +310,116 @@ const Link = <C extends React.ElementType = 'a'>({as, children, ...rest}: PropsW
...
+   */
+  focusTargetId?: string
+  /**
+   * Runs before the built-in behavior. Call `event.preventDefault()` to cancel scrolling and focus transfer.
+   */
+  onClick?: React.MouseEventHandler<HTMLButtonElement>
+  /**
+   * The scrolling behavior used when reduced motion is not preferred.
+   */
+  scrollBehavior?: ScrollBehavior
```
rezrah: Does this need to be configurable @danielguillan?

apps/next-docs/content/components/MinimalFooter/index.mdx
```diff
@@ -73,27 +75,86 @@ import {MinimalFooter} from '@primer/react-brand'
...
+| `className` | `string`    | `undefined` | Sets a custom class on the root element |
+| `children`  | `ReactNode` | `undefined` | Renders custom content inside the slot  |
+
+### MinimalFooter.BackToTop
+
+| Name             | Type                                                 |   Default   | Description                                                                                                                                            |
+| :--------------- | :--------------------------------------------------- | :---------: | :----------------------------------------------------------------------------------------------------------------------------------------------------- |
+| `children`       | `ReactNode`                                          |  Required   | The visible and accessible localized label.                                                                                                            |
+| `focusTargetId`  | `string`                                             | `undefined` | ID of the semantic top-of-page element that receives. Defaults to the first `<main>` element.                                                          |
```
rezrah: The description is cut short here:

Think it's supposed to be:

```suggestion
| `focusTargetId`  | `string`                                             | `undefined` | ID of the semantic top-of-page element that receives focus after keyboard activation. Defaults to the first `<main>` element.                                                           |
```

packages/react/src/MinimalFooter/MinimalFooter.tsx
```diff
@@ -202,12 +236,22 @@ function Footnotes({children, className}: PropsWithChildren<FootnoteProps>) {
   )
 }
 
+type MinimalFooterContentProps = React.HTMLAttributes<HTMLElement> & Omit<BaseProps<HTMLElement>, 'animate'>
+
+function Content({children, className, ...rest}: PropsWithChildren<MinimalFooterContentProps>) {
```
rezrah: @danielguillan calling it content feels a bit too broad, generic and permissive. I'm guessing this is for edge-case content, not a general purpose slot. 

For arbitrary content, we've always used props like `leadingComponent` or `trailingComponent` partly because they inherently have intrinsic friction. Giving slots a named child feels like we're encouraging this as first-class API. 

I missed any API explorations here, but did you have other options you considered? E.g. `centerComponent` or similar.

packages/react/src/MinimalFooter/MinimalFooter.tsx
```diff
@@ -103,72 +145,64 @@ function Root({
...
+        <div className={styles['Footer__top-row']}>
+          <LogoLink logoHref={logoHref} />
+          {backToTop}
+        </div>
+      </div>
+    </section>
+  )
 
-  const currentYear = new Date().getFullYear()
+  const bottomSection = (
```
rezrah: Can we inline this and the `section` below in the return like before @danielguillan? I think it helps human readability. Unless there's a reason it's been hoisted.

packages/react/src/MinimalFooter/MinimalFooter.tsx
```diff
@@ -103,72 +145,64 @@ function Root({
...
+            </Text>
+            <div className={styles['Footer__links']}>{links}</div>
+          </div>
+          {hasSocialLinks ? <SocialLinks socialLinks={renderedSocialLinks} /> : null}
+        </div>
+      </div>
+    </section>
+  )
+
+  const sections: {name: FooterSectionName; node: React.ReactNode; className: string}[] = [
```
rezrah: Similar to comment above, wondering why we need to extract prop values like this rather than writing as straight JSX. I'd expect it to be more declaratively written like in other components, but maybe there's some optimization i'm missing here?

packages/react/src/MinimalFooter/MinimalFooter.module.css
```diff
@@ -132,6 +180,94 @@
   text-decoration: none;
 }
 
+.Footer__copyright,
+.Footer__link-text {
+  font-family: var(--brand-fontStack-monospace);
+  font-size: var(--base-size-12);
```
rezrah: Random note: We should consider adding 12px to our type scale. I think we use it more than we realize.

packages/react/src/MinimalFooter/MinimalFooter.module.css
danielguillan: Agree. Moved it to a design token so it can be customized.

packages/react/src/MinimalFooter/MinimalFooter.module.css
danielguillan: Great idea!

packages/react/src/MinimalFooter/MinimalFooter.module.css
danielguillan: 100%. It's fixed now.

packages/react/src/MinimalFooter/MinimalFooter.module.css
danielguillan: Yeah, I can't think of a nice way to keep these in sync while not coupling to button. I've simply added footer-specific tokens using equivalent underlying values for now.

packages/react/src/MinimalFooter/MinimalFooter.tsx
danielguillan: We only need one. Added the one we needed to `base.css` instead

packages/react/src/MinimalFooter/MinimalFooter.tsx
danielguillan: Great call out. Falling back to body/html now.

packages/react/src/MinimalFooter/MinimalFooter.tsx
danielguillan: Agree that calling it `content` is generic and permissive. I don't think it's necessarily for edge-cases, but making it a slot prop feels more appropriate and cohesive. I've named it `centerComponent` for now but I'm open to other suggestions.

packages/react/src/MinimalFooter/MinimalFooter.module.css
danielguillan: Yes, definitely. I always feel bad when I use or see the `--base-size-12` token for a font size.

packages/react/src/MinimalFooter/MinimalFooter.tsx
danielguillan: I’ve removed it. My initial thought was that we should account for different site-specific scrolling behaviors, but Primer Brand should set its own standard. It’s better for the site experience if we define those defaults rather than letting them vary.

packages/react/src/MinimalFooter/MinimalFooter.tsx
rezrah: sounds good, only alternatives to `centerComponent` are adjacent terms like `middle`, etc. Not sure those are better.

the test is going to be, can you see this prop being used in other components (where arbitrary centered content is applicable).

.changeset/bright-dots-rest.md
```diff
@@ -0,0 +1,9 @@
+---
+'@primer/brand-primitives': minor
+'@primer/react-brand': minor
+---
+
+Updated `MinimalFooter` visual design and added a center component prop and back to top control.
```
rezrah, line 6: @danielguillan can we add a little ⚠️ here advising users to check their site manually to ensure it works with the new design, given we're adding some really opinionated defaults here like the green line + background colors.

packages/react/src/MinimalFooter/MinimalFooter.tsx
```diff
@@ -75,11 +77,13 @@ const socialLinkData = {
...
 type SocialLinkName = keyof typeof socialLinkData
-type SocialLink = (typeof socialLinkData)[SocialLinkName]
-
 const socialLinkNames = Object.keys(socialLinkData) as SocialLinkName[]
 
 export type MinimalFooterProps = {
+  /**
+   * An optional component rendered between the top and bottom sections.
+   */
+  centerComponent?: React.ReactElement
```
rezrah, line 89: Should we add the usual annotation here for escape hatches?

```suggestion
  /**
   * Escape-hatch for inserting custom React components.
   * Warning:
   *   This prop isn't advertised in our docs but remains part of the public API for edge-cases.
   *   Need to use this prop? Please check in with #primer-brand first to confirm correct usage.
   */
  centerComponent?: React.ReactElement
```

packages/react/src/MinimalFooter/MinimalFooter.tsx
```diff
@@ -207,14 +211,14 @@ type SocialLinkProps = {name: SocialLinkName}
 const SocialLink = ({name}: SocialLinkProps) => {
   const link = socialLinkData[name]
   return (
-    <li key={name}>
+    <li key={name} data-social-link={name}>
```
rezrah: Didn't spot this earlier, what's the data attribute for? Do we need it?

rezrah (approved):
Did a quick scan and it's looking good @danielguillan. Thanks for addressing the feedback I shared earlier.

packages/react/src/MinimalFooter/MinimalFooter.tsx
danielguillan: That was a leftover. Good catch!

packages/react/src/MinimalFooter/MinimalFooter.tsx
danielguillan: Agree! I've adde it

jesussandreas:
@danielguillan stupid question, why is the footer notes included with the footer, isn't that a separate component but we are merging them together here? 👀

jesussandreas:
@danielguillan where can i see the deployment? it seeems its failing 🥺

danielguillan:
@jesussandreas sorry about the missing deployment, it seems there was a hiccup with GitHub Actions. Here’s the [Storybook link](https://primer-779ad42239-26139705.drafts.github.io/brand/storybook/?path=/story/components-minimalfooter--default)

Regarding footnotes, they’ve been part of the MinimalFooter component since day 1, no changes in this PR. See [current production docs](https://primer.style/brand/components/MinimalFooter/#footnotes).

jesussandreas:
Thank you for that @danielguillan 

Design notes:

Can we then have "try github, enterprise and email" on the top right

Screenshot 2026-08-07 at 10 10 23 AM

above the icons, per design.

Screenshot 2026-08-07 at 10 10 43 AM

the background of the footer being canvas/subtle instead of bg/white.

the font size of the footer, i have as 12px, currently in production shows as 14px - is this because is that the smallest we can go?

the logo size on desktop should be 32px by 32px.

IF we need to keep the footnotes, then i think we should separate them visually.

to something like this, (added on the figma file for reference)

Screenshot 2026-08-07 at 10 20 27 AM

danielguillan:
@jesussandreas feedback addressed. Thanks! ✨

jesussandreas:
Screenshot 2026-08-21 at 10 38 46 AM


@danielguillan it seems it's adding a padding around, is this a storybook thing?

danielguillan:
@jesussandreas yes, that's default Storybook padding on all stories.

jesussandreas (approved):
ship it 🚢

PR #1448, opened 2026-08-19
Adds new InlineCode component

### LINKED ISSUE #1419: [Feature Request] Add `code` to `as` options on `Text` component
### Proposal

There are currently [temporary overrides for inline code styles](https://github.com/github/github-ui/blob/f464e59c3a800187a50f3c89c36ee16df9992c1b/packages/swp-core/components/contentful/shared/inlineCode.module.css#L14-L18) in SWP to address inline `code` markup.

I propose we add styles directly to Primer Brand and add `code` as an option in the `Text` component. The currently allowed strings for the `as` prop are: 'p', 'span', 'div', 'strong' and 'em'

rezrah:
## Summary

Resolves https://github.com/primer/brand/issues/1419

We previously added custom styles to certain components to support inline `<code>` elements. This isn't scalable across all components, and we should have a native component for this use-case. That's what InlineCode is. It's functionally identical to InlineLink, where it applies some standardised spacing and colors, but inherits typographic qualities.

🔗 [Preview docs](https://primer-7772ea065a-26139705.drafts.github.io/brand/components/InlineCode/)
🔗 [Preview storybook examples](https://primer-7772ea065a-26139705.drafts.github.io/brand/storybook/?path=/story/components-inlinecode--default)

## List of notable changes:

- Added InlineCode
- Replaced examples of inline `<code>` in component stories with <InlineCode>
- Matches the styling from BrandX designs latest prototype
- Retained backwards compat

- Is the styling correct across color modes?
- Are sizing and weights correct?

## Steps to test:

1. [Use the InlineCode stories ](https://primer-7772ea065a-26139705.drafts.github.io/brand/storybook/?path=/story/components-inlinecode--default)and verify accuracy of implementation
1. Check the component works [across the type scale](https://primer-7772ea065a-26139705.drafts.github.io/brand/storybook/?path=/story/components-inlinecode-features--text-scale)
1. Use this story for [RIver](https://primer-7772ea065a-26139705.drafts.github.io/brand/storybook/?path=/story/components-river-features--with-inline-code-gridline)
1. Use this story for [CTABanner](https://primer-7772ea065a-26139705.drafts.github.io/brand/storybook/?path=/story/components-ctabanner-features--with-inline-code-element)
1. Use this story for [SectionIntro](https://primer-7772ea065a-26139705.drafts.github.io/brand/storybook/?path=/story/components-sectionintro-features--with-inline-code)

## Contributor checklist:

## Reviewer checklist:

## Screenshots:

**Color modes:** 
Screenshot 2026-08-19 at 11 07 54

**River with gridlines:**
Screenshot 2026-08-19 at 15 03 12

**Localized:**
Screenshot 2026-08-19 at 15 11 35

Before/after comparison of native code solution

<table>
<tr>
</tr>
<tr>
<td valign="top">
Screenshot 2026-08-19 at 11 06 13

 </td>
<td valign="top">
Screenshot 2026-08-19 at 11 06 05

</td>
</tr>
</table>

<table>
<tr>
</tr>
<tr>
<td valign="top">

Screenshot 2026-08-19 at 11 05 38

 </td>
<td valign="top">
Screenshot 2026-08-19 at 11 05 46
</td>
</tr>
</table>


jesussandreas:
couple of changes @rezrah.

for headings, lets increase the border radius to `6px`

i think we also need to update the padding to be a bit more optically aligned.

Screenshot 2026-08-19 at 2 27 58 PM

added 4px of padding bottom just to see here, because the top has a lot but not on the bottom.
Screenshot 2026-08-19 at 2 29 56 PM

thought starter for you, what if we make the font-weight: regular default to 450?
it seems it renders quite bold, resembling more of a medium/semi bold style and doing so, makes it slightly thinner / more on par with mona sans.

lastly, @natalierharris how do you feel if we on headings only, default to text-primary instead of secondary?

reply from nat:

> yes! totally makes sense it helps take away some of the emphasis

rezrah:
>thought starter for you, what if we make the font-weight: regular default to 450?

@jesussandreas we don't support 450 as a weight option since Brand Refactor (we explicitly removed it, if you recall)

Can we use one of these?

Screenshot 2026-08-20 at 12 49 08

apps/next-docs/next-env.d.ts
danielguillan, line 1: Not sure why this was reformatted since it should be auto-generated?

.changeset/add-inline-code.md
```diff
@@ -0,0 +1,28 @@
+---
+'@primer/react-brand': patch
```
danielguillan:
```suggestion
'@primer/react-brand': minor
```
I think we should consider this a `minor` update instead of a `patch` based on our current convention. Also, it looks like we might need `@primer/brand-primitives` too.

danielguillan (approved):
The code looks good. 👍

apps/next-docs/next-env.d.ts
rezrah: it was my local prettier script. I've rolled this back, but also added this file to a new nested prettierignore as this isn't the first or last time it happens. Thanks Dani.

.changeset/add-inline-code.md
rezrah: Yeah, new component, good shout.

rezrah: Don't think primitives needs the same though.

rezrah:
@jesussandreas some updates:

- inline code in headings now use default text color + 6px border radius
- Used 400 instead of 450 for weight as 450 is unsupported and you're trying to reduce weight (see earlier comment above)
- All instances now optically position the text higher in the background

  Adding padding to the bottom doesn't work well. Only for larger headings. It causes overlap on wrapped text and looks bad on smaller sizes like cards
  
  
  Screenshot 2026-08-20 at 12 30 30

Instead i've added some sub-pixel vertical uplift across the board.

  <table>
<tr>
</tr>
<tr>
<td valign="top">
  
Screenshot 2026-08-20 at 14 08 02

 </td>
<td valign="top">
Screenshot 2026-08-20 at 14 08 06

</td>
</tr>
</table>


  <table>
<tr>
</tr>
<tr>
<td valign="top">
  
Screenshot 2026-08-20 at 14 08 02


 </td>
<td valign="top">

Screenshot 2026-08-20 at 14 08 06

</td>
</tr>
</table>



Note: The root cause of the issue is that all examples use lowercase text where the glyphs don't take up the full available space. This gives the impression it's not vertically centered.

There's no easy solution to this AFAIK. I think it's reasonable to assume that those code fragments will be lowercase for the most part, but so we can track the impact long term i've added a story [here](https://primer-7772ea065a-26139705.drafts.github.io/brand/storybook/?path=/story/components-inlinecode-features--uppercase). This clearly shows that the optical alignment doesn't work well if all the text is uppercase. 

Hope this makes sense.

jesussandreas:
Gotcha, thank you for detailed write out.

- Lets use regular then.
- Yeah, I was worried about that merging and so your solution works better.

Lastly regarding the border radius, can you just double check that it applies to these other ones, it seems only large and small one has a border radius but the rest don't?

If they don't have it, perhaps we do something like 4px?

Screenshot 2026-08-20 at 8 50 54 AM

jesussandreas (approved):

PR #1477, opened 2026-09-15
Add `gridline` variant to Section

danielguillan:
## Summary

Resolves https://github.com/github/marketing-engineering/issues/6396

Adds a `gridline` variant to `Section`.

## List of notable changes:

- **added** the `gridline` Section variant.
- **added** responsive inline padding.
- **updated** `rounded` handling so it is ignored when combined with `gridline`, with a warning in development/test environments.

- Verify the gridline borders align with the inner container edges across viewport sizes.
- Verify the responsive inline padding and vertical border extensions at narrow and wide viewport sizes.
- Verify that `rounded` is ignored and emits the expected warning when combined with `gridline`.

## Steps to test:

1. Open the Section component stories in Storybook.
2. Open the `Gridline`, `Gridline With Full Width Container`, and `Gridline With Background Color` stories.
3. Verify the horizontal gridlines, responsive spacing, container-edge alignment, and background-color behavior.
4. Run the Section tests and the repository checks.

- https://github.com/github/marketing-engineering/issues/6396

## Contributor checklist:

## Reviewer checklist:

## Screenshots:

Section gridline Storybook preview


packages/react/src/Section/Section.module.css
```diff
@@ -153,4 +166,29 @@
   .Section-wide--paddingBlockEnd-spacious {
     --padding-block-end: var(--brand-Section-padding-spacious);
   }
+
+  .Section--variant-gridline .Section__container:not(.Section__container--fullWidth) {
+    position: relative;
+    border-inline: var(--brand-borderWidth-thin) solid var(--brand-color-border-muted);
+    padding-inline: var(--base-size-64);
```
danielguillan: @rezrah I’m unsure about adding inline padding to the inner container here. It could make the `gridline` variant more convenient for consumers in some cases, but it feels inconsistent with the default `Section`, which doesn’t provide inner padding and generally expects consumers to use a `Grid` when layout spacing is needed.

One complication is that the padding values referenced in the issue differ from those applied by `Grid`, so I’m not sure whether `Section` should be responsible for reproducing them. What do you think? Should the  gridline  variant own this padding, or should spacing remain the consumer’s responsibility?

packages/react/src/Section/Section.features.stories.tsx
```diff
@@ -81,6 +81,46 @@ export const ResponsivePadding: StoryFn<typeof Section> = () => {
...
+  return (
+    <Section variant="gridline" fullWidth>
+      <Text as="div" className={styles.paddingInnerItem}>
+        Gridline section with a full width container
+      </Text>
+    </Section>
+  )
+}
+
+export const GridlineWithBackgroundColor: StoryFn<typeof Section> = () => {
```
rezrah, line 165: @danielguillan fyi the background isn't stretching to the edge on this viewport like the vertical borders do

Screenshot 2026-09-16 at 12 01 57

packages/react/src/Section/Section.module.css
rezrah: Yeah, i see the concern, but I think it's okay in this instance because:

- gridline doesn't work without that inline padding
- asking users to add it feels prone to error
- we already add block padding, so we already have precedent for internal padding (unless i'm misunderstanding?)
- other gridline variants already include inline padding, mostly because of the first point. This change feels consistent to me.

I think we should go ahead and add the internal padding.

Re. 

> it feels inconsistent with the default Section, which doesn’t provide inner padding and generally expects consumers to use a Grid when layout spacing is needed.

I'm not seeing a lot of folks using Grid alongside Sections in single column layouts. It's more for multi-column layouts. 

> I’m unsure about adding inline padding to the inner container here. 

I am admittedly curious about this. Why would we add this to inner container and not the outer one where block padding is already specified? I found it a little confusing to see two different elements defining their own internal padding on different axes.

packages/react/src/Section/Section.module.css
```diff
@@ -40,20 +49,20 @@
 
 .Section--paddingBlockStart-none,
 .Section-narrow--paddingBlockStart-none {
-  padding-block-start: 0;
+  --padding-block-start: 0;
```
rezrah, line 56: Not super familiar with the history of this component. @danielguillan why do we have local variables for things like block-start here, when we already have tokens to padding like `--brand-Section-padding-normal`. Can we consolidate them into primitives?

packages/react/src/Section/Section.module.css
simurai: > One complication is that the padding values referenced in the issue differ from those applied by Grid

Hmm.. yeah. In the use-case of the issue (https://github.com/enterprise), the `Section` would be used under a RiverBreakout component that uses `--brand-RiverBreakout-variant-gridline-spacing-outerInline`. So having the same inline padding on `Section` would be nice.

Screenshot 2026-09-16 at 20 33 46

But yeah, maybe not as a default if each component can have different inline padding. I saw `FAQ` has a bit less. And maybe sometimes you want to stretch edge to edge without inline padding?

simurai:
Also curious about the gap between multiple `Section`s. Is that a `margin`? There will be cases where we don't want a gap and just a single border in between. Maybe another prop like `noSpacing`?

Screenshot 2026-09-16 at 21 03 40

packages/react/src/Section/Section.features.stories.tsx
danielguillan: That's Storybook's default paddings. I've set the story layout to full so the background now reaches the edges.

packages/react/src/Section/Section.module.css
danielguillan: The `--brand-Section-padding-*` tokens still provide the spacing values. The local variables store the padding selected by the responsive props so we can reuse it for the rounded overlap and gridline borders. Since that value can differ for each Section and breakpoint, it belongs in the component rather than in primitives.

danielguillan:
Thanks @rezrah and @simurai. So the decision is to have inline padding as part of the `gridline` variant and live on the inner container.  `fullWidth` removes the padding and vertical borders for edge-to-edge layouts, so we don’t need another prop. The padding can also be overridden with  `--brand-Section-variant-gridline-paddingInline`, as an escape hatch.

@simurai the spacing between stacked `Section`s is controlled by their parent, for example with `<Stack gap="none">`. I've added [a story for this](https://primer-47cbd1de62-26139705.drafts.github.io/brand/storybook/?path=/story/components-section-features--adjacent-gridline-sections) and made the borders overlap so adjacent `Section`s share one divider instead of showing a double border.

